PR #17, opened 2017-08-07
Disambiguate derivation using the sum rule score.

uriva:
This is untested, but just to get the general idea across.
The priority can be thought of as a negative log probability.

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -140,7 +140,7 @@ class Lark:
                 self.options.ambiguity = 'resolve'
         else:
             assert self.options.parser == 'earley'
-        assert self.options.ambiguity in ('resolve', 'explicit', 'auto')
+        assert self.options.ambiguity in ('resolve', 'explicit', 'sum', 'auto')
 
         # Parse the grammar file and compose the grammars (TODO)
         self.grammar = load_grammar(grammar, source)
@@ -198,4 +198,3 @@ class Lark:
         # else:
         #     l = list(self.lex(text))
         #     return self.parser.parse(l)
-
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -66,11 +66,10 @@ class Earley_NoLex:
 
         rules = [(n, list(self._prepare_expansion(x)), a, o) for n,x,a,o in parser_conf.rules]
 
-        resolve_ambiguity = (options.ambiguity=='resolve') if options else True
-        self.parser = earley.Parser(rules, 
+        self.parser = earley.Parser(rules,
                                     parser_conf.start,
                                     parser_conf.callback,
-                                    resolve_ambiguity=resolve_ambiguity)
+                                    resolve_ambiguity=options.ambiguity or 'resolve')
 
     def _prepare_expansion(self, expansion):
         for sym in expansion:
@@ -93,11 +92,10 @@ class Earley(WithLexer):
 
         rules = [(n, self._prepare_expansion(x), a, o) for n,x,a,o in parser_conf.rules]
 
-        resolve_ambiguity = (options.ambiguity=='resolve') if options else True
-        self.parser = earley.Parser(rules, 
+        self.parser = earley.Parser(rules,
                                     parser_conf.start,
                                     parser_conf.callback,
-                                    resolve_ambiguity=resolve_ambiguity)
+                                    resolve_ambiguity=options.ambiguity or 'resolve')
 
     def _prepare_expansion(self, expansion):
         return [Terminal_Token(sym) if is_terminal(sym) else sym for sym in expansion]
@@ -113,13 +111,12 @@ class XEarley:
 
         rules = [(n, list(self._prepare_expansion(x)), a, o) for n,x,a,o in parser_conf.rules]
 
-        resolve_ambiguity = (options.ambiguity=='resolve') if options else True
         ignore = [Terminal_Regexp(x, self.token_by_name[x].pattern.to_regexp()) for x in lexer_conf.ignore]
 
         self.parser = xearley.Parser(rules,
                                     parser_conf.start,
                                     parser_conf.callback,
-                                    resolve_ambiguity=resolve_ambiguity,
+                                    resolve_ambiguity=options.ambiguity,
                                     ignore=ignore,
                                     )
 
@@ -159,6 +156,3 @@ def get_frontend(parser, lexer):
             raise ValueError('Unknown lexer: %s' % lexer)
     else:
         raise ValueError('Unknown parser: %s' % parser)
-
-
-
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -69,7 +69,7 @@ class Earley_NoLex:
         self.parser = earley.Parser(rules,
                                     parser_conf.start,
                                     parser_conf.callback,
-                                    resolve_ambiguity=options.ambiguity or 'resolve')
+                                    resolve_ambiguity=options.ambiguity if options else 'resolve')
 
     def _prepare_expansion(self, expansion):
         for sym in expansion:
@@ -95,7 +95,7 @@ class Earley(WithLexer):
         self.parser = earley.Parser(rules,
                                     parser_conf.start,
                                     parser_conf.callback,
-                                    resolve_ambiguity=options.ambiguity or 'resolve')
+                                    resolve_ambiguity=options.ambiguity if options else 'resolve')
 
     def _prepare_expansion(self, expansion):
         return [Terminal_Token(sym) if is_terminal(sym) else sym for sym in expansion]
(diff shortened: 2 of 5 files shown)


erezsh:
So what you're saying is that you would like priority to be cumulative?
In your code, from my understanding, a rule with priority 1 can be chosen of a rule with priority 2, if one of its children has priority 2 as well (even if there is no ambiguity).

uriva:
Right, this is like summing log probabilities. This way I can say that some rule is super improbable in a way that affects the entire parse.

uriva:
Somewhat like what is described here: https://en.wikipedia.org/wiki/Statistical_parsing

erezsh:
Sounds like you need a very specific application of disambiguation. Did you consider passing  ambiguity='explicit' to Lark, and handle the ambiguity in your code?

uriva:
https://en.wikipedia.org/wiki/Stochastic_context-free_grammar

erezsh:
Stochastic parsing means choosing a derivation randomly according to some distribution. That is definitely outside the scope of Lark at the moment.

You can implement it yourself by adapting the existing code (in earley.py), or by passing ambiguity='explicit' and handling the tree however you like. You can add priorities to the rules, or you can use them as percentages, like 20 and 80.

If you need any help regarding Lark I'll be more than happy to help. And when you're done, if it works, I will be happy to add it to Lark as a contribution.

uriva:
Let me try and convince you that the way prioritization is currently implemented is not useful. I mean - when we want to prioritize rules, we usually really want to prioritize entire parses.

This is because in the general case two given rules don't appear in the same place in the parse, so will never get compared. So the prioritization currently works on very specific (not useful in my opinion) cases.

So if you agree that we essentially want to prioritize parses, the only simple way to do so is to sum across rules used in a parse. I do agree that whether the scores are interpreted as probability or not, is entirely out of scope.

uriva:
In addition, I don't think that requiring external users should be aware of things like `tree.set()` or `'_ambig'`, and without these concepts they might find it hard to do what I'm suggesting.

uriva:
When trying to implement this from outside using 'explicit' I'm getting more 'drv' nodes than the other way around. With the additional drv nodes my transformer rules are not applied correctly...

erezsh:
Well, I agree that what I have right now sucks. If I wasn't flooded with work, I would try to fix it. But unfortunately it will have to wait, until I can give it proper attention.

I'm not convinced that to "sum across rules used in a parse" is the best way. I would have to examine a few use-cases to conclude anything, and right now I don't have any use-case to consider. Would you like to suggest one?

I agree that it would be nice to find a solution that satisfies everyone, without straining the users. If you would like to design such a solution, I would be happy to consider it. Otherwise, the only thing you really need to know, if to work within the tree, and to be aware of the '_ambig' branch, which doesn't seem that bad to me.

erezsh:
P.S. 'drv' nodes are just children of '_ambig'  nodes. When you resolve ambiguity, it should resolve itself.

uriva:
I'm getting 'drv' nodes with parents other than '_ambig'...

uriva:
I want to implement this - this is essentially this pull request.
But I'm not clear about something - If this is working will you be willing to merge it?

erezsh:
If you are getting additional drv nodes, it might be a bug. Give me some example and I will fix it (or tell you why it's there).

Yes, if you implement a clean solution that passes all the tests, then I will merge it.

uriva:
I think it's ready.

Note:
The broken test seems unrelated (and has some comment stating it is due to a bug?).
The code quality comment is about already existing code / conventions.

erezsh:
At first glance, it looks like something I might merge. But you need a better test for "sum" resolution than just two flat competing rules. I would really like to see a test demonstrating that the sum works, in a situation where the old code wouldn't.
And if you could test more than just one use-case, I would be very happy.

uriva:
No problem, will add something soon.
Thanks for the review.

uriva:
I'm encountering something weird, see #18

uriva:
Ok I made the testing a bit more robust. Wdyt?

erezsh:
So, if I understand the test correctly, adding priority .1 to a rule makes it less likely to be chosen?

uriva:
Right, the score is a negative weight.

erezsh:
So how do you add a positive weight?

uriva:
you increase others (or you can support a minus priority, but that doesn't seem as elegant).

uriva:
Again, the inspiration comes from log probablility, so log probability of 1 is 0, that's the max priority of a rule.
Anything higher than that symbolizes lower probability to choose this rule.
(However, nothing enforces this to be true probability that sums to 1 other than the user)

erezsh:
It seems to me that allowing preference to a specific rule over others is a basic requirement in the mechanics or prioritization. Perhaps more commonly used than negating it.

So you're using a logarithmic scale, but then you sum up the numbers using addition? What's the math behind it? (a link that explains it would be fine)

uriva:
https://en.wikipedia.org/wiki/Likelihood_function#Log-likelihood

uriva:
In general it seems natural to me that all rules have 0 contribution by default, and only when they are annotated with a weight will they affect ambiguity resolution.

Since we start with 0 we are bound to either increase the others, or use negative values.

As the first user of this feature - it is good enough for me without negative numbers for now. If we have a proposition that is better than negation, we can definitely consider it, otherwise let's start with this and iterate as needs arise.

erezsh:
The wikipedia article is pretty vague.

You've yet to demonstrate a single use-case that isn't too contrived or simplistic. The new test is slightly better, but still shows no need for logarithms, and doesn't even test this mechanic. I don't have to know why *you* need this feature, but I need some sense of why *anyone* would need it, if I'm going to admit it into my code-base, which I make an effort to keep small and clear.

uriva:
I'm not sure I understand, if you mean that I haven't demonstrated a need for rule probabilities/log probs then a quick google search for "statistical parsing" will show you a lot of cases.  Almost all NLP tasks using parsers need to prioritize rules in this manner. Here's another link: http://www.cs.upc.edu/~gatius/mai-inlp/parsing_6.pdf

uriva:
Another example: https://books.google.co.il/books?id=A9VD-E5Fl00C&pg=PA24&lpg=PA24&dq=parsing+log+likelihood+probability&source=bl&ots=5uNDJuGsLJ&sig=LvgsZDmFBbkJhdl0xXQolZz0AIs&hl=en&sa=X&ved=0ahUKEwifuJrH_cfVAhXQLVAKHUTGCYw4ChDoAQgkMAA#v=onepage&q=parsing%20log%20likelihood%20probability&f=false

uriva:
> The new test is slightly better, but still shows no need for logarithms, and doesn't even test this mechanic

I only mentioned the log likelihood to give you intuition why it makes sense to sum. Of course the tests don't test this logic - the code simply sums the weight of the rules, and that is what's tested.

If you like, I can add multiple weights on different rules to the tests. Would you like that?

erezsh:
The first article doesn't mention the words "log" or "score", and the second one refers to training weights and ML. They don't pertain to this feature, so I'm not sure why you linked them.

You have not demonstrated that antiscore is more useful than a positive score. While a simple sum might "make sense", nothing about parsing actually makes sense. That's why it has to be either backed up by math (relevant, descriptive math, not hand-wavy estimates), or by extensive tests for specific use-cases. Do you understand my point?

uriva:
The second link explains what a PCFG is and states a method to select parse by summing log probabilties of rules.
Do you have an alternative solution?

uriva:
Here I screenshotted it for you: https://ibb.co/k1OkEv

erezsh:
Okay, maybe I've been going at it the wrong way.
The way it seems to me, you are trying to add an essentially ML feature, which is intended to interface with a ML algorithm. It's probably very difficult to balance it by hand.
However, Lark is first and foremost intended to be used by humans. I have no objection to adding a ML interface, but it has to come in the right form. Based on my understanding so far, I can add your contribution to Lark. However, it will be presented as an experimental plugin, and not as an equal alternative to the existing ambiguity resolver.
That means it sits in its own file (not earley.py), and is treated in the documentation as experimental.
Are you okay with it?

uriva:
Sure, I'm okay with this, as I think disambiguation should be algorithm agnostic. Put it wherever you decide.

If you're curious I'll just add that I don't need this for ML, I'm choosing weights by hand. Also the relation between PCFG and ML as is the one between CFG and ML is strictly optional. Both can be induced or built by hand.

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

uriva:
Hi, any news?

erezsh:
Hi, I will do it soon. I've been busy chasing an elusive bug half the day ;)

erezsh:
Done. Pushed it to latest master.

(It was easier to just rewrite it)

Issue #26, opened 2017-09-25
Nearley macros not supported


kasbah:
I had a go at [converting my nearley grammar](https://github.com/monostable/electro-grammar/tree/python) but I encountered the error:

```  
File "/home/kaspar/.local/lib/python3.5/site-packages/lark/tools/nearley.py", line 116, in _nearley_to_lark
    assert False, directive
AssertionError: include
```

So I am guessing `@include` is not supported yet. I also noticed [the TODO for macros](https://github.com/erezsh/lark/blob/3b0ab867ef330435f3aa83738854a5af7cdaa96f/lark/tools/nearley.py#L122). Is my assessment correct that includes and macros are not yet supported?

erezsh:
That is correct! Do you need both? I imagine includes would be easier to add.

kasbah:
I do make use of both in my grammar, yes. There are always workarounds of course. I think I'll have a go at implementing includes when I get a chance though.

erezsh:
Cool, let me know if you have any questions.

kasbah:
Do you agree that the best way to add macros would be to add support for them in Lark and then translate them?

erezsh:
That would be my choice as well.
Since you are familiar with macros in nearley, can you describe their main properties (i.e. features) ?

kasbah:
They are simple pre-processor functions with variables, that are expanded to several rules. I [added a test case to nearley yesterday](https://github.com/Hardmath123/nearley/pull/311/commits/f17550ad52879bf5f1129f814efbe1020ca78d8d) that should be a simple example of a macro.

erezsh:
The way I understand it, there are the following sub-tasks, all in load_grammar.py:

1) Add macro syntax, both definition and usage
2) In GrammarLoader.load_grammar, verify and create macros (also add macros to Grammar object)
3) In Grammar.compile, apply the macros using a transformer

kasbah:
So, starting with syntax, are you opposed to e.g.:

```
combine[X, Y]: $X | $Y | $X "/" $Y | $Y "/" $X
```

erezsh:
Well, square brackets are already used for optional expressions. But I'm fine with

    combine{X, Y}: $X | $Y | $X "/" $Y | $Y "/" $X

erezsh:
Now that Lark supports templates, this should be a lot easier to do.

erezsh:
Closing at there seems to be no interest in this.

(Possibly nearley support will be removed altogether in the future)

PR #28, opened 2017-10-03
Add support for nearley @include

kasbah:
Solves the first part of #26

There is probably still a problem if you include the same file twice from different locations.


erezsh:
The patch looks good! Any chance you can add a sample nearley grammar with the include directive?

kasbah:
Where should that go, into [tools/nearley.py](https://github.com/erezsh/lark/blob/master/lark/tools/nearley.py#L166), or into the tests dir?

erezsh:
Ideally into tests.

kasbah:
Ok, I have made a unit testcases for the old css grammar test and the include. I added Nearley as a submodule of my own fork as I am currently [waiting for an include test to be merged](https://github.com/Hardmath123/nearley/pull/297) that I made use of. 

I will try and add a test for including the same file twice from different files to Nearley itself and then add a test case to Lark. How do you think Lark will handle that case as it stands?

kasbah:
So, it currently fails the new multi-include test with:

```
  File "lark/load_grammar.py", line 621, in load_grammar
    raise GrammarError("Rule '%s' defined more than once" % name)
GrammarError: Rule 'n_b' defined more than once
```

kasbah:
Ok, this should be good to go. Not sure if I should do anything about the style errors as I have been following what has been done in existing code.

erezsh:
Thanks, merged!
I made a few small corrections, but overall I left it as you wrote it.

PR #32, opened 2017-10-15
Add test for UTF-8 characters in grammar

kasbah:
For python3 it seems ok. It currently fails on python2 with:

```
  File "tests/test_parser.py", line 61, in test_utf8
    l = Lark(g)
  File "lark/lark.py", line 151, in __init__
    self.grammar = load_grammar(grammar, source)
  File "lark/load_grammar.py", line 538, in load_grammar
    tree = self.canonize_tree.transform( self.parser.parse(grammar_text+'\n') )
  File "lark/parser_frontends.py", line 31, in parse
    return self.parser.parse(tokens)
  File "lark/parsers/lalr_parser.py", line 67, in parse
    token = next(stream)
  File "lark/lexer.py", line 159, in lex
    t = Token(type_, value, lex_pos, line, lex_pos - col_start_pos)
  File "lark/lexer.py", line 24, in __new__
    inst = Str.__new__(cls, value)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 1: ordinal not in range(128)
```

I am looking into it but if you can see an obvious fix let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,6 @@ python:
   - "3.6"
   - "pypy"  # PyPy2 2.5.0
   - "pypy3" # Pypy3 2.4.0
-script: python -m tests
+script:
+  - pip install -r nearley-requirements.txt
+  - python -m tests
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -55,17 +55,17 @@ class TestParsers(unittest.TestCase):
         self.assertRaises(ParseError, l.parse, 'a')
 
     def test_utf8(self):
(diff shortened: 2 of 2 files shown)


erezsh:
Ah, yes. For Python2 you have to prefix unicode string with a "u"
So:
    u'±a'
That should fix it.

kasbah:
Ah, right. I fixed that test. There are several issues when using unicode in a Nearley grammar though. It  ends up with the same stack trace when building with python2 or using a python3 build from python2. I'll try and add some tests for this.

PR #33, opened 2017-10-16
Fix unicode issues in Nearly transformation

kasbah:
There is still a caveat: you can't use a unicode Nearley grammar transformed with python3 from python2. If you use python2 to do the transformation then the resulting grammar should be usable from both. The main issue is python3 `repr` won't put a `u` infront of unicode strings. A hacky solution would definitely be possible but I am not sure it's worth it.


erezsh:
Looks good, thanks!

PR #51, opened 2017-12-18
Add discard feature to transformers

kevinlatimer:
This pull requests changes slightly how transformers work in that it adds the ability to discard objects when transforming a tree. The user need only raise Discard() in the transformer.


erezsh:
Hey, so you added a reversed(...) there when it's been removed in latest commits.

If you don't mind, I can just fix it and push it myself.

erezsh:
:+1:

PR #69, opened 2018-01-21
Fixed TypeError on pretty-printing tuples in tree

NighttimeDriver50000:
When a tuple is passed as the argument to percent-formatting, its
elements are interpreted as multiple arguments. The pretty printer
previously passed tuples (e.g. those introduced via a Transformer) from
the tree directly to the percent operator, causing a TypeError because
the format string only calls for a single argument. This fix simply
wraps the argument in a one-tuple to ensure it is not interpreted as
multiple arguments if it itself is a tuple.


erezsh:
Good catch, thanks!

Issue #73, opened 2018-01-25
Performance issues in Earley for highly ambiguous grammars


night199uk:
Hey,

Still having some odd issues I just can't troubleshoot and workaround with precedence in  earley. Precedence just doesn't work the way I expect:

Simple example for parsing a INI style file:
```
CR:             "\r"
LF:             "\n"
CRLF:           CR LF
_TS:            ("\t" | " ")+
_EOL:          _TS? ( LF | CRLF )
_EQ:           _TS? "=" _TS?

WORD:                  /[A-Za-z0-9_][a-zA-Z0-9_.-]*/
ASCII_STRING:          /[\t !-~]+/
CONFIG_TOKEN:          "CONFIG_TOKEN"

config_sections:     config_section
config_section:    "[Config]" _EOL config_statements*
config_statements: _TS? CONFIG_TOKEN _EQ WORD _EOL

text_line:          ASCII_STRING _EOL
empty_line:        _EOL

config_file: ( config_sections | text_line | empty_line )*
```

And the test input:
```
[Config]
    CONFIG_TOKEN = ThisBreaks

Some test here
```

The output from the tree being:
```
config_file
  text_line	[Config]
  text_line	    CONFIG_TOKEN = ThisBreaks
  empty_line
  text_line	This is a test

```
What I'm trying to achieve is for the configuration sections that are understood and specifically configured to be parsed, and for the remaining contents of the file to simply be scanned as text lines. I've tried raising the precedence on config_sections to 2, config_section, config_statements and a bunch of other ways to rejig the rules. I just can't figure out why earley won't parse this. 

What I would expect to see (I get this with lalr/contextual):
```
config_file
  config_sections
    config_section
      config_statements
        CONFIG_TOKEN
        ThisBreaks
  empty_line
  text_line	This is a test
```

Unfortunately I can't use LALR. This is a section of a much larger grammar and that grammar gets Reduce Collisions with LALR.

erezsh:
Hi,

This seems like a bug in one of my optimizations. Sorry! Meanwhile, you can use:

    Lark( ... , earley__predict_all=True)

To make it work. Using 

    config_section.2: ...

should be enough to make sure the parser does what you want.

night199uk:
Okay, this does get me working - but on a larger input file (~200-300 lines), I can't complete the parse and the performance has gone through the floor. 

I see two things:
[...]

PR #77, opened 2018-01-30
Make the Earley parser closer to the spec and add a complete SPPF forest implementation.

night199uk:
Key changes:
Add Items to the current Column and ensure unique *before* adding derivations
  - Ensures all derivations get added to the same unique items.

Add rudimentary SPPF type implementation to derivations, indexed on start and end, end as per:
  - https://www.sciencedirect.com/science/article/pii/S1571066108001497
  - This was required after with fixed _ambig detection.

Remove earley__predict_all property.
  - No longer needed after the above two changes.

### DIFF
--- a/lark/common.py
+++ b/lark/common.py
@@ -8,7 +8,7 @@ Py36 = (sys.version_info[:2] >= (3, 6))
 
 ###{standalone
 def is_terminal(sym):
-    return sym.isupper()
+    return sym.isupper() if sym is not None else False
 
 class GrammarError(Exception):
     pass
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -58,7 +58,6 @@ class LarkOptions(object):
         self.profile = o.pop('profile', False)
         self.ambiguity = o.pop('ambiguity', 'auto')
         self.propagate_positions = o.pop('propagate_positions', False)
-        self.earley__predict_all = o.pop('earley__predict_all', False)
         self.lexer_callbacks = o.pop('lexer_callbacks', {})
 
         assert self.parser in ('earley', 'lalr', 'cyk', None)
@@ -172,7 +171,8 @@ class Lark:
     def _build_parser(self):
         self.parser_class = get_frontend(self.options.parser, self.options.lexer)
 
-        self._parse_tree_builder = ParseTreeBuilder(self.rules, self.options.tree_class, self.options.propagate_positions, self.options.keep_all_tokens)
+        ambiguity = self.options.ambiguity == 'explicit'
+        self._parse_tree_builder = ParseTreeBuilder(self.rules, self.options.tree_class, self.options.propagate_positions, self.options.keep_all_tokens, ambiguity)
         callback = self._parse_tree_builder.create_callback(self.options.transformer)
         if self.profiler:
             for f in dir(callback):
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -4,6 +4,7 @@ from .common import is_terminal, GrammarError
 from .utils import suppress
 from .lexer import Token
 from .grammar import Rule
+from itertools import repeat, product
 
 ###{standalone
 
@@ -79,15 +80,37 @@ def maybe_create_child_filter(expansion, filter_out):
     if len(to_include) < len(expansion) or any(to_expand for i, to_expand in to_include):
         return partial(ChildFilter, to_include)
 
+class AmbiguousExpander:
+    def __init__(self, to_expand, tree_class, node_builder):
+        self.node_builder = node_builder
+        self.tree_class = tree_class
+        self.to_expand = to_expand
+
+    def __call__(self, children):
+        def _is_ambig_tree(child):
+            return hasattr(child, 'data') and child.data == '_ambig'
+
+        ambiguous = [i for i in self.to_expand if _is_ambig_tree(children[i])]
+        if ambiguous:
+            expand = [iter(child.children) if i in ambiguous else repeat(child) for i, child in enumerate(children)]
+            return self.tree_class('_ambig', [self.node_builder(list(f[0])) for f in product(zip(*expand))])
+        return self.node_builder(children)
+
+def maybe_create_ambiguous_expander(tree_class, expansion, filter_out):
+    to_expand = [i for i, sym in enumerate(expansion) if sym not in filter_out and _should_expand(sym)]
+
+    if to_expand:
+        return partial(AmbiguousExpander, to_expand, tree_class)
 
 class Callback(object):
     pass
 
 class ParseTreeBuilder:
-    def __init__(self, rules, tree_class, propagate_positions=False, keep_all_tokens=False):
+    def __init__(self, rules, tree_class, propagate_positions=False, keep_all_tokens=False, ambiguity=False):
         self.tree_class = tree_class
         self.propagate_positions = propagate_positions
         self.always_keep_all_tokens = keep_all_tokens
+        self.ambiguity = ambiguity
 
         self.rule_builders = list(self._init_builders(rules))
 
@@ -103,12 +126,14 @@ class ParseTreeBuilder:
             keep_all_tokens = self.always_keep_all_tokens or (options.keep_all_tokens if options else False)
             expand_single_child = options.expand1 if options else False
             create_token = options.create_token if options else False
+            ambiguity = self.ambiguity
 
             wrapper_chain = filter(None, [
                 create_token and partial(CreateToken, create_token),
                 (expand_single_child and not rule.alias) and ExpandSingleChild,
                 maybe_create_child_filter(rule.expansion, () if keep_all_tokens else filter_out),
                 self.propagate_positions and PropagatePositions,
+                ambiguity and maybe_create_ambiguous_expander(self.tree_class, rule.expansion, () if keep_all_tokens else filter_out),
             ])
 
             yield rule, wrapper_chain
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -113,7 +113,6 @@ class XEarley:
                                     self.match,
                                     resolve_ambiguity=get_ambiguity_resolver(options),
                                     ignore=lexer_conf.ignore,
-                                    predict_all=options.earley__predict_all
                                     )
 
     def match(self, term, text, index=0):
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -13,135 +13,16 @@
 # Author: Erez Shinan (2017)
 # Email : erezshin@gmail.com
 
+#from ..common import ParseError, UnexpectedToken, is_terminal
 from ..common import ParseError, UnexpectedToken, is_terminal
-from ..tree import Tree, Transformer_NoRecurse
+from ..tree import Transformer_NoRecurse
+#from ..lexer import Token
+#from ..grammar import Rule
 from .grammar_analysis import GrammarAnalyzer
+from .earley_common import Column, Item, LR0
+from .earley_forest import Forest, ForestToTreeVisitor
 
-
-class Derivation(Tree):
-    _hash = None
-
-    def __init__(self, rule, items=None):
-        Tree.__init__(self, 'drv', items or [])
-        self.rule = rule
-
-    def _pretty_label(self):    # Nicer pretty for debugging the parser
-        return self.rule.origin if self.rule else self.data
-
-    def __hash__(self):
-        if self._hash is None:
-            self._hash = Tree.__hash__(self)
-        return self._hash
-
-class Item(object):
-    "An Earley Item, the atom of the algorithm."
-
-    def __init__(self, rule, ptr, start, tree):
-        self.rule = rule
-        self.ptr = ptr
-        self.start = start
-        self.tree = tree if tree is not None else Derivation(self.rule)
-
-    @property
-    def expect(self):
-        return self.rule.expansion[self.ptr]
-
-    @property
-    def is_complete(self):
-        return self.ptr == len(self.rule.expansion)
-
-    def advance(self, tree):
-        assert self.tree.data == 'drv'
-        new_tree = Derivation(self.rule, self.tree.children + [tree])
-        return self.__class__(self.rule, self.ptr+1, self.start, new_tree)
-
-    def __eq__(self, other):
-        return self.start is other.start and self.ptr == other.ptr and self.rule == other.rule
-
-    def __hash__(self):
-        return hash((self.rule, self.ptr, id(self.start)))   # Always runs Derivation.__hash__
-
-    def __repr__(self):
-        before = list(map(str, self.rule.expansion[:self.ptr]))
-        after = list(map(str, self.rule.expansion[self.ptr:]))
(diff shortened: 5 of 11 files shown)


night199uk:
Fixes #73.

erezsh:
It's great if it's working for you, but the test-suite returns 91 errors for this pull request. And it's a lot of different exceptions too.

Whatever you fixed, came at the expense of other important properties of the parser.

night199uk:
Pushing up a new revision for feedback now. 

Unfortunately still 2 test case failures.  I know the cause of both; but both are harder to fix. I think to fix both this would need to move to a 'proper' SPPF implementation. The current one is kind of a hack that uses the Items as the SPPF nodes, and it doesn't store the linkages between nodes.

erezsh:
>  I know the cause of both

What is the cause?

night199uk:
Actually I know the cause of one in detail. 
This grammar:
```
start: A A
A: "a"+
```
against the text:
"aaa"
Will fail now with the dynamic lexer. It *will* still parse with no lexer, scanless earley.  

This PR makes items in to_reduce properly unique and pushes a lot of the ambiguity detection into the derivation, moving towards SPPF. For this test case - both possible solutions - a, aa; and aa, a; complete at the same time with same unique item and same PTR and push their results to the same item (hence derivation). I cannot yet figure out a way right now to have the *item* detect that this should be two solutions and make itself ambiguous.

That said :-) I spent a *lot* of time on this, and my thinking right now is I'm not sure this case *should* be handled for lexer based earley. The rule here is ambiguous *at the lexer*. I fee the way this is dealt with in lark right now (dissecting the lexer token and re-parsing it) essentially hacks around that. :-)

This can be fixed (I tested it) by creating tokens for A and A as scanless earley does; essentially turning the lexer based earley into dynamic earley.  I didn't commit that here as I felt it broke the spirit of what you were trying to achieve. I'd be interested in your thoughts.

My position is this test case should be marked as a "won't work" on the lexer based earley though.

The other case I didn't spend as much time on yet so I won't go into detail yet.

night199uk:
BTW, additionally I think it would be easy after this change to support the zero width regexes at the lexer (I think maybe lark can support them without this change as well) - by fundamentally treating them like empty rules, and removing them from to_predict after parsing. After I spend some more time on the last test case I'll see if I can add support here for that as well.

erezsh:
Hi, first let me say - great work! I should have said it before, I know it takes a lot of time and effort, and I'm properly impressed that you managed to improve on my implementation.

I had a couple of hours to gloss over your code. I made some changes, which cleaned it up a bit, and also solved `test_earley_scanless3`, which you addressed in your recent comment. I pushed my changes to the branch `night199uk-sppf`

The only test that's still failing is `test_ambiguity1`, which I feel is important. I make the claim that Lark can parse *any* CFG correctly, and I intend to stand behind it. If you feel like taking a look at it, I'd appreciate it. Or I can take some more time and solve it myself.

erezsh:
Upon further inspection, there is a bug in `DerivationList.__setitem__`: In the entire test-suite (including those with ambiguity), `key in self.tree` always returns false. Thus, the subsequent code that creates `_ambig` trees never runs.

night199uk:
Interesting - thanks, let me check. I've found a couple of other bugs as well here.

night199uk:
Hey, thanks. Okay - I'm about to push up a new branch, hopefully today with some important new improvements. 

I took a look through the mods in night199uk-sppf. Some I can incorporate, but one key one not. Explanation here:
https://github.com/erezsh/lark/blob/de28db6971d9646284ef7517aee012b034844b02/lark/parsers/earley.py#L155

So this is the same problem I originally ran into. Because Derivations are owned by items, we can't simply set a new item into the column (dropping any previous derivations), as this would drop any previous items already at that key (and thus, any derivations attached to it). What we need to do is add derivations to the *existing* item in the column. Hence the flow I took. The root of this is the fact we create new items with empty derivations whose hashes collide with those in the column. Essentially, we create a new item, try to set it into the column, if something already exists at that hash in the column, we use it instead (as it may already have existing derivations). Then we add OUR derivation to whichever item we selected.

In a similar vein, we should not use the item itself as the key for the dict. The item is not immutable, meaning that using it as the key for the dict almost certainly leads to an immutable copy being created under the covers (or hash being called, since it is hashable). Either way hash is called, but this is explicit for an immutable item.

erezsh:
Okay, looking forward.

Just one note: You shouldn't use the explicit hash of the item as a key, because it will allow collisions. Whatever you store as a key, it must implement the same `__eq__` function as items do.

night199uk:
```Just one note: You shouldn't use the explicit hash of the item as a key, because it will allow collisions. Whatever you store as a key, it must implement the same __eq__ function as items do.```

Can you explain what you mean by this? The keys used for a dict (or set) must be immutable. Adding __hash__ to the Items essentially implies they are immutable (even though in this case the objects - Items - are not immutable). Under the covers, even if you're using the Item itself as a key, the hash of that item *at the time of insertion* is what will be used as the key internally by the dict.

In this version, we're using mutable integers as keys; which just happens to be the hash of the object at the time of insertion. It is an explicit version of the contract we're already working to. 

In the original we did this:
```
class Derivation(Tree):
...
    def __hash__(self):
        if self._hash is None:
            self._hash = Tree.__hash__(self)
        return self._hash
...
class Column:
    def add(...):
            item_key = item, item.tree  # Elsewhere, tree is not part of the comparison
...
                    old_tree = self.completed[item_key].tree
...
                    if old_tree.data != '_ambig':
                        new_tree = old_tree.copy()
                        new_tree.rule = old_tree.rule
                        old_tree.set('_ambig', [new_tree])
                        old_tree.rule = None    # No longer a 'drv' node

                    if item.tree.children[0] is old_tree:   # XXX a little hacky!
                        raise ParseError("Infinite recursion in grammar! (Rule %s)" % item.rule)

                    if item.tree not in old_tree.children:
                        old_tree.children.append(item.tree)
...
                    k = item_key if self.predict_all else item
                    if k in self.predicted:
                        continue
                    self.predicted.add(k)
```

This code modifies the tree of an item which is already in the dict/set, thus making it's __eq__ and __hash__ dissimilar and making it's hash invalid and treating it as a mutable object.   This violates the contract principle here, two objects which are equal may now have dissimilar hashes.
http://effbot.org/pyfaq/why-must-dictionary-keys-be-immutable.htm

```
Furthermore it must always be the case that if o1 == o2 (i.e. o1.__cmp__(o2)==0) then hash(o1)==hash(o2) (i.e., o1.__hash__() == o2.__hash__()), regardless of whether the object is in a dictionary or not. 
```
https://wiki.python.org/moin/DictionaryKeys

Secondly, in this case, two NTs completing onto the same item with dissimilar trees (ambiguity) fork 2 distinct new Derivation objects each of which has a copy of the tree of the original items, plus a copy of the new set of items from each completing NT. The desire is to identify that they are two completions of the same object, make the copy in the hash _ambig, and add the tree from the newer item to the former, and then update the prediction with the single ambiguous copy. However, this doesn't work correctly - since the tree is included in the hash this creates two new items in the completed set and the _ambig code is not hit. 

As I mentioned, collisions were occurring already and were exactly the problem.   Imagine two completions (ambiguity) onto the same incomplete non term:
```
class Item(object):
...
    def advance(self, tree):
        assert self.tree.data == 'drv'
        new_tree = Derivation(self.rule, self.tree.children + [tree])
        return self.__class__(self.rule, self.ptr+1, self.start, new_tree)
...
            return [i.advance(item.tree) for i in item.start.to_predict if i.expect == name]
...
                    k = item_key if self.predict_all else item
                    if k in self.predicted:
                        continue
                    self.predicted.add(k)
```

Now these two completions will both be reduced in to_predict and two new items will be created in to_predict: one with the tree of the first completion, one with the tree of the latter. Since (without predict_all) the two new items will hash to the same value, only one will be added to the predictions set and the Derivations added to the other are discarded. 



Sorry, I can't explain this much better right now as it's been a couple of weeks since I went into the original issue and it was a complex one to track down and even harder to explain, so forgive me that. :-) Did you get a chance to play with the test case I uploaded to the original issue?

**Comment by night199uk** [comment]
Sorry, I deleted my original comment as I just understood what you meant about hash collisions. 
What I'm trying to get away from is this code in the original:
```
class Derivation(Tree):
...
    def __hash__(self):
        if self._hash is None:
            self._hash = Tree.__hash__(self)
        return self._hash
...
class Column:
    def add(...):
            item_key = item, item.tree  # Elsewhere, tree is not part of the comparison
...
                    old_tree = self.completed[item_key].tree
...
                    if item.tree not in old_tree.children:
                        old_tree.children.append(item.tree)
...
                    k = item_key if self.predict_all else item
                    if k in self.predicted:
                        continue
                    self.predicted.add(k)
```

This code modifies the tree of an item (and thus the hashable value, if not actually the hash itself) which is already in the dict/set, thus making it's __eq__ and __hash__ dissimilar and treating it as a mutable object.   This violates the contract principle, two objects which are equal may now have dissimilar hashes. 
http://effbot.org/pyfaq/why-must-dictionary-keys-be-immutable.htm
This creates problems for me in complex parses. In the newer SPPF based version, Items are truly immutable and all sets are unique, so this problem goes away.

Did you get a chance to play with the config.py example I original posted?

night199uk:
Pushed - please review. This only adds SPPF, I will add Transitive Items in a separate PR later.
The code can almost certainly be polished some more, but this is as much as I plan to do to it for now.

Couple of notes:
1) I marked the same parser test case I mentioned previously as a skip as I feel that it should be handled at the lexer :-) You could change that if you think it's important to support it and add back the code that dissects parser tokens.
2) Infinite recursion should now be supported/handled natively in earley and no longer throws a ParseError. SPPF forests should support nodes that link back towards the root. The forest->tree mechanism detects infinite loops, so those test cases were updated to check for the right output rather than waiting for a ParseError. You could change that back to throwing a ParseError if you feel it important; but I think this is more consistent with your stated goal of handling any grammar. :-)

night199uk:
Oops, forgot to test on 3.3. Will fix & push once done + clean up some of the codacy issues (I likely won't remove the asserts), shouldn't stop you taking a look - the changes will be nominal.

erezsh:
It's better to avoid names like H, s, w, v, etc., so any name is better than a letter, unless it's misleading.

**Regarding `production` and `children`: ** I looked at your code and it seems to me (correct me if I'm wrong), that what you call production is actually an lr0 instance, so it might be called lr0. And that what you call children are actually a single child (to be added to self.children). But what kind of child? What is its purpose as a datum? That's what I would be asking myself.

**forest.py**: Yes, sorry, you did and that was a good choice! But I would place it in lark/parsers, since that's where that is parser-specific belongs. It's a subtle distinction, between the general idea of grammars and the implementation of how to parse them. It's also a good "beware: hardcore" sort-of enclosure to protect users who are new to the project and trying to find their way in the code :)

Objects such as LR0, Derivation, etc. probably belong in "earley_utils.py" or "earley_structs.py" and once both forest.py and earley.py import from there, there won't be any need for dynamic imports (which are, btw, very very slow).

**walk_norecurse**: I don't know, it's a very complicated function. Usually when a problem doesn't solve easily for me, I start to untangle it in small humble steps. For example:

                    if current.s is None:
                        input_stack.pop()
                        if output_stack:
                            output_stack[-1].children.append(None)
                        continue

                    elif current.is_token:
                        input_stack.pop()
                        if output_stack:
                            output_stack[-1].children.append(current.s)
                        continue         


Can be written as:

                    if current.s is None or current.is_token:
                        input_stack.pop()
                        if output_stack:
                            output_stack[-1].children.append(current.s)
                        continue         

So I repeat this over and over, and soon enough patterns become easier to see. Hope this helps for now :)

night199uk:
Updated with the latest changes and tidy-ups. I'm using a slightly modified version of this + the Joop Leo branch for what I need to get done. There is one failing test here still related to explicit ambiguity. Fixing it is slightly annoyingly complicated to do 'properly', but there are a few ways to work around it. 

For what I need, I don't need explicit ambiguity and I have slightly modified version that does ambiguity resolution as a part of the ForestToTreeVisitor. I can push that into another PR if you're interest; I think it's much cleaner and more flexible than the current ambiguity resolver, but there is no support for explicit ambiguity with it right now so I'm holding back.

night199uk:
This one is complete. I've also added a simpler test case for the original issue that drove me down this path.

night199uk:
++ A bunch of usability & significant performance improvements (~50% on large parses).

jyapayne:
Any updates on merging this in? It looks like a pretty neat change that would be cool to use!

night199uk:
I have a newer version that improves performance a further ~200% or so, that I’ve gradually been refining as we built out some stuff on top of this. Definitely don’t merge this until I can push that up. The new version should have comparable performance to the original Earley, though I haven’t tested that in anger yet. It also should (in theory) have a significantly smaller memory footprint than the current Earley parser. Happy to push that here in a week or two and I need to rebase on master at the same time.

erezsh:
Sounds good!

night199uk:
One thing probably worth mentioning discussing here for want of a better forum: the next version of this PR I push up of this will likely deprecate ambiguity='explicit'.

The easiest way to think about why this is, is to understand that for Earley today lark does:
Parse->Tree (ambiguous)->Tree (unambiguous)

The problem is that Trees do not fundamentally or efficiently support all kinds of ambiguity that Earley can generate. You see this in some of the deep tree copies that were required in the currently Earley implementation. Most academic work on Earley moved to Forests (SPPF); which can support the ambiguity correctly and efficiently and is what I've implemented here.

The version of this PR you see here does this in order to keep ambiguity = 'explicit':
Parse->Forest (ambiguous)->Tree (ambiguous)->Tree (unambiguous)

But this does show up some of the disparities. I've currently avoided doing deep copies and complex conversions for Forest->Tree by not supporting some kinds of ambiguity which is fine since I have control of my grammars and can restructure them.

The new Earley which I will PR when I have time would do:
Parse->Forest (ambiguous)->Tree (unambiguous)

This would mean the old ambiguity='explicit', would instead be replicated by getting the handle to the Forest, instead of the tree and operations to examine the ambiguity would operate there. The advantages of the is that it would be significantly more efficient (CPU/memory) than either this PR /or/ the current version of Earley in use in lark.

I don't believe this is fundamentally problematic; and over time probably just requires adding some niceties to the Forest implementation to make it 'walk and talk' like a Tree so from a user perspective it should be as simple to work with; and documentation for working with the Forest. To me; ambiguity='explicit' seems to mostly be a development/testing tool anyway, since what you almost always want in practise is an unambiguous tree. But I thought it worth mentioning/discussing since it does make kind of an ethos change and removing ambiguity='explicit' might be a breaking change for some folks.

erezsh:
I agree with everything you said except for

> what you almost always want in practise is an unambiguous tree

Actually many Lark users use it for things like NLP, and want to be able to iterate the many representations so they can decide between them using statistical tools, etc.

(Perhaps one day Lark could have better support for applying such statistical weights internally, but that's a big and separate issue)

night199uk:
Thanks, that's very actually useful to know. Let me push up that PR when it's ready and we can look at whether it'll affect those users in a breaking way and what would be needed to smooth that out.

night199uk:
Because this has gone through many iterations now; I'm opting to close this PR and open a new one with the new implementation described above, which is now mostly finished and significantly cleaner.
Please see #164

PR #85, opened 2018-02-14
WIP: Implement Joop Leo's right recursion optimisations using TransitiveItems

night199uk:
This is based on the work on Elizabeth Scott's paper; so it'll probably show both commits. Like that work, this is still a little bit WIP but it's working here on Python 2.7 so I want to raise this PR for initial feedback and review on the approach.


night199uk:
Still needs more work & testing.

night199uk:
I'm closing this; will rework this and rebase once the updated Earley parser is committed if necessary.

PR #89, opened 2018-02-21
Use loops for flow control instead of catching exceptions

psboyce:
While optimizing hot spots in a tool I wrote I saw spot in the LaLR parser where you use try/except for flow control with StopIteration instead of a for loop. Catching exceptions in Python is slower than using a branch or loop and in this case it's unnecessary. In my case changing the try/except ladder to use a for loop granted a minor speed boost.

### DIFF
--- a/lark/parsers/lalr_parser.py
+++ b/lark/parsers/lalr_parser.py
@@ -66,9 +66,7 @@ class _Parser:
             value_stack.append(value)
 
         # Main LALR-parser loop
-        try:
-            token = next(stream)
-            i += 1
+        for i, token in enumerate(stream):
             while True:
                 action, arg = get_action(token.type)
                 assert arg != self.end_state
@@ -77,12 +75,9 @@ class _Parser:
                     state_stack.append(arg)
                     value_stack.append(token)
                     if set_state: set_state(arg)
-                    token = next(stream)
-                    i += 1
+                    break # next token
                 else:
                     reduce(arg)
-        except StopIteration:
-            pass
 
         while True:
             _action, arg = get_action('$END')


erezsh:
Thanks! It's impressive that your tool caught that. What's it called?

psboyce:
My tool didn't catch it; the tool I'm working on is a compiler for
generating bytecode from SVF files. The bytecode is used with a JTAG
programmer that runs on an embedded microprocessor. It's for my job so I'm
afraid the tool is proprietary. It takes 10 minutes to compile a 240,000
line ASCII text file under Python 3 and the parse method here was in the
top 10 longest running functions when I ran cProfile. I saved a couple
seconds. :/

I think it's something to do with my grammar, but since we use the tool
infrequently it's not worth spending the effort figuring out what's wrong.

erezsh:
If you're using LALR(1), then the parse complexity will always be O(n), so although some grammars are faster than others, I don't believe it's a dramatic difference. Also, spending 2.5ms per line is not that bad for an interpreted language. But here are a few tips for speed:

1) Use PyPy. It can be many times faster than CPython. I think you can expect at least x5 times faster parsing. If you have to use CPython, Cython can shave off ~20% as well, without code modifications.

2) Use tree-less parsing, by passing a transformer to Lark, instead of running it afterwards.

3) Use tokens instead of rules whenever possible. So avoid single rules like `plus: PLUS` and try to merge rules like `date: NUM "/" NUM "/" NUM` into `DATE: NUM "/" NUM "/" NUM`. Of course, it's less convenient, but it can shave a few % from the runtime.

Or 4), pay me to either make Lark faster or your specific project faster ;)

Either way, good luck and let me know if you have any more questions!

psboyce:
1. I've tried PyPy, it brings the run time down to just over 2 minutes, but
the reason I used Python is for portability. Most of my coworkers are grey
beards and I don't want to frustrate them by forcing them to install PyPy
if they ever need to run this tool. If portability was not an issue I would
have used Bison.

2. I am using a transformer, it packs SVF statements into bytecode for me
and then I join them all together when the parse is complete.

3. I'll look for more ways to add tokens.

I don't think I'm observing linear complexity but I need to double check
that, I'll let you know what's wrong but I don't have much more time to
spend on optimizing my compiler.

PR #92, opened 2018-02-24
Some tidying and comments on the ParseTreeBuilder.

night199uk:
Some minor tidying and consistency work in the ParseTreeBuilder to try and make it a little easier to understand. No functionality change expected.

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -5,51 +5,119 @@ from .grammar import Rule
 
 ###{standalone
 
-class NodeBuilder:
-    def __init__(self, tree_class, name):
+
+class TreeFilter(object):
+    """Base TreeFilter class.
+       Each TreeFilter should have a create_if_required classmethod, that will return a 
+       callable to create an instance of this filter, called while building the chain.
+    """
+    def __init__(self, next_filter):
+        self.next_filter = next_filter
+
+    @classmethod
+    def create_if_required(cls, rule, **kwargs):
+        raise NotImplemented
+
+    def __call__(self, children):
+        """Called with a list of children for each matching Tree object.
+           Should optionally call up to next_filter.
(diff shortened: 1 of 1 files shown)


night199uk:
Sure, I checked it out. A couple of things from my perspective - a lot of this is subjective and stylistic. 

The create_if_required was designed to isolate all of the logic for a filter into the filter itself, instead of having the decision logic separate from the class in the ParseTreeBuilder.

I find this chunk of code incredibly hard to parse:
```
            wrapper_chain = filter(None, [
                create_token and partial(CreateToken, create_token),
                (expand_single_child and not rule.alias) and ExpandSingleChild,
                maybe_create_child_filter(rule.expansion, () if keep_all_tokens else filter_out),
                self.propagate_positions and PropagatePositions,
            ])
```
Its not clear to me at all what this code is doing just from reading it; I found it a bit 'too clever'. There is so much going on in 4 lines of code and each is not consistent with the last. Function calls, constructors, variables... and using 'and' to return a class in that way breaks my head. :-)

This was the same as the code in the list comprehension:
```
    to_include = [(i, not is_terminal(sym) and sym.startswith('_'))
                  for i, sym in enumerate(expansion)
                  if keep_all_tokens
                  or not ((is_terminal(sym) and sym.startswith('_')) or sym in filter_out)
                  ]
```
Again, the complex logic here was confusing. It literally took me 30 minutes to figure out exactly what this was trying to do, especially the repetitive, but actually different:
```[not] is_terminal(sym) and sym.startswith('_')```
The new version of that is much simpler.

erezsh:
I agree the wrapper chain code is terse and hard to understand. Maybe it would be better if I replaced it with a few if statements.

night199uk:
I'll close this since you've already had a crack at this :-)

PR #95, opened 2018-02-26
Propagated positions are wiped out by Transformers

night199uk:
The mechanism of calling the __default__ method on the Transformer recreates tree items, wiping out any line and column numbers if you have propagated_positions on. This only calls the __default__ method if it exists; otherwise it fixes the tree in place (preserving properties).

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -116,12 +116,13 @@ class Transformer(object):
         try:
             f = self._get_func(tree.data)
         except AttributeError:
-            return self.__default__(tree.data, items)
+            if hasattr(self, '__default__'):
+                return self.__default__(tree.data, items)
+            else:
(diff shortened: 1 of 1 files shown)


erezsh:
Hi, thanks for submitting the PR, but unfortunately I can't merge it, because it breaks the tests.

I'm planning to do an overhaul of how transformers work internally,  and hopefully I'll be able to fix this problem too.

Meanwhile, you can use `Transformer_NoRecurse`, which doesn't alter the tree unless you tell it to.

night199uk:
Yeah, I was in a hurry and saw that afterwards - sorry about that. It was as easy as raising an issue anyway and I needed the workaround for now. You can leave here until you rewrite the Transformers or I can close if you want.

PR #102, opened 2018-03-08
Allow functools partials to be used with the InlineTransformer

night199uk:
This allows you to do:
```
import functools

class MyTransformer(InlineTransformer):
    hexnumber = functools.partial(int, base=16)
    integer = functools.partial(int, base=10)
```
etc.


erezsh:
Good one!

PR #129, opened 2018-04-15
example driven parser errors

drslump:
> Based on this paper [Generating LR Syntax Error Messages from Examples](http://people.via.ecp.fr/~stilgar/doc/compilo/parser/Generating%20LR%20Syntax%20Error%20Messages.pdf).

Implementing user friendly error reporting on a parser is not a especially simple task. While additional rules can be added to the grammar to catch common errors, they might interfere with the correct rules if we're not careful, moreover they grow the grammar and can make the parsing slower.

How this works is by crafting a set of malformed syntax examples, which upon an error can be provided to find the one that generates an equivalent parser state, hence we can override the generic error with a custom label.

This is an example usage for the JSON parser:

```py
    try:
        ast = parse(my_code)
    except UnexpectedToken as ut:
        label = ut.match_examples(parse, {
            'missing opening': ['{"foo": ]}', '{"foor": }}'],
            'missing closing': ['{"foo": [}', '{"foo": {'],
            'trailing comma': ['[,]', '[1,]', '[1,2,]' '{"foo":1,}', '{"foo":false,"bar":true,}']
        })

        if label:
            raise SyntaxError('{} at line {}, column {}.'.format(label, ut.line, ut.column))

        raise SyntaxError(str(ut))
```

Open points:

 - Not sure if this is genarizable to other parser types besides the LALR one
 - Undecided about where to put the `match_example` function, should it be a method of `UnexpectedError` or just a helper function in a module?

### DIFF
--- a/lark/common.py
+++ b/lark/common.py
@@ -17,12 +17,13 @@ class ParseError(Exception):
     pass
 
 class UnexpectedToken(ParseError):
-    def __init__(self, token, expected, seq, index, considered_rules=None):
+    def __init__(self, token, expected, seq, index, considered_rules=None, state=None):
         self.token = token
         self.expected = expected
         self.line = getattr(token, 'line', '?')
         self.column = getattr(token, 'column', '?')
         self.considered_rules = considered_rules
+        self.state = state
 
         try:
             context = ' '.join(['%r(%s)' % (t.value, t.type) for t in seq[index:index+5]])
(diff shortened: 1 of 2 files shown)


erezsh:
Really cool idea! Thanks for the PR

I added this example: https://github.com/erezsh/lark/blob/master/examples/error_reporting_lalr.py

PR #130, opened 2018-04-19
[WIP] antlr-style visitor class for dynamic evaluation of subtrees

Gentle:
all current transformers and visitors and iterators implement strict bottom up parsing, always visiting all leaf nodes first. Some languages might not want that behavior, which is why a more complex visitor is useful

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -10,6 +10,7 @@
 
 ### Advanced
 
+- [error\_reporting\_lalr.py](error_reporting_lalr.py) - A demonstration of example-driven error reporting with the LALR parser
 - [python\_parser.py](python_parser.py) - A fully-working Python 2 & 3 parser (but not production ready yet!)
 - [conf.py](conf.py) - Demonstrates the power of LALR's contextual lexer on a toy configuration language
 - [reconstruct\_json.py](reconstruct_json.py) - Demonstrates the experimental text-reconstruction feature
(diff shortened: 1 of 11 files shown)


Gentle:
consider the following language fragment:

if_statement: "if" bool_expr "then" statement1 "else" statement2

(statement1 and statement2 are the same non-terminal, just numbered for clarity)

whether you want to visit statement1 or statement2 is dependant on the value of bool_expr

example code:

```
class MyVisitor(ComplexVisitor):
    def if_statement(self, tree):
        checker = self.visit(tree.children[0])
        if checker:
            return self.visit(tree.children[1])
        else:
            return self.visit(tree.children[2])
```

Gentle:
if you want to accept this code, I will write tests for it, I just commited the bare basics because I wanted to ask for your opinion first

Gentle:
Example of regular use:

`add: NUMBER ( "+" NUMBER)*`

```python
def add(self, tree):
    values = self.visit_children(tree)
    return sum(values)
```

erezsh:
I understand the need for a top-down visitor, but I'm not sure if `aggregate_result` is important. I think there's a simpler way to write this without losing functionality.

What do you think about calling it `Interpreter`?

And maybe a decorator will make the common use nicer:

```python
class MyInterp(Interpreter):
    @visit_children
    def add(self, values):
        return sum(values)
```

Gentle:
see commit

PR #133, opened 2018-04-23
interpreter: default behavior changed to return the values

Gentle:
this behavior seems to be almost always the preferred one, now you don't need to define a handler for nodes that just visit all their children and return all values


erezsh:
Thanks, I somehow missed it.

PR #144, opened 2018-05-17
toxing for easy multiversion tests

asmodehn:


asmodehn:
Useful in my case, because for testing the importer, I want to do it locally as much as possible, on all my interpreters, before uploading broken stuff...

erezsh:
Why is there a mention of Django?

Looks good otherwise.

PR #145, opened 2018-05-17
adding importer and related tests

asmodehn:
Not ready for merge. For feedback only.

Tested locally on python 3.5 only (tox PR #144, this PR will need rebasing)

Few notes : 

- the filefinder2 dependency is here to support PEP420 (namespace packages) even in python2, and unify the python import API between python2 and python3. It might be desirable or not here... we could drop it and add one importer for python2 ( no namespace package support) and keep this one for python3 (much better import logic in interpreter). But that might complexify the import logic here.

- the test code might be too simplistic, or misguided... the point here is not to check the grammar, but to check the import system works, no matter the type of file structure and environment one can have. But in practice this can be quite hard to define and include in an automated python test... I ll take any suggestion

- Even if import doesn't provide any syntax for it, we might want to somehow parametrize the parser ? maybe through LarkImporter

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ python:
   - "pypy"  # PyPy2 5.8.0
   - "pypy3" # Pypy3 5.8.0-beta0
 script:
-  - pip install -r nearley-requirements.txt
+  - pip install -r nearley-requirements.txt filefinder2
   - python -m tests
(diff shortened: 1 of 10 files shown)


erezsh:
Hi, thanks for the pull request, however..

To restate what I said previously, but with more specificity -- The contents of the following files:

* lark/importer/__init__.py
* lark/importer/_utils.py
* lark/importer/finder.py

And most of `lark/importer/loader.py` too, are *not* specific to Lark, and should reside outside of Lark, in a separate package.
If you prefer, I can refactor the code myself, sometime when I have time in the near future.

erezsh:
Closing this as there hasn't been progress in a while. If you still want to contribute this to Lark, we can discuss it further.

Issue #150, opened 2018-05-23
How to use GLSL lark parser on a QScintilla


brupelo:
Hi, nice to meet you.

Today I've discovered this really promising project and although I haven't tested it yet from what I've read I can say it looks very promising to use. Few days ago I was testing this little pyparsing [glsl parser](https://stackoverflow.com/a/39705506/3809375) and the results I was getting weren't too great for simple piece of code (~0.2-0.3s) so I wanted to give it a shot to lark to see whether I'll be getting faster results.

That said, I've opened this [question](https://stackoverflow.com/questions/50459326/qscintilla-glsl-lexer) couple of days ago but since I've discovered lark I was thinking that instead considering using a `QsciLexerCPP` subclass to address this problem (that is just wrong) I'd write my own `QsciLexerCustom` subclass using lark, more info about [here](https://qscintilla.com/subclass-qscilexercustom). But some questions arise to me:

 - Assuming I've got the lark grammar and built the AST, how would you create the styles with it on a QSciScintilla, could we come up with a little example to illustrate this? Could we come up with a 5-min example showing how to use one of the lark examples on a qscintilla?
 - I'll want proper glsl syntax highlighting on the qscintilla editor but I'll also want to show autocompletions of function definitions, function declarations, ... , will be that possible to achieve by using lark?
 - Is there any freenode IRC channel / forum / SO tag to talk/discuss about lark? I'm posting questions here assuming you'll have the tag "help_wanted" :)

Got few more questions but I don't want to make the thread longer, summing up, trying to learn how to use lark to syntax highlight text on a QScintilla component properly.

Thanks in advance!
Bruno

erezsh:
Hi Bruno,

It looks doable with Lark.

In terms of pseudo-code, I presume it might look something like:

```python
def styleText(self, start, end):
        self.startStyling(start)
        text = self.parent().text()[start:end]

        for token in self.lark_parser.lex(text):
            self.setStyling( len(bytearray(token, "utf-8")), self.style.get(token.type, 0))
```
Where `self.style` is a dictionary of `token_name -> style-number`.

Lark has a newgroup: https://groups.google.com/forum/#!forum/lark-parser

We can continue the discussion there.

Good luck,
Erez

brupelo:
Erez,

[...]

PR #152, opened 2018-05-24
New example showing how to use lark+qscintilla (Issue #150)

brupelo:


erezsh:
File "/home/erez/lark/examples/qscintilla_json.py", line 10, in <module>
    from editor import Editor
ModuleNotFoundError: No module named 'editor'

PR #160, opened 2018-05-29
Makes cyk.py deterministic

ehudt:
Makes cyk.py deterministic by replacing frozenset(...) with tuple(sorted(...))

Non deterministic behavior in cyk.py gives inconsistent results in our parses.


erezsh:
I'm guessing this PR is no longer necessary. If you think otherwise, feel free to re-open it and answer my question.

PR #164, opened 2018-06-04
Make the Earley parser closer to the spec and add a complete SPPF forest implementation (2)

night199uk:
Make the Earley parser closer to the spec and add a complete SPPF forest implementation to handle more types of ambiguity correctly. Should be significantly less memory intensive and perhaps faster.

Further performance improvements will come later.
 - Rewritten along the lines of Elizabeth Scott's parser.
https://www.sciencedirect.com/science/article/pii/S1571066108001497

 - Implement SPPF trees per Elizabeth Scott and Bram van der Sanden's work.
http://www.bramvandersanden.com/post/2014/06/shared-packed-parse-forest/


night199uk:
This is the updated version of the Earley work previously in #77.

This is significantly cleaner now and has a bunch more performance optimisations in it. It should be fast and use a very small memory footprint.

A couple of notes:
1) As per the discussion on #77, this is a breaking change for ambiguity='explicit'. ambiguity='explicit' will now return the root node of the SPPF. Implication is, if you want to work with ambiguities you need to work with the SPPF. Although a slightly higher learning curve this in the end is significantly better for folks using explicit ambiguity; as the original explicit ambiguity trees had quite a few problems in practical usage.
2) There are some tests still failing, all (I think) are due to the new ambiguity resolvers; as I couldn't work out how to fully replicate all the behaviours of the original ambiguity resolvers and get the exact same determinism. I opted to leave the tests failing so others could take a look.  I'd appreciate if someone (@erezsh ?) can take a look and perhaps give some feedback or tweak the new ambiguity resolvers (see earley_forest.py) to be more deterministic or better replicate the originals.
3) This is pretty damn quick now. There are some more performance improvements I have waiting in the wings, but I want to hold out on those until this change gets merged and/or battle tested as they will make the code slightly less readable; but are worth it for significant performance improvements (30-40% minimum improvement + additional memory savings).

erezsh:
Nice, it does seem a lot cleaner. And the errors seem to be solvable.

It's still a bit opaque for me, but hopefully that will improve after a bit of study. Meanwhile, any chance you could add a few comments in `earley.py`?

night199uk:
Here; have a whole pile of comments :-) Let me know if that helps.

erezsh:
Another question. You create symbol nodes and token nodes using memoization on `(s, start, end)`. And then you pass those to the nodes, so they can be used for comparison. But actually, since you memoize them, you only need to test identity, and comparing them is pointless. Right?

Is there any reason at all to save the `start` and `end` inside TokenNode and SymbolNode, except for debugging purposes? The tests, btw, seem to suggest that it makes no difference. But I wanted your take on it.

erezsh:
And another one: `Item` accepts a `node` argument, but it seems that every time you create a new Item, you immediately overwrite it's `node` attribute. So perhaps providing `node` to Item is pointless and can be removed?

erezsh:
@night199uk  Just making sure you didn't miss my questions!

night199uk:
Just saw these; thanks for the nudge. Will look and respond in a few.

night199uk:
Hey, so finally got some time to reply to these:
1) SymbolNodes and TokenNodes:
Good question. I thought on this a bit and you might be right; but some earlier tests led me to believe not. The SymbolNodes are added to PackedNodes and disambiguated again through the set() implementation of children on PackedNodes. The memoization deals with SymbolNodes that are repeated in the same parse cycle, but doesn't deal with a new SymbolNode being added in a later parse cycle. However, the later parse cycle SHOULD have a higher 'start' value and shouldn't be duplicated anyway, so I need to find some more tests. I don't have time to deep-dive this right this second, but I'm still working on this long-term as I still need yet more performance out of this anyway. Another round of tuning on the SPPF is definitely on the horizon. Long term I want to remove SymbolNode and TokenNodes completely and replace them with tuple's and lark.Token's anyway to avoids additional object creation so that will change the implementation a bit. But I want to do that in a way that keeps readability.
2) Item & node: Yes, I think you're probably right. The Item doesn't need a Node constructor; I think it's a remnant from a previous life :-)
3) Item.rule: Still need to dig on this one.

I need to check out the work you did on the PR but I hopefully have some time later this week, so I'll pull and rebase all my parsers on this. :-)

erezsh:
Thanks, looking forward for the rest.
I push to `earley_sppf` a few refactoring that I've done. The tests all pass just the same as before, but since I don't have the whole picture yet, I may have messed something subtle up :)
However, I think these changes make it a little clearer and easier to reason about the code. Let me know what you think. I hope it will be useful as it is, but even if not, hopefully it will inspire future iterations.

erezsh:
I've made a lot of changes to Lark, so I merged them back into the `earley_sppf` branch.

It's basically working, BUT, it occasionally fails on very simple parses. Sometimes it works, sometimes it doesn't. Any chance you can take a look and see why that happens? @night199uk

erezsh:
Okay, nevermind, I solved it. Sorry!

You're still welcome to have a look and see if you can improve it :)

I think I might merge it to master in the next major release. That will give me some time to polish it.

PR #168, opened 2018-06-18
0.6.0 - A backwards-incompatible release, with better code & interface

erezsh:
Main features:
* New transformer & visitor implementations, with a `v_args` decorator, to replace `inline_args` and provide extra functionality.
* Attributes such as `line` and `column` moved to `tree.meta`, for order and performance reasons.
* Official file extension grammars is now `.lark`
* Re-ordered files. Transformers & visitors moved from `tree.py` to `visitors.py`. All exceptions moved to `exceptions.py`
* UnexpectedInput is now superclass of UnexpectedToken and UnexpectedCharacters,
  all of which support the get_context() and match_examples() methods.
* Added %declare, for improved plugin support.
* There are also a few bugfixes attached to these changes.

I'm adding this as a pull request so you can comment and provide notes. In a few days I will merge it to master, and after updating the docs I'll create an official release.

@RobRoseKnows  @dirk-thomas @night199uk @clouds56


RobRoseKnows:
Nice! I've forked GitHub's linguist repository on my account, I was going to add Lark to the list of languages supported by GitHub's highlighting. I was going to point them to the `.tmLanguage` file @dirk-thomas wrote for Sublime for highlighting.

erezsh:
@RobRoseKnows  Sounds good! Let me know if you need any help with that.

EivindEklundGoogle:
I see you're switching to the contextual lexer as a default also in LALR.  I've been having substantial problems with that and Early; it's been doing very, very odd things to my lexing in edge conditions, and I was considering suggesting switching to the standard lexer and ambiguity='explicit' as the default setup, to avoid people getting into weird cases (or alternatively requiring people to always specify what they want, removing all the defaults.)

I can provide a bunch of context on my exact problems, but I presume the discussion of the pull request isn't the optimal place.

erezsh:
@EivindEklundGoogle   Yes, please provide some more info & context on what sort of problems you're having. You can open a new issue, or post in the google-groups.

Btw, I'm aware of two problems with the Earley parser: one, that it's inconsistent with its choice of derivations. And two, that its performance declines rapidly when there's a lot of ambiguity. We're currently working on fixing both of these issues, on branch `earley_sppf`. 

However, I'm not aware of any issue with the contextual lexer, except that the error reporting isn't as good (but much better now in 0.6.0). If you can provide examples of problematic behavior in LALR with the contextual lexer, please do!

PR #170, opened 2018-06-22
Correct parameters to UnexpectedToken

EivindEklundGoogle:
Without this, UnexpectedToken will always display <no context> even if it could show context.

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -197,7 +197,7 @@ class Parser:
 
             if not next_set:
                 expect = {i.expect for i in column.to_scan}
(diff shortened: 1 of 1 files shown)


erezsh:
Thanks for the PR.

The error messages have been changed and improved in Lark 0.6.0  (https://github.com/lark-parser/lark/pull/168)

Can you check if your fix is still necessary there?

EivindEklundGoogle:
From code inspection it does not look like it is necessary there; the refactor looks like it takes care of it.  (Rather than passing the position, it gets it from the token.)  I unfortunately can't quite trivially switch my project to that version and check if that makes things just work - it's wrapped in certain build structures etc.

PR #180, opened 2018-07-07
Added relative imports

### LINKED ISSUE #147: Extend %import functionality: Relative import

julienmalard:
fixes #147 
Please let me know if this fits your expectations!

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -1,6 +1,7 @@
 "Parses and creates Grammar objects"
 
 import os.path
+import sys
 from itertools import chain
 import re
 from ast import literal_eval
@@ -75,6 +76,7 @@ TERMINALS = {
     '_RBRA': r'\]',
     'OP': '[+*][?]?|[?](?![a-z])',
     '_COLON': ':',
+    '_COMMA': ',',
     '_OR': r'\|',
     '_DOT': r'\.',
     'TILDE': '~',
@@ -89,6 +91,7 @@ TERMINALS = {
     '_IGNORE': r'%ignore',
     '_DECLARE': r'%declare',
     '_IMPORT': r'%import',
+    '_FROM': r'%from',
     'NUMBER': r'\d+',
 }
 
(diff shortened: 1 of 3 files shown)


erezsh:
It looks like a great start! Also, good job on creating the tests.

I have a few suggestions:

1) I like to think of the `%` as a command prefix, and not as part of the keyword. In other words, I think the syntax would be better as `%from foo import bar`, instead of `%import`

2) I can see why you chose to have `import` and `rel_import` as two separate rules (and same with `from`), but I still think it probably makes more sense to have only `import` and `from`, and allow `import_args` (probably needs a better name) to be either relative or absolute. I think it will make the grammar shorter, and also simplify the post-processing.

3) There is some repetition in the code you added to `load_grammar`. I think you can merge these two paths into a single one. But if not, then move the repeated code to its own function. (aka DRY)

julienmalard:
Thanks! 2&3 I can easily do; however, I might need a bit of guidance for (1).

I had originally tried it as you said, but when I defined `import` as a separate token (`_IMP_FROM`), the word `import` in `%from x import y`  gets erroneously identified as a `RULE` by the lexer, and not as `_IMP_FROM`, and the parser crashes. I tried using the contextual lalr parser in load_grammar.py, but kept on having the same problem...

Any tips? :)

erezsh:
I can see how it might be tricky to implement.. It's possible to do, but I want to keep Lark's grammar readable using simple implementations (i.e. not require a contextual lexer). That way it maintains a bit of universality. And I don't want to create keywords that can't be used as rules.

So that leaves us two options:

1) We can leave it as `%import` for now.

2) We can change the syntax dramatically. Perhaps instead of

```python
%from path %import a, b, c
```

We can write it as
```python
%import path (a b c)
```
It's less pythonic, but more terse. And maybe a little more elegant, given the restrictions I mentioned.

julienmalard:
Hello,

Sorry for the delay. I personally like the second option better. It seems more aesthetic and also seems to be easier and more elegant to implement. Will push my implementation of it soon.

-Julien

erezsh:
Thanks! I merged it to `master`.

I adapted it to Python2, and made a few refactorings. You check them out at commit https://github.com/lark-parser/lark/commit/454c88b58a3aa15167c7a45b275d2fefb7998498 if you're curious.

julienmalard:
Thank you! I found the edits quite insightful. By the way, out of curiosity, which line(s) were adapted to Python2?

erezsh:
Python2 doesn't have `FileNotFoundError`, so I replaced it with `IOError`. And in order to throw the correct exception (since it's different between 2 and 3) you could either save the last exception, or take the shortcut like I did, and just call `open` again :)

erezsh:
On second thought, that's actually a bug, lol

I should change it to saving the exception instead.

Issue #177, opened 2018-07-02
Reconstructor enters infinite loop on some example grammars


julienmalard:
Hello,

I notice that code reconstruction with both the lark and python (2&3) example grammars enters infinite loops (sometimes it exists with an infinite recursion error message; other times it just eats up half of the CPU usage and nearly all of my RAM). JSON works quite well though.

I know that this functionality is still experimental, but since it would be very important for a new project of mine I was wondering if I could contribute to help solve this issue.

Many thanks!
-Julien

julienmalard:
Hello,

Thanks for the insights. I tried understanding what is going on inside the reconstructor, with modest success, so instead I started making a list of things that work and things that don't. My hope would be to understand which are due to errors in the grammar (i.e., a badly written grammar for code reconstruction purposes) and which are due to errors in Lark (that could be fixed).

Most of these are minimalist examples from errors I encountered while trying to reconstruct with the .lark grammar. 
All use LALR. and the following terminals:

```
_NL: /(\\r?\\n)+\s*/
%import common (WS_INLINE, NUMBER, WORD)
%ignore WS_INLINE
```

Example 1: Starred object
--------------------------
*Grammar*
```
start: (rule | _NL)*
rule: WORD ":" WORD
```

*Code to be parsed and reconstructed*
```
SALUT: BONJOUR
```

*Error*
`lark.exceptions.ParseError: Infinite recursion detected! (rule <NonTerminal('__anon_star_0') : NonTerminal(Token(RULE, '__anon_star_0'))>)`

*Comments*
Same thing happens with a plus instead of a star. Does not seem to cause a problem when included deeper inside a rule, as in the json parser. This seems like an error in Lark?


Example 2: Starred rule (works!)
--------------------------------
*Grammar*
```
start: item*
item: NL
    | rule
rule: WORD ":" NUMBER
```

*Code to be parsed and reconstructed*
```
ABC: 1
DEF: 2
```

*Comments*
Example 1 works when rephrased as in this example so that no parenthesised object is starred.
However, if `NL` is written as an ignored terminal `_NL`, then the parsed tree looks like this:
`Tree(start, [Tree(item, []), Tree(item, [Tree(rule, [Token(WORD, 'ABC'), Token(NUMBER, '1')])]), ...`
...and the reconstructor fails with
`AttributeError: 'WriteTokensTransformer' object has no attribute 'item'`
...because the first new line in the code cannot be reconstructed. This seems like a grammar specification error to me?

[...]

PR #196, opened 2018-07-24
Tests for reconstructors

julienmalard:
Here are a few tests for reconstructors, as discussed in #177. Please let me know if the format is what you expected!

As of now, only `test_starred_rule()` passes; the other three tests currently fail.


erezsh:
Looks good, thanks!

PR #199, opened 2018-07-30
Reconstructor tests

julienmalard:
Two more tests for the reconstructor (check that "!" and "?" prefixes on rules are working).

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -282,6 +282,7 @@ class PrepareAnonTerminals(Transformer_InPlace):
         self.token_set = {td.name for td in self.tokens}
         self.token_reverse = {td.pattern: td for td in tokens}
         self.i = 0
+        self.rule_options = None
 
 
     @inline_args
@@ -323,7 +324,9 @@ class PrepareAnonTerminals(Transformer_InPlace):
             self.token_reverse[p] = tokendef
             self.tokens.append(tokendef)
 
-        return Terminal(token_name, filter_out=isinstance(p, PatternStr))
+        filter_out = False if self.rule_options and self.rule_options.keep_all_tokens else isinstance(p, PatternStr)
+
+        return Terminal(token_name, filter_out=filter_out)
 
 
 def _rfind(s, choices):
@@ -475,13 +478,16 @@ class Grammar:
         # =================
 
         # 1. Pre-process terminals
-        transformer = PrepareLiterals() * PrepareSymbols() * PrepareAnonTerminals(tokens)   # Adds to tokens
+        anon_tokens_transf = PrepareAnonTerminals(tokens)
+        transformer = PrepareLiterals() * PrepareSymbols() * anon_tokens_transf   # Adds to tokens
 
         # 2. Convert EBNF to BNF (and apply step 1)
         ebnf_to_bnf = EBNF_to_BNF()
         rules = []
         for name, rule_tree, options in rule_defs:
-            ebnf_to_bnf.rule_options = RuleOptions(keep_all_tokens=True) if options and options.keep_all_tokens else None
+            rule_options = RuleOptions(keep_all_tokens=True) if options and options.keep_all_tokens else None
+            ebnf_to_bnf.rule_options = rule_options
+            anon_tokens_transf.rule_options = rule_options
             tree = transformer.transform(rule_tree)
             rules.append((name, ebnf_to_bnf.transform(tree), options))
         rules += ebnf_to_bnf.new_rules
--- a/lark/reconstruct.py
+++ b/lark/reconstruct.py
@@ -72,6 +72,7 @@ class Reconstructor:
 
         self.write_tokens = WriteTokensTransformer({t.name:t for t in tokens})
         self.rules = list(self._build_recons_rules(rules))
+        self._parser_cache = {}  # Cache for reconstructor parser trees
 
     def _build_recons_rules(self, rules):
         expand1s = {r.origin for r in rules if r.options and r.options.expand1}
@@ -114,7 +115,11 @@ class Reconstructor:
 
     def _reconstruct(self, tree):
         # TODO: ambiguity?
-        parser = earley.Parser(ParserConf(self.rules, None, tree.data), self._match, resolve_ambiguity=resolve_ambig.standard_resolve_ambig)
+        if tree.data in self._parser_cache:
+            parser = self._parser_cache[tree.data]
+        else:
+            parser = earley.Parser(ParserConf(self.rules, None, tree.data), self._match, resolve_ambiguity=resolve_ambig.standard_resolve_ambig)
+            self._parser_cache[tree.data] = parser
         unreduced_tree = parser.parse(tree.children)   # find a full derivation
         assert unreduced_tree.data == tree.data
         res = self.write_tokens.transform(unreduced_tree)
--- a/lark/reconstruct.py
+++ b/lark/reconstruct.py
@@ -85,7 +85,7 @@ class Reconstructor:
         rule_names = {r.origin for r in rules}
         nonterminals = {sym for sym in rule_names
                        if sym.name.startswith('_') or sym in expand1s or sym in aliases }
-
+        seen = set()
         for r in rules:
             recons_exp = [sym if sym in nonterminals else Terminal(sym.name)
                           for sym in r.expansion if not is_discarded_terminal(sym)]
@@ -97,10 +97,13 @@ class Reconstructor:
             sym = NonTerminal(r.alias) if r.alias else r.origin
 
             if sym in expand1s:
-                expand1s.remove(sym)
-                yield Rule(sym, [Terminal(sym.name)], MakeMatchTree(sym.name, [NonTerminal(sym.name)]))
-
-            yield Rule(sym, recons_exp, MakeMatchTree(sym.name, r.expansion))
+                if sym.name not in seen:
+                    yield Rule(sym, [Terminal(sym.name)], MakeMatchTree(sym.name, [NonTerminal(sym.name)]))
+                    seen.add(sym.name)
+                if sym.name == 'start' or not(len(r.expansion)==1 and isinstance(recons_exp[0], NonTerminal)):
+                    yield Rule(sym, recons_exp, MakeMatchTree(sym.name, r.expansion))
+            else:
+                yield Rule(sym, recons_exp, MakeMatchTree(sym.name, r.expansion))
 
         for origin, rule_aliases in aliases.items():
             for alias in rule_aliases:
--- a/lark/reconstruct.py
+++ b/lark/reconstruct.py
@@ -91,7 +91,7 @@ class Reconstructor:
                           for sym in r.expansion if not is_discarded_terminal(sym)]
 
             # Skip self-recursive constructs
-            if recons_exp == [r.origin]:
+            if recons_exp == [r.origin] and r.alias is None:
                 continue
 
             sym = NonTerminal(r.alias) if r.alias else r.origin
--- a/lark/reconstruct.py
+++ b/lark/reconstruct.py
@@ -96,6 +96,10 @@ class Reconstructor:
 
             sym = NonTerminal(r.alias) if r.alias else r.origin
 
+            if sym in expand1s:
+                expand1s.remove(sym)
+                yield Rule(sym, [Terminal(sym.name)], MakeMatchTree(sym.name, [NonTerminal(sym.name)]))
+
             yield Rule(sym, recons_exp, MakeMatchTree(sym.name, r.expansion))
 
         for origin, rule_aliases in aliases.items():
@@ -114,14 +118,8 @@ class Reconstructor:
         assert False
 
     def _reconstruct(self, tree):
-        if self._parser_cache is None:
-            # TODO: ambiguity?
-            parser = earley.Parser(ParserConf(self.rules, None, tree.data), self._match,
-                                   resolve_ambiguity=resolve_ambig.standard_resolve_ambig)
-            self._parser_cache = parser
-        else:
-            parser = self._parser_cache
-            parser.parser_conf.start = tree.data
+        parser = self._parser_cache
+        parser.parser_conf.start = tree.data
         unreduced_tree = parser.parse(tree.children)   # find a full derivation
         assert unreduced_tree.data == tree.data
         res = self.write_tokens.transform(unreduced_tree)
@@ -133,5 +131,9 @@ class Reconstructor:
                 yield item
 
     def reconstruct(self, tree):
+        # TODO: ambiguity?
+        parser = earley.Parser(ParserConf(self.rules, None, tree.data), self._match,
+                               resolve_ambiguity=resolve_ambig.standard_resolve_ambig)
+        self._parser_cache = parser
         return ''.join(self._reconstruct(tree))
 
--- a/lark/reconstruct.py
+++ b/lark/reconstruct.py
@@ -72,7 +72,7 @@ class Reconstructor:
 
         self.write_tokens = WriteTokensTransformer({t.name:t for t in tokens})
         self.rules = list(self._build_recons_rules(rules))
-        self._parser_cache = {}  # Cache for reconstructor parser trees
+        self._parser_cache = None  # Cache for reconstructor parser trees
 
     def _build_recons_rules(self, rules):
         expand1s = {r.origin for r in rules if r.options and r.options.expand1}
@@ -114,12 +114,14 @@ class Reconstructor:
         assert False
 
     def _reconstruct(self, tree):
-        # TODO: ambiguity?
-        if tree.data in self._parser_cache:
-            parser = self._parser_cache[tree.data]
+        if self._parser_cache is None:
+            # TODO: ambiguity?
+            parser = earley.Parser(ParserConf(self.rules, None, tree.data), self._match,
+                                   resolve_ambiguity=resolve_ambig.standard_resolve_ambig)
+            self._parser_cache = parser
         else:
-            parser = earley.Parser(ParserConf(self.rules, None, tree.data), self._match, resolve_ambiguity=resolve_ambig.standard_resolve_ambig)
-            self._parser_cache[tree.data] = parser
+            parser = self._parser_cache
+            parser.parser_conf.start = tree.data
         unreduced_tree = parser.parse(tree.children)   # find a full derivation
         assert unreduced_tree.data == tree.data
         res = self.write_tokens.transform(unreduced_tree)
--- a/tests/test_reconstructor.py
+++ b/tests/test_reconstructor.py
@@ -92,11 +92,11 @@ class TestReconstructor(TestCase):
         NL: /(\\r?\\n)+\s*/
         """ + common
 
-        code = """
-        a*b
-        """
+        code = ['a', 'a*b', 'a+b', 'a*b+c', 'a+b*c', 'a+b*c+d']
 
-        self.assert_reconstruct(g, code)
+        for c in code:
+            with self.subTest(c):
+                self.assert_reconstruct(g, c)
 
     def test_json_example(self):
         test_json = '''
--- a/tests/test_reconstructor.py
+++ b/tests/test_reconstructor.py
@@ -1,17 +1,19 @@
 import json
 import unittest
 from unittest import TestCase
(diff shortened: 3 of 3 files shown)


erezsh:
Thanks.

Btw, I'm pretty sure you could fix these on your own. Want to give it a try?

julienmalard:
Hello,

I could give it a try! But I will likely have a few questions, if you don't mind...

The first one would be: in the case of the "!" prefix, each token in the list gets assigned an automatic name (e.g., "__ANON_2", "PLUS", etc), which leads to an unexpected token error during reconstruction. Am I right in supposing that the mistake would be in the `WriteTokensTransformer` object?

For the second case ("?") prefix, it seems (to me) that ?-prefixed rules do not show up in the inverse grammar specification, and so cause errors when that rule happens to be used in a parsed tree (since no errors occur when the parsed code "skips over" and does not match that rule directly). I am planning on editing the `_build_recons_rules()` function.

A final observation: I have found the reconstructor to be very slow on Python grammar reconstruction (it uses a very high amount of CPU, up to 50%, for quite a few seconds even for the reconstruction of very simple statements). Could this be due to the fact that, since I could not use ?-prefixed rules, the parse tree inevitably ended up very much complex? Or could there be other places to look at for optimisation (such as using LALR, perhaps?), in your opinion?

Many thanks for your guidance!
-Julien

erezsh:
1. The unexpected token error is exactly that: The token isn't expected, because the reconstructor doesn't know about the `!` operator, and assumes they would be filtered away. You have to tell it that it's not the case (look at the `rule.options` object)

2. For reconstruction, we can treat `?rule` the same way we treat `_rule`. It's essentially the same, but only applies when `len(expansion) == 1`. And since we're using a parser to reconstruct the parsed tree, it will match it only if it's relevant (there's a chance for ambiguity, but that's another story).

3. Yes, it's very slow. Using LALR could make it much faster, but I don't know if LALR is strong enough to reconstruct the Python grammar. This is all very new research, and I don't think anyone ever attempted it before, so I don't have any answers!

If you want to improve the speed, the first thing I would try is to cache to reconstruction parsers. Right now we're building them over and over for each branch of the parsed tree.

julienmalard:
Hello,

Thanks! I implemented both 1 and 3 (and 3 is a dramatic speedup!). For 2, though, I am having some complications. Am I right in supposing that the change should be done in

erezsh:
I don't understand your implementation of 1. Why did you change load_grammar? It seems you put your attention on `keep_all_tokens`, which is a Lark instance option, and it is separate from the `!` rule modifier.

Regarding 2, I think you're on the right track. `rule` is a terminal, and `_rule` is a nonterminal, and so '?rule' is both a terminal *and* a non-terminal.

julienmalard:
Hello,

Thanks! I had a few more questions, if you don't mind...

1. I thought that the `keep_all_tokens` flag was set according to the "!" flag (from the `options_from_rule()` function in `load_grammar`, where `keep_all_tokens = name.startswith('!')`. I then told the transformer to keep all tokens on a specific rule only if that rule's `keep_all_tokens` flag was set to ``True``. I didn't want to do it token by token, because the same token, for example `+`, could theoretically appear in one !-prefixed rule and also in a regular rule, in which case its text should only be expected to have been kept in the former case...
2. I am still a bit uncertain, though, with regards to what the reconstructed rules should look like. As of now, I see that they have the general form `NonTerminalRuleName: TerminalChildRule [...]` when the child rule is a normal rule and `NonTerminalRuleName: NonTerminalChildRule [...]` when the child rule is prefixed with `_`. However, in the case of `?` prefixes on the child rule, what should the rule look like? Should it be duplicated, one specifying a NonTerminal for the child rule and the other specifying a Terminal? I do not see how the parent rule can know whether its child rule will have `len(expansions) > 1` or not ahead of time...and when a rule has multiple children (e.g., `Terminal(sum_stmt) Terminal(sum_stmt)` for a `mult_stmt`, would that mean that the `mult_stmt` rule will have to be copied four times (2 * 2 possibilities for each ?-prefixed child rule)? I think I could implement that, but wanted to be sure that my conceptual understanding is correct before getting bogged down in something. :)

Thanks a lot,
Julien

erezsh:
1. I got confused, yes `keep_all_tokens` is the correct option. I still don't understand why you needed to change load_grammar. Can you explain that?

2. You *could* duplicate the rule, that would work. But perhaps a more clever solution is to produce a non-terminal for the rule, and include the terminal inside it. Does that make sense?

erezsh:
1. That actually makes a lot of sense. I had a different idea, but I like your approach better!

2. Precisely so!

julienmalard:
Thanks!

One the second point, I am having some complications. It seems that the Earley parser is misreading a Var, Var tree as an ambiguous sum_stmt or mult_stmt, when only the latter is correct. It then resolves `a*b` as a sum instead of a multiplication in the test suite (`a*b + c` is correctly reconstructed, though, because it leaves no ambiguity).

Anyways, I am not too sure how to fix this. I have a feeling that I defined the rules incorrectly for the reconstruction of ?-prefixed rules, but can't seem to figure out where the mistake is. I pushed my progress so far, in case you might be able to help!

Thanks,
Julien

erezsh:
Hi Julien,

Your implementation is correct. The problem is deeper than that. 

For example, consider this change:

        ?mult_stmt: [mult_stmt "*"] _sum_stmt
        _sum_stmt: [var "+"] var

Both 'a+b' and 'a*b' parse into 

    mult_stmt
        var	a
        var	b

So 'a+b' is a legitimate interpretation by the parser in both cases. Such a grammar can't be reconstructed faithfully.

Obviously to us, the `?sum_stmt` case is different, because they do parse differently, due to the `len(expansion)==1` limitation. But the parser doesn't know about that.

So what is the bug?

This rule:

    <NonTerminal('sum_stmt') : Terminal('var') Terminal('var')>

Should not possible.

We should remove expansions bigger than 1 for all rules of the form `?rule`. **except** when they are the starting rule, in which case we should allow them.

I hope I didn't make it too complicated, it's a little late here. If you have more questions I'll try to answer tomorrow. And if you understand my idea, then good luck :)

julienmalard:
Hello,

Thanks! Which time zone are you on? I updated the test to include different possible code snippets to reconstruct and interestingly enough I am only getting passes on all of them when I remove the rule

`<NonTerminal('mult_stmt') : NonTerminal('sum_stmt')>`

(Removing `<NonTerminal('sum_stmt') : Terminal('var') Terminal('var')>` does not seem to work as the reconstructor gets confused when parsing `a+b`.)
But I am really not sure why this is working (I found it through trial and error), or whether it is the real fix or simply cancelling out one error with another one...

Thank you!
-Julien

erezsh:
I'm on GMT+2

Well, it's not a real fix. It only works on these specific examples.

Note that I said

> except when they are the starting rule, in which case we should allow them.

In `a+b` the starting rule is `sum_stmt`, in which we should keep `sum_stmt: var var`

I get that it's a little complicated. If this is going over your head, I don't mind continuing this on my own from this point.

julienmalard:
Hello,

Sure, help would be very much appreciated. :) I understand what you mean, but am not too sure where to implement it...

Thanks for all the guidance!
-Julien

erezsh:
Done, all tests passing. Pushed changes to `fix_recons` branch.

julienmalard:
Wow, many thanks! It also seems a lot faster than before. I am getting quite fast parses and reconstructs for the python parser test (times are in seconds):

image

erezsh:
Really?  Are you sure you're reconstructing them?

On my computer, even the simplest python code takes forever to reconstruct.

julienmalard:
Fairly sure! I am using an adapted Python grammar from [here](https://github.com/lassi-samaaj/Lassi/blob/master/%E0%A8%B2%E0%A9%B1%E0%A8%B8%E0%A9%80/Extensions/%E0%A8%AA%E0%A9%88%E0%A8%A7%E0%A8%BE%E0%A8%A8_%E0%A9%A9/%E0%A8%AA%E0%A9%88%E0%A8%A7%E0%A8%BE%E0%A8%A8%20%E0%A9%A9.lark) and it is working quite rapidly!
I only have two issues to resolve,
1. Overintenting on some lines during reconstruction, and 
2. Adding spaces after keywords during reconstruction.

erezsh:
That's interesting. Do you mind sharing your code?

julienmalard:
Sure! I am using the Python parser example from Lark, with the slight modification in `test_python_lib()` as follows (only for Python3):

```
files = glob.glob(path+'/*.py')
start = time.time()
reconst = Reconstructor(python_parser3)
for f in files:
    print( f , end=" ")
    try:
        # print list(python_parser.lex(_read(os.path.join(path, f)) + '\n'))
        try:
            xrange
        except NameError:
            time_file = time.now()
            tree = python_parser3.parse(_read(os.path.join(path, f)) + '\n')
            new = reconst.reconstruct(tree)
            print(round(time_file - time.now())
        else:
            python_parser2.parse(_read(os.path.join(path, f)) + '\n')
```

Apologies for any bugs; the original code is on my office computer so I recreated this in the GitHub editor now.

-Julien

julienmalard:
Hello,

Sure! Here is the full final code:

```
import sys
import os, os.path
from io import open
import glob, time

from lark import Lark
from lark.indenter import Indenter

# __path__ = os.path.dirname(__file__)

class PythonIndenter(Indenter):
    NL_type = 'NEWLINE'
    OPEN_PAREN_types = ['LPAR', 'LSQB', 'LBRACE']
    CLOSE_PAREN_types = ['RPAR', 'RSQB', 'RBRACE']
    INDENT_type = 'INDENT'
    DEDENT_type = 'DEDENT'
    tab_len = 8

kwargs = dict(rel_to=__file__, postlex=PythonIndenter(), start='file_input')

python_parser3 = Lark.open('C:/Users/jmalar1/PycharmProjects/Lassi/ਲੱਸੀ/Extensions/ਪੈਧਾਨ_੩/ਪੈਧਾਨ ੩.lark',parser='lalr', **kwargs)

def _read(fn, *args):
    kwargs = {'encoding': 'iso-8859-1'}
    with open(fn, *args, **kwargs) as f:
        return f.read()

def _get_lib_path():
    if os.name == 'nt':
        if 'PyPy' in sys.version:
            return os.path.join(sys.prefix, 'lib-python', sys.winver)
        else:
            return os.path.join(sys.prefix, 'Lib')
    else:
        return [x for x in sys.path if x.endswith('%s.%s' % sys.version_info[:2])][0]

def test_python_lib():

    path = _get_lib_path()

    files = glob.glob(path+'/*.py')
    Recons = Reconstructor(python_parser3)

    start = time.time()

    for f in files:
        print( f , end=" ")
        try:
            start_file = time.time()
            tree = python_parser3.parse(_read(os.path.join(path, f)) + '\n')
            new = Recons.reconstruct(tree)
            print(round(time.time() - start_file, 2))
        except:
            print ('At %s' % f)
            raise

    end = time.time()
    print( "test_python_lib (%d files), time: %s secs"%(len(files), end-start) )


if __name__ == '__main__':
    test_python_lib()
```

And the results...
image

julienmalard:
And a shorter example with the reconstructed code. The reconstruction completes in a fraction of a second on my computer (excluding parsing time).
With the same parser and reconstructor configuration as my previous comment, and the following code:
```

c = """

\"\"\"un long commentaire\"\"\"

a = 2
a*=3
b+=1
e**3
b and c or d
b or c and d
class b(object):
    @staticmethod
    def mult(x, y=1):
        func(2)
        super()
        return x*y
class c(object):
    def mult(x, y=1):
        return x*y

"""

arbre = python_parser3.parse(c)
print(arbre)
nouveau = Reconstructor(python_parser3).reconstruct(arbre)
print(nouveau)
```

... I get the following output:
```
Tree(file_input, [Token(NEWLINE, '\n\n'), Tree(simple_stmt, [Tree(await_expr, [Tree(string, [Token(LONG_STRING, '"""un long commentaire"""')])]), Token(NEWLINE, '\n\n')]), Tree(simple_stmt, [Tree(expr_stmt, [Tree(await_expr, [Tree(var, [Token(NAME, 'a')])]), Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '2')])])]), Token(NEWLINE, '\n')]), Tree(simple_stmt, [Tree(expr_stmt, [Tree(await_expr, [Tree(var, [Token(NAME, 'a')])]), Tree(augassign, [Token(__ANON_4, '*=')]), Tree(testlist, [Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '3')])])])]), Token(NEWLINE, '\n')]), Tree(simple_stmt, [Tree(expr_stmt, [Tree(await_expr, [Tree(var, [Token(NAME, 'b')])]), Tree(augassign, [Token(__ANON_2, '+=')]), Tree(testlist, [Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '1')])])])]), Token(NEWLINE, '\n')]), Tree(simple_stmt, [Tree(power, [Tree(await_expr, [Tree(var, [Token(NAME, 'e')])]), Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '3')])])]), Token(NEWLINE, '\n')]), Tree(simple_stmt, [Tree(or_test, [Tree(and_test, [Tree(await_expr, [Tree(var, [Token(NAME, 'b')])]), Tree(await_expr, [Tree(var, [Token(NAME, 'c')])])]), Tree(await_expr, [Tree(var, [Token(NAME, 'd')])])]), Token(NEWLINE, '\n')]), Tree(simple_stmt, [Tree(or_test, [Tree(await_expr, [Tree(var, [Token(NAME, 'b')])]), Tree(and_test, [Tree(await_expr, [Tree(var, [Token(NAME, 'c')])]), Tree(await_expr, [Tree(var, [Token(NAME, 'd')])])])]), Token(NEWLINE, '\n')]), Tree(compound_stmt, [Tree(classdef, [Token(NAME, 'b'), Tree(arguments, [Tree(await_expr, [Tree(var, [Token(NAME, 'object')])])]), Tree(suite, [Token(NEWLINE, '\n    '), Token(INDENT, '    '), Tree(compound_stmt, [Tree(decorated, [Tree(decorators, [Tree(decorator, [Token(AT, '@'), Tree(dotted_name, [Token(NAME, 'staticmethod')]), Token(NEWLINE, '\n    ')])]), Tree(funcdef, [Token(NAME, 'mult'), Tree(parameters, [Token(NAME, 'x'), Token(COMMA, ','), Tree(paramvalue, [Token(NAME, 'y'), Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '1')])])])]), Tree(suite, [Token(NEWLINE, '\n        '), Token(INDENT, '        '), Tree(simple_stmt, [Tree(await_expr, [Tree(funccall, [Tree(var, [Token(NAME, 'func')]), Tree(arguments, [Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '2')])])])])]), Token(NEWLINE, '\n        ')]), Tree(simple_stmt, [Tree(await_expr, [Tree(funccall, [Tree(var, [Token(NAME, 'super')])])]), Token(NEWLINE, '\n        ')]), Tree(simple_stmt, [Tree(flow_stmt, [Tree(return_stmt, [Tree(testlist, [Tree(term, [Tree(await_expr, [Tree(var, [Token(NAME, 'x')])]), Token(STAR, '*'), Tree(await_expr, [Tree(var, [Token(NAME, 'y')])])])])])]), Token(NEWLINE, '\n')]), Token(DEDENT, '')])])])]), Token(DEDENT, '')])])]), Tree(compound_stmt, [Tree(classdef, [Token(NAME, 'c'), Tree(arguments, [Tree(await_expr, [Tree(var, [Token(NAME, 'object')])])]), Tree(suite, [Token(NEWLINE, '\n    '), Token(INDENT, '    '), Tree(compound_stmt, [Tree(funcdef, [Token(NAME, 'mult'), Tree(parameters, [Token(NAME, 'x'), Token(COMMA, ','), Tree(paramvalue, [Token(NAME, 'y'), Tree(await_expr, [Tree(number, [Token(DEC_NUMBER, '1')])])])]), Tree(suite, [Token(NEWLINE, '\n        '), Token(INDENT, '        '), Tree(simple_stmt, [Tree(flow_stmt, [Tree(return_stmt, [Tree(testlist, [Tree(term, [Tree(await_expr, [Tree(var, [Token(NAME, 'x')])]), Token(STAR, '*'), Tree(await_expr, [Tree(var, [Token(NAME, 'y')])])])])])]), Token(NEWLINE, '\n\n')]), Token(DEDENT, '')])])]), Token(DEDENT, '')])])])])

"""un long commentaire"""

a=2
a*=3
b+=1
e**3
bandcord
borcandd
classb(object):
        @staticmethod
    defmult(x,y=1):
                func(2)
        super()
        returnx*y
classc(object):
        defmult(x,y=1):
                returnx*y
```

Three minor glitches:
1. I cannot seem to be able to prefix the `await_expr` rule with a `?` in the grammar specification. If I do, it parses just fine but the reconstructor runs into an infinite loop that eats up all the alloted RAM and CPU. So all my parse trees contain `await_expr` rules.
2. Keywords are not wrapped with spaces in the reconstruction.
3. The indentor adds an extra indent for the first line of every indented block.

But I think I should be able to fix those soon.

Thanks a lot!
-Julien

julienmalard:
P.S. I just noticed that I had not pushed the working most recent version of my modified Python3 grammar to the link above. Sorry about that! It's fixed now.

By the way, by adding a postprocessor option to reconstructors (similar to postlexers), I was able to correctly reconstruct Python3 code for the entire standard lib, reparse it, and assert that the reconstructed parse tree is identical to that generated for the original code! Times for initial parsing and reconstruction are still around 1-10 seconds for each standard lib file.

Please do let me know if you would be interested in a pull request for the postprocessor option!

```
import sys
import os, os.path
from io import open
import glob, time
from pprint import pprint

from lark import Lark
from lark.indenter import Indenter

# __path__ = os.path.dirname(__file__)

class PythonIndenter(Indenter):
    NL_type = 'NEWLINE'
    OPEN_PAREN_types = ['LPAR', 'LSQB', 'LBRACE']
    CLOSE_PAREN_types = ['RPAR', 'RSQB', 'RBRACE']
    INDENT_type = 'INDENT'
    DEDENT_type = 'DEDENT'
    tab_len = 8

kwargs = dict(rel_to=__file__, postlex=PythonIndenter(), start='file_input')

python_parser3 = Lark.open('C:/Users/jmalar1/PycharmProjects/Lassi/ਲੱਸੀ/Extensions/ਪੈਧਾਨ_੩/ਪੈਧਾਨ ੩.lark',parser='lalr', **kwargs)
def postproc_python3(l):
    for x in l:
        if not hasattr(x, 'type') or not x.type in ['INDENT', 'DEDENT']:
            yield '{} '.format(x) if len(x.strip()) and not x.strip(' \t').endswith('\n') else x

def _read(fn, *args):
    kwargs = {'encoding': 'iso-8859-1'}
    with open(fn, *args, **kwargs) as f:
        return f.read()

def _get_lib_path():
    if os.name == 'nt':
        if 'PyPy' in sys.version:
            return os.path.join(sys.prefix, 'lib-python', sys.winver)
        else:
            return os.path.join(sys.prefix, 'Lib')
    else:
        return [x for x in sys.path if x.endswith('%s.%s' % sys.version_info[:2])][0]

def test_python_lib():

    path = _get_lib_path()

    files = glob.glob(path+'/*.py')
    Recons = Reconstructor(python_parser3)

    start = time.time()

    for f in files:
        print( f , end=" ")
        try:
            # print list(python_parser.lex(_read(os.path.join(path, f)) + '\n'))
            start_file = time.time()
            tree = python_parser3.parse(_read(os.path.join(path, f)) + '\n')
            new = Recons.reconstruct(tree, postproc=postproc_python3)
            print(round(time.time() - start_file, 2))
            assert python_parser3.parse(new) == tree
        except:
            print ('At %s' % f)
            raise

    end = time.time()
    print( "test_python_lib (%d files), time: %s secs"%(len(files), end-start) )


if __name__ == '__main__':
    test_python_lib()
```

erezsh:
Ah, that makes more sense! You had to change the Python parse-tree so that it will contain a lot of punctuation tokens and assisting rules. Still, I'm glad it works for you, and that the performance is reasonable. (PyPy would make it a LOT faster, of course)

> Please do let me know if you would be interested in a pull request for the postprocessor option!

Sure, that would be nice!

PR #205, opened 2018-08-06
Slots fix

henryiii:
Once a class is `__dict__` based, slots can no longer be added on subclasses (that is, they have no effect). The current SlottedTree does not therefore really have slots. ~~This reverses the inheritance of `Tree` and `SlottedTree`, so that `SlottedTree` really will be slots based instead of `__dict__` based.~~

Edit: This now drops SlottedTree entirely, simply making Tree slotted. This should provide the benefits of slotted trees, and since it's hard to get access to the original Tree, should be pretty safe for user code. Let me know if the original dual method is needed. If a user wants a Dict based tree, it's just two lines:

```python
class UserTree(Tree):
    pass
````

I've also added `(object)` in more places, since forgetting it will create an old-style class in Python 2.

I would highly recommend adding `__slots__` in more places; on classes that have many instances you should see a nice memory reduction/speed up I think (especially on older Pythons, I think 3.6 or so improved dictionaries a bit for instances).

## Example illustrating slots inheritance

```python
In [4]: class NoSlots: # Skipping (object) is okay on Python 3
   ...:     pass
   ...:
   ...: class ReallySlots:
   ...:     __slots__ = ('x',)
   ...:
   ...: class NotReallySlots(NoSlots):
   ...:    __slots__ = ('x',)
   ...:

In [5]: NoSlots().x = 1

In [6]: a = ReallySlots()

In [7]: a.x = 1

In [8]: a.y = 3
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-8-af5b34806ad0> in <module>()
----> 1 a.y = 3

AttributeError: 'ReallySlots' object has no attribute 'y'

In [9]: b = NotReallySlots()

In [10]: b.x = 1

In [11]: b.y = 3

In [12]: hasattr(a, '__dict__')
Out[12]: False

In [13]: hasattr(b, '__dict__')
Out[13]: True
```

### DIFF
--- a/lark/common.py
+++ b/lark/common.py
@@ -11,14 +11,14 @@ Py36 = (sys.version_info[:2] >= (3, 6))
 
 
 
-class LexerConf:
+class LexerConf(object):
     def __init__(self, tokens, ignore=(), postlex=None, callbacks=None):
         self.tokens = tokens
         self.ignore = ignore
         self.postlex = postlex
         self.callbacks = callbacks or {}
 
-class ParserConf:
+class ParserConf(object):
     def __init__(self, rules, callback, start):
         self.rules = rules
         self.callback = callback
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -46,7 +46,7 @@ class Rule(object):
         return 'Rule(%r, %r, %r, %r)' % (self.origin, self.expansion, self.alias, self.options)
 
 
-class RuleOptions:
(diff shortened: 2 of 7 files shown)


henryiii:
Unfortunately, all the `isinstance(*, Tree)`s will need to become `isinstance(*, SlottedTree)`. It will take me longer to get to that.

PR #236, opened 2018-09-18
Error Recovering Feature - WIP (Work in Progress)

### LINKED ISSUE #227: Error recovery: Can I add some error recovery strategy?
I see parsers like ANTLR offer some extension support for it: [ANTLRErrorStrategy](https://www.antlr.org/api/Java/org/antlr/v4/runtime/ANTLRErrorStrategy.html) and [Error Reporting and Recovery](http://media.pragprog.com/titles/tpantlr/errors.pdf)

I looked into the [error_reporting_lalr.py](https://github.com/lark-parser/lark/blob/master/examples/error_reporting_lalr.py), and these examples, Lark just bail out parsing upon the first syntax error. It seem to correspond to the ANTLR [BailErrorStrategy](https://www.antlr.org/api/Java/org/antlr/v4/runtime/BailErrorStrategy.html).

I found this thesis from 2005, [lark.pdf](http://www.cocolab.com/products/cocktail/doc.pdf/lark.pdf), there, they talk about error recovering. Is that thesis related with this Lark, or it is some other project with the same name?

evandrocoan:
I cherry picked some changes from my master branch where I worked on the basic error feature support. Now lark does not stop on the first error, but the old behavior can be set with:
```python
json_parser = Lark(json_grammar, parser='lalr', debug=0)
json_parser.parser.parser.parser.error_recovering = False
```

I think the the error recovering strategy implemented on `lalr_parser.py` is not optimal because while running sometimes `examples.error_reporting_lalr.py` it reports 2 errors, other times 1 error only. For example, running  `python -m examples.error_reporting_lalr`, you got now:
```
Parsing example 1
Caught 2 exceptions:
Missing Closing at line 1, column 14.

{"example1": "value"
             ^

Unexpected token Token(<EOF>, Token(ESCAPED_STRING, '"value"')) at line 1, column 14.
Expected one of: 
	* RBRACE
	* COMMA


Parsing example 2
Caught 2 exceptions:
Missing Opening at line 1, column 14.

{"example2": ] 
             ^

Missing Opening at line 1, column 12.

{"example2": ] 
           ^
```

But another time latter you can get:
```
Parsing example 1
Caught 1 exceptions:
Missing Closing at line 1, column 14.

{"example1": "value"
             ^


Parsing example 2
Caught 2 exceptions:
Missing Opening at line 1, column 14.

{"example2": ] 
             ^

Missing Opening at line 1, column 12.

{"example2": ] 
           ^
```
i.e., the first error is dangling between runs.

___
I also added a wrapper library for the python built-in logging module as an optional dependency. I wrote and use everywhere to help me understand the system and log messages. I tried to take care on how I affects performance, but still need to do some benchmarks. In any case, a quick find/replace/comment all can be used.

The debugging library usage is configured in 3 phases. The first one is on the file `utils.py`:
```py
log = getLogger('lark', force=1)
```
This forces all loggers loaded after to be set with level 1 of log level. This keeps them quiet while lark does its static member initialization.

The second phase is on the `lark.py` file. Enabling it here, allows us to see all logging messages after the static initialization of the lark library components:
```py
if debug: log.setup(force=debug)
```
This uses the same value passed to `Lark()` constructor for the debug level. But now, instead of accepting a boolean, `debug` is an integer for the debug level where 0 is completely disabled. 

The third phase are all `getLogger()` function calls or `log.debug_level = 1` assignments. When not instantiating a logger with kwarg `force=1`, each loggers level is controlled by the file they are in:
```py
log = getLogger('lark', 1)

# or
log = getLogger(1, 'lark')

# or
log = getLogger('lark')
log.debug_level = 1
```
Creates a logger with the level 1 valid only for the current file. This allows fine grained logging output control for each file while developing.

___ 
### Todo

1. [ ] Maybe disable lexical errors for the next lines after a syntax error because when a syntactical error happens, it can create some lexical errors on the following lines. (Perhaps they are induced by the contextual lexer?)
1. [ ] Try to move most error recovering code from `error_reporting_lalr.py` into lark
1. [ ] Allow to dynamically exchange the syntactical error recovering heuristics
   1. [ ] Then, add a new alternative syntactical error recovering heuristic implemenation
1. [ ] Fix any regression Unit Tests broken
1. [ ] Unit Tests for Lexical Error Recovering implemented
1. [ ] Unit Tests for Syntactical Error Recovering implemented
1. [ ] Fixes #227 Error recovery: Can I add some error recovery strategy?


erezsh:
Hi @evandrocoan,

That's a really big pull request! Usually I prefer smaller PRs. That way I can discuss the different changes separately, and merge them in selectively.

Honestly, I was thinking about a very different interface.

My idea was to add a new parameter for Lark, like this:

```python
parser = Lark(grammar, parser="lalr", on_error=my_on_error
```

And then the user can do something like:

```python
errors = [ ]
def on_error(error):
    if len(errors) < MAX_ERROR:
         errors.append( error )
         return error.expected[0]
    else:
         raise TooManyErrors( )
```

Or alternatively, something like:

```python
COMMON_MISTAKES = {",", "}", ";"}
def on_error(error):
     ok_mistake = set(error.expected) & COMMON_MISTAKES
     if ok_mistake:
          return list(ok_mistake)[0]
     else:
          raise error
```

It seems to me much simpler and flexible, and allows the user to easily do whatever they need.

What do you think?

evandrocoan:
Hi @erezsh,

I think it a very nice interface. I did not had though about using closures.

I can open a series of pull requests integrating all the features on this pull request.

I would have to start with the following sequence of difficulty:
1. [ ] Add .gitignore (easy)
1. [ ] Add the logger (medium)
1. [ ] Add Lexical Error Recovering + Unit Tests & Regression Testing (hard)
1. [ ] Add Syntactical Error Recovering + Unit Tests & Regression Testing (very hard)

erezsh:
Hi,

See my comments regarding .gitignore

I'm not too enthusiastic about adding a logger, but I'm willing to consider it.

Regarding the on_error interface, I'll add it myself in the upcoming days. If you want to add to it, or provides tests, that'd be great.

evandrocoan:
Closed by #296 comment.

PR #243, opened 2018-10-01
Allow to pass to lark/tree.py pydot__tree_to_png

evandrocoan:
For the pydot tree shaping. For example, passing "TB" instead of "LR" makes the tree
to be draw vertically instead of horizontally.

duplicated_contexts

duplicated_contexts

https://stackoverflow.com/questions/29003465/pydot-graphviz-how-to-order-horizontally-nodes-in-a-cluster-while-the-rest-of-t

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -138,7 +138,7 @@ def pydot__tree_to_png(tree, filename, rankdir="LR"):
     """
 
     import pydot
-    graph = pydot.Dot(graph_type='digraph', rankdir)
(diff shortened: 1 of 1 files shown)


erezsh:
Looks good, thanks

erezsh:
Btw, next time please rebase, to get rid of extra "fix" commits.

evandrocoan:
Sorry, I did another commit because I already had committed the first one, pushed it (pushed too soo), and though my upstream was protected against forced push. But now I remember it is actually not protected again forced push, I could amend the commit and force push it.

PR #245, opened 2018-10-02
Created the token.pretty method

evandrocoan:


erezsh:
That looks more like "debug" than "pretty"

erezsh:
Traditionally, `pretty()` methods are used for generating strings that are more concerned with readability (through style) than with providing the full information (or a repr).

I'm not convinced your addition is useful in the general sense, outside of a very specific use-case.

evandrocoan:
At the time, seeing the token info on that format seemed "pretty" and I just picked this name. But as you said now `(through style)`, I am not sure the name `pretty` applies anymore. It could be called `debug` instead of `pretty` then.

To know whether is addition in general is useful or not, we could conduce a `User Experience Testing` with grammar developers and ask them how much often would like like the token representation be while working:
1. `(true|false) `
1. `Token(__ANON_2, '(true|false) ')`
1. `[@7,162:174='(true|false) '<__ANON_2>,12:10]`

Particularly, I would use either the first or the third option for token representation while developing, because sometimes I do not want to know specifics about the token, while other times I would like to know the most about them.

I think most developers are not used to show much details as the third option does. The question whether it helps or not still need to be investigated. For my experience, the more detailed the better to early spot bugs. For example, the other day I found C++ saying `-0.9 < -0.9` is true.

erezsh:
> To know whether is addition in general is useful or not, we could conduce a User Experience Testing with grammar developers

Sure, sounds good

PR #260, opened 2018-10-24
Method iter_subtrees_topdown added

worldmind:
I am not a big hacker, but looks working

### DIFF
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -76,7 +76,11 @@ Returns all nodes of the tree whose data equals the given data.
 
 #### iter_subtrees(self)
 
-Iterates over all the subtrees, never returning to the same node twice (Lark's parse-tree is actually a DAG)
+Iterates over all the subtrees, never returning to the same node twice (Lark's parse-tree is actually a DAG).
+
+#### iter_subtrees_topdown(self)
+
+Iterates over all the subtrees, return nodes in order like pretty() does.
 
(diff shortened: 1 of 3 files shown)


erezsh:
Hi, looks like a good start!

For a method that is part of the core tree class, it's a good idea to add a unit-test. For example, in `tests/test_trees.py`

Also, I believe the `if` at the start of the method is unnecessary?

worldmind:
> it's a good idea to add a unit-test

yes, I think same, but not found tests for `iter_subtrees` and did not do, I make it later, maybe tomorrow.

> I believe the if at the start of the method is unnecessary?

yes, agree, I will remove it

worldmind:
@erezsh why you use unittest module? As for me it so unpythonic, maybe I can make new test with pytest? It can [run unittests tests](https://docs.pytest.org/en/latest/unittest.html).

erezsh:
Is there really such a difference? I prefer not to add a new dependency unless there's a good reason (right now Lark has no dependencies at all, except for optional support for Nearley).

worldmind:
I can not say that is so important, I just like write beautiful code without no needed boilerplate text.
I push some simple test.

worldmind:
Maybe I will try squash three commits to one?

erezsh:
Looks good. Thanks!

worldmind:
Thanks for your project!

PR #263, opened 2018-10-25
Improved LALR conflict reporting code

SupraSummus:
This patch fixes "shift/reduce conflict" message printed when there is in fact reduce/reduce conflict. It also adds some
code comments and pulls message constructing code into separate function.

### DIFF
--- a/lark/parsers/lalr_analysis.py
+++ b/lark/parsers/lalr_analysis.py
@@ -50,9 +50,18 @@ class IntParseTable(ParseTable):
         return cls(int_states, start_state, end_state)
 
 
-
-
 class LALR_Analyzer(GrammarAnalyzer):
+
+    @classmethod
+    def action_str(cls, act):
+        return '%s: %s' % act
+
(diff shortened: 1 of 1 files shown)


erezsh:
No, you can't just ignore reduce/reduce errors, you have to fail with an exception.

lark/parsers/lalr_analysis.py
```diff
@@ -79,21 +88,25 @@ def step(state):
...
                         if x[0] is Shift:
-                            lookahead[k] = [x]
-
-            for k, v in lookahead.items():
-                if not len(v) == 1:
-                    raise GrammarError("Collision in %s: %s" %(k, ', '.join(['\n  * %s: %s' % x for x in v])))
+                            last_shift = x
+
+                    if last_shift is None:
+                        raise GrammarError(self.conflict_str(k, v))
```
SupraSummus: @erezsh failing on reduce/reduce happens here

Issue #258, opened 2018-10-22
Does lark warn about conflicts in LALR?


SupraSummus:
Lets take grammar:
```
start: as as
as: a*
a: 'a'
```
I believe there is shift/reduce conflict when building LALR for this grammar, but Lark constructs a parser without any warning.

Is there a switch to make lark warn about conflicts? It would be very helpful, for me at least ;)

erezsh:
You're right. I pushed a new fix to `master` that restores the collision warnings. Call Lark with `Lark(..., parser='lalr', debug=True)` to see them (make sure your logging level is also at logging.DEBUG)

SupraSummus:
Yep, now I get tons of shift/reduce conflict warnings. Thanks for the fix and example code.

SupraSummus:
One more improvement would be to add instructions on how to get shift/reduce conflict warnings to docs. Maybe I'll do it soon.

PR #266, opened 2018-10-26
Instructions how to get Shift/Reduce messages printed (Issue #258)

SupraSummus:


erezsh:
Thanks! Any contribution to the docs is appreciated.

PR #267, opened 2018-10-26
Removed unused imports and variables

SupraSummus:
Fixes for arbitrarily selected subset of flake8 warnings.


SupraSummus:
@erezsh could you merge this (or give me feedback why not)? This patch is not immediately necessary in terms of functionality but it improves code quality and performance (a bit, but it does).

erezsh:
It will cause a collision with a big merge that's coming soon. I will accept it after I merge.

SupraSummus:
Sure, I understand ;) Will wait patiently and resolve any possible conflicts then.

PR #290, opened 2018-12-18
Implement Joop Leo's optimisations for right recursion

night199uk:
This **massively** improves the performance when using right recursive grammars by avoiding the quadratic behaviour of base Earley when using right recursion.

Viz, simple worst case all right recursive grammar.

```
grammar = """
a: "A" a
 |
"""
test = 'A' * 1500
parser = Lark(grammar, parser='earley', start='a', ambiguity = 'explicit', lexer="dynamic")
tree = parser.parse(test)
```

0.7b timings: 
```
real	2m5.618s
user	2m4.413s
sys	0m0.960s
```

0.7b-joop timings:
```
real	0m0.509s
user	0m0.338s
sys	0m0.111s
```
(yes, really).

There are a couple of cleanup commits in here I'd like to commit first - we can either do them as part of this PR or I can nudge those in separately; also this is based off my pyDot branch, so we'll probably want to commit that PR first.

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -16,7 +16,7 @@ from ..visitors import Transformer_InPlace, v_args
 from ..exceptions import ParseError, UnexpectedToken
 from .grammar_analysis import GrammarAnalyzer
 from ..grammar import NonTerminal
-from .earley_common import Column, Item
+from .earley_common import Item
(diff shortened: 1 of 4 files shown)


erezsh:
Very cool!

Hope it's okay that I just merged it, I only saw your email afterwards :)

PR #292, opened 2018-12-18
number add BIN_NUMBER

Haski-server:
Without BIN_NUMBER, the following error occurs:
lark.exceptions.UnexpectedCharacters: No terminal defined for 'b' at line 1 col 2

0b10
 ^


erezsh:
Thanks!

PR #296, opened 2018-12-22
Lexer error recovery

evandrocoan:
Continuation from: https://github.com/lark-parser/lark/pull/236#issuecomment-423159458

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.pyc
 *.pyo
+/.tox
 /lark_parser.egg-info/**
 tags
 .vscode
--- a/lark/common.py
+++ b/lark/common.py
@@ -1,11 +1,14 @@
 
 
+from .utils import reraise
+
 class LexerConf:
-    def __init__(self, tokens, ignore=(), postlex=None, callbacks=None):
+    def __init__(self, tokens, ignore=(), postlex=None, callbacks=None, on_error=reraise):
         self.tokens = tokens
         self.ignore = ignore
         self.postlex = postlex
         self.callbacks = callbacks or {}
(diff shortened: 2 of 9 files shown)


erezsh:
This seems like a reasonable patch. But, I'm not sure what's the benefit here?

What do you provide in on_error that won't cause a corrupted state in case of errors?

erezsh:
In the same vein, a few unitests or an example would be very helpful

evandrocoan:
I am working on it right now. First I had to figure out how to run a single unit test.
```shell
python -m unittest tests.test_parser.TestLalrStandard.test_lexer_error_recovering
```

erezsh:
Just for the record, I just run the tests like this

    python -m tests

Or

    python -m tests TestClass.test_name

erezsh:
Thanks for the test, but honestly I'm still not sure what is a practical way to use this feature.

Do you have an example (even tiny) that could be considered real-world error recovery?

evandrocoan:
I was trying to get an example. I though it would be simple to find an example, but I also see no use for lexical error recovering. As long as I feel, I am very happy with lark as it is. I can only say parsing theory is very a hard subject, and requires a very hard study and understanding to write an good parser. Thank you very much for writing such marvelous parser with your free time. Sorry for wasting your time reviewing my code for this error recovering thing.

For now, I will be closing these pull requests for error recovering. If someday I really find some use for this error recovering, this code will be saved here for us to use and continue from it.

erezsh:
No worries. You're right, it's a difficult subject!

I do have some ideas for error recovery, so perhaps I will try them when I have the time.

PR #297, opened 2018-12-22
Added tox.ini posargs option, you can pass arguments to unit tests

evandrocoan:
https://stackoverflow.com/questions/35697353/how-to-run-only-one-test-in-tox


evandrocoan:
The difference, it that now I can call tox like this to run a single test:
```python
tox -e py27 -- TestClass.test_name
```

PR #298, opened 2018-12-22
Documented on README.md how to run the Unit Tests with tox

evandrocoan:


erezsh:
As I mentioned in your other PR, that's actually not the recommended way to run the tests.

erezsh:
Sure, I'm all for docs, when they're correct :)

erezsh:
But, it probably doesn't belong on the README

It would be good to have somewhere in the docs

erezsh:
This is nice! But it's still incorrect that you need Tox to run the tests. I'm fine if that's provided as another option, but really all you need, and should be the default, is:

    python -m tests

Another way is

   python setup.py test    # or something like that

It's also possible using pytest.

erezsh:
Merged! Thanks and sorry for the delay

PR #317, opened 2019-02-01
start with empty callbacks, then assign UnlessCallback

jettero:
If I'm to supply lexer_callbacks={???} then they can't all be predefined as UnlessCallback and then also check assert type_ not in self.callbacks

https://lark-parser.readthedocs.io/en/latest/recipes/#recipes

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -270,12 +270,13 @@ class TraditionalLexer(Lexer):
 
         terminals.sort(key=lambda x:(-x.priority, -x.pattern.max_width, -len(x.pattern.value), x.name))
 
-        terminals, self.callback = _create_unless(terminals)
-        assert all(self.callback.values())
-
+        self.callback = dict()
         for type_, f in user_callbacks.items():
             assert type_ not in self.callback
             self.callback[type_] = f
+
+        terminals, self.callback = _create_unless(terminals)
+        assert all(self.callback.values())
 
         self.terminals = terminals
 
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -270,13 +270,12 @@ class TraditionalLexer(Lexer):
 
         terminals.sort(key=lambda x:(-x.priority, -x.pattern.max_width, -len(x.pattern.value), x.name))
 
-        self.callback = dict()
-        for type_, f in user_callbacks.items():
-            assert type_ not in self.callback
-            self.callback[type_] = f
-
         terminals, self.callback = _create_unless(terminals)
         assert all(self.callback.values())
+
+        for type_, f in user_callbacks.items():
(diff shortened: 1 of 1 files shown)


jettero:
actually, my solution solves nothing... fixing

jettero:
Actually, it seems the problem is that my code is creating more than one parser object ... I'm supplying lexer_callbacks to each, but there's some kind of collision on the second TraditionalLexer() instance... this solution may not be a great idea based on realizing that.

erezsh:
Hi @jettero , can you explain what problem you're trying to solve? A small example of input and expected behavior are usually the best explanation.

jettero:
nope. I can't really provide a sample. I've found no way to do what I actually want to do...

Suppose I have a list of a dozens of python objects with a .name property. Some are swords and some are people. They have various context permissions properties (eg, `obj.can_verb_obj_with_obj`) — more or less how the Lima MUD parser worked (circa 1995).  I want the lexer or the parser to compare the given word characters to the names and accept the token if the name matches and its various context methods allow the object to be in that context.

Feels like the "code separate from parse" philosophy line item is causing me trouble. I do have this working as a pure text parse and I try to do some of the above in the transformer (or after). It just makes the parser a little dumb, as it can never tell which words work in which contexts.

jettero:
Ahh, I got it. It's not making more than of the Lark()s it's the other patterns the UnlessCallback is trying to cover... the below will trip the error. If you comment out the `another_thing` rule it'll run fine.

```
import lark

class SomeFancyObject:
    pass

lookup_table = {
    'b1': SomeFancyObject(),
    'b2': SomeFancyObject(),
}

def do_lookup(tok):
    if tok in lookup_table:
        tok.value = lookup_table[tok]
        return tok

grammar = '''
%import common.WS
%ignore WS
?start: BLAH1 | BLAH2
BLAH1: WORD
BLAH2: WORD
WORD: /\w+/
another_thing: "supz" | "whatever"
'''

import logging
logging.basicConfig(level=10)
log = logging.getLogger(__name__)
log.debug('supz')

class Transformer(lark.Transformer):
    pass

p = lark.Lark(grammar, parser='lalr', debug=True,
 transformer=Transformer(), lexer_callbacks={'BLAH1': do_lookup})

print( p.parse('b1') )
print( p.parse('b2') )
```

erezsh:
Good find! I added callbacks as an afterthought, and didn't think about this use-case, but it makes sense.

Someone even suggested that I add this ability to change tokens to objects at the Transformer level, but I think for now callbacks are the right place for it.

I pushed a fix to `master`, commit `1b5f734`

Let me know if it solves your issue.

erezsh:
Whoops, I forgot to push it! 

It's in `master` now, let me know if it works.

jettero:
Yes, that seems to do the trick. Thanks.

PR #322, opened 2019-02-12
Support hex escape (\xhh) in string literal

tomyun:
I'm trying to parse some binary opcodes with Lark and quite satisfied with what it offers so far. A minor gripe I had was limited support for string literal escape in Lark grammar.

Let's say I want to capture `\x01` in a string with a rule with embedded literal, while not keeping its terminal node. Introducing non-capturing regex literal (i.e. `_A: /x01/`) was not an option since otherwise grammar would get too complicated.

Then only left with some partial solutions:

```
A: /\x01/ ... // string literal always captured
B: "\u0001" ... // not captured, but verbose and maybe confusing to represent a byte
C: /\u0001/ ... // captured, confusing?
```

Here I want to directly use hex escape in my string literal like this:

```
D: "\x01" // string literal not captured, clearly representing a byte
```

It can be possible with a very little change in `_fix_escaping()` although I have no idea if there was any other reasons it didn't have one on the first place.

Maybe we could also add other [escape sequences](https://docs.python.org/3/reference/lexical_analysis.html) (i.e. `\ooo` for octal) if interested.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -346,7 +346,7 @@ def _fix_escaping(s):
             n2 = next(i)
             if n2 == '\\':
                 w += '\\\\'
-            elif n2 not in 'unftr':
+            elif n2 not in 'uxnftr':
                 w += '\\'
             w += n2
     w = w.replace('\\"', '"').replace("'", "\\'")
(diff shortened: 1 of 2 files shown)


erezsh:
Thanks for the PR!

Issue #148, opened 2018-05-20
Extend %import functionality: Allow to import rules


erezsh:
Requires recursive import of dependent rules and terminals.

Could cause collisions, at least in LALR. Do we merge or throw an error?

erezsh:
An experimental implementation was added at commit 21bbf9b.

There's no support yet for collisions caused by importing the same file from two or more different imported files.

Rogdham:
Hello, I would also be interested in importing rules. My use case is to split the grammar into several files, but maybe there is some easier way to achieve that?

Anyway, I would expect:
 * rule names to belong to one file only (for example they could be renamed to avoid collisions)
 * `%ignore` to apply to a file only

---

Example:

 * `a.lark`
```
start: expr

expr: "ab"
    | "a" expr "b"

%ignore "_"
```
 * `b.lark`
```
start: expr

expr: "X"
    | "[" inner "]"

%import a.start -> inner
```

The grammar defined in `b.lark` should match: `X`, `[aabb]` and `[aab_b__]` but not `_[ab]_` (`_` not ignored in `b.lark`) nor `[aXb]` (the scope of the `expr` rule should be restricted to each file)

---

@erezsh you mention having written an experimental implementation.

I see that  `import_from_grammar_into_namespace` is in `master`, but I'm not sure how to use it.

erezsh:
@Rogdham 

> * rule names to belong to one file only (for example they could be renamed to avoid collisions)

If I understand you correct, that's already implemented

> * `%ignore` to apply to a file only

`%ignore` isn't imported, because it can't be scoped. At least, it would be complicated to implement.

> I'm not sure how to use it.

I'm pretty sure it's already in `0.6.6`.

Rogdham:
> > * `%ignore` to apply to a file only
> 
> `%ignore` isn't imported, because it can't be scoped. At least, it would be complicated to implement.

Fair enough!

> > * rule names to belong to one file only (for example they could be renamed to avoid collisions)
> 
> If I understand you correct, that's already implemented

It seems that you are right indeed! :+1: 

However, I was not able to "rename" an imported rule… I am able to do `%import .a.expr`, but say I already have a rule `expr` defined in the file I am importing into, I would like to either do `%import .a.expr -> foobar`, or something like `%import .a` and then use `.a.expr` as a rule name (maybe it needs its own syntax).

[...]

PR #325, opened 2019-02-20
Add rule renaming (and fix import bug)

Rogdham:
Following discussion in #148, this PR adds the syntax `%import .local.file.rule -> foobar` allowing to rename a rule while importing it.

While writing tests, I've encountered a bug: `%import .local.file.rule` internally renames the rule as `local.file.rule`, but this is not a valid group identifier for Python, which makes `lexer._build_mres` crash. I've changed so that the named rule is `local__file__rule` to avoid the issue.

Also added some more doc, tests, examples.

Please review carefully as I don't know your codebase, so I may have broken things inadvertently!


erezsh:
Other than that it looks good!

Rogdham:
> * All tests that perform relative imports should use a .lark file (see `test_relative_import`), so that it works with pytest and other test libraries.

Could you tell me what I should change? The tests performing relative imports that I wrote are using the `tests/grammars/ab.lark` file I believe.

> * The since you have several (4) files for the same example, it's better that you create a directory inside examples specifically for that example (or group of examples of the same sort).

I've edited my commit.

erezsh:
For example `test_relative_import_rename`

Rogdham:
> For example `test_relative_import_rename`

That test imports from the file `tests/grammars/test.lark` due to [this line](https://github.com/lark-parser/lark/pull/325/files#diff-ed9d66898a1188fc8f00f74cc941be36R1021):
`%import .grammars.test.NUMBER -> N`

I'm not sure I understand, could you tell me what I should do instead?

erezsh:
Just put the string in a separate .lark file, and use _Lark_open, otherwise the relative import won't always be relative to the right directory

Rogdham:
Oh I see now, like the code examples that have been pushed to `master` before I did this PR.

I tried to change that, and I have rebased on `master` as well, tell me what you think!

PR #327, opened 2019-02-23
Fix literal range escape assertion

tomyun:
Currently string literal range syntax doesn't work for unicode/hex escaped string because it assumes the length of literal always one which is not true for those escaped strings. It can be simply fixed by  using `_fix_escaping()` on `len()` although assert now looks a bit busy.


erezsh:
Looks good, thanks!

PR #352, opened 2019-04-09
Run tests against Python 3.7 and pypy 6.0

michael-k:


michael-k:
Xenial (16.04 LTS) and trusty (14.04 LTS) are Ubuntu releases, see https://en.wikipedia.org/wiki/Ubuntu_version_history.

On travis, trusty is the default, but Python 3.7 requires xenial.  pypy and pypy3 however do not work on xenial (despite its age of almost 3 years; https://travis-ci.com/michael-k/lark/builds/107679057).

This would also work, but I prefer to use xenial (newer) as default:

```yaml
matrix:
  include:
    - python: "3.7"
      dist: xenial
```

Maybe replacing `pypy` and `pypy3` with `pypy2.7-6.0` and `pypy3.5`/`pypy3.5-6.0` would work in combination with xenial.  I've never tried it.  https://travis-ci.community/t/confusing-pypy-versions/2505

erezsh:
Great. So can you update the PR?

erezsh:
Also, doesn't travis use the newest by default?

michael-k:
> Great. So can you update the PR?

Sure; done.  Trying out pypy 6.0 was just one of the last things I did before going home yesterday. :)

> Also, doesn't travis use the newest by default?

No, they are quite conservative (in my eyes) or in other words: they use the most stable by default.  Support for xenial isn't very old.  Here's an article about trusty becoming the default: https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming
They might switch to xenial being the default after trusty reaches its EOL on 2019-04-30.  The next LTS after xenial (Bionic Beaver, 18.04) is [not yet supported](https://travis-ci.community/t/ubuntu-18-04-1-lts-bionic-beaver/1270).

erezsh:
Looks good. Thanks!

Issue #353, opened 2019-04-09
inline rules are not preserved when imported from another file


PJCampi:
So I found another bug related to relative imports from another grammar file. Let me know if you are not interested in these and I'll just update in my fork of your repo.

Consider the following example:
a.lark
```
A: "A"
B: "B"

a: A
_b: B

c: _b
```

b.lark
```
start: a " " c

%import .a (a, c)
```

main.py
```
from lark import Lark
parser = Lark.open('grammar/b.lark', rel_to=__file__)
print(parser.parse('A B'))
```

`_b` will fail to inline because it is renamed to `a_ _ _b` instead of `_a_ _b`.

This is easily fixed by updating `get_namespace_name` to:

```
    def get_namespace_name(name):
        try:
            return aliases[name].value
        except KeyError:
            if name[0] == '_':
                return '_%s__%s' % (namespace, name[1:])
            else:
                return '%s__%s' % (namespace, name)
```

erezsh:
The looks like a reasonable solution. Mind creating a PR for that? A test-case would also be appreciated.

PJCampi:
sure, I will do tomorrow.

PR #356, opened 2019-04-10
get_namespace_name now preserves leading underscore on relative import

PJCampi:
fixes: https://github.com/lark-parser/lark/issues/353


erezsh:
Thanks!

Issue #347, opened 2019-04-05
File not found error in nested imports


PJCampi:
Hi there,

First off thanks a lot for the hard work on the package!

I think I have come across a bug in the grammar parsing when there are more than one nested imports. 

Consider the following folder structure:
```
grammar/
 -- a.lark
 -- b.lark
 -- c.lark
main.py
```

with files of the following content:

a.lark
```
start: FOO

FOO: "foo"
```

b.lark
```
start: FOO BAR

BAR: "bar"

%import .a.FOO
```

c.lark
```
start: BAR BAZ

BAZ: "baz"

%import .b.BAR
```

main.py
```
from lark.lark import Lark

parser = Lark.open('grammar/c.lark', rel_to=__file__)
```

Instantiating the parser causes the following exception:
_FileNotFoundError: [Errno 2] No such file or directory: 'b.lark'_

I think that the error is due to the fact that the recursive call to `load_grammar` in `import_grammar` does not include the `base_path` that is needed for nested imports.
It's available for the first level of nesting (i.e. we can create a parser from b.lark) but it's no longer there if you have two recursive import like in the example above (c.lark imports from b.lark that imports from a.lark).

This could be fixed by changing the logic in `import_grammar` from:
```
                grammar = load_grammar(text, grammar_path)
```

to:
```
                if import_path in base_paths:
                    recursive_path = os.path.join(import_path, grammar_path)
                else:
                    recursive_path = grammar_path
                grammar = load_grammar(text, recursive_path)
```

That would ensure that we keep the path information that is needed for the nested import.

I am happy to make a pull request for this if you want. Let me know what you think.

erezsh:
I'll be happy to accept a PR for this, and as usual with a test-case if possible.

But I think the solution is different and simpler:

```python
                joined_path = os.path.join(import_path, grammar_path)
                with open(joined_path) as f:
                    text = f.read()
                grammar = load_grammar(text, joined_path )  # or maybe dir_path(joined_path)
```

But it's been a while since I wrote this code, so maybe I missed something about it.

PR #357, opened 2019-04-10
import_grammar now include base_path in recursive call to load_grammar

PJCampi:
fixes https://github.com/lark-parser/lark/issues/347


erezsh:
Looks good. Thanks!

PR #360, opened 2019-04-12
load_grammar now collects all imports to make before loading them to …

### LINKED ISSUE #348: False Terminal defined more than once error
Consider the following example:
a.lark
```
%import common.INT

a: INT "A"
b: INT "B"
```

b.lark
```

start: a | b

%import .a.a
%import .a.b
```

main.py
```
from lark.lark import Lark

parser = Lark.open('grammar/b.lark', rel_to=__file__)
print(parser.parse('A'))

```
This fails with error:
`lark.exceptions.GrammarError: Terminal 'a__INT' defined more than once`

that's because the tokens from the grammar `a.lark` are added to the namespace on every rule import from a.

One can work around by ensure that rules are either aliased or imported only once like that. That's also more efficient (as the grammar is loaded only once):
```
start: a | b

%import .a (a, b)
```
but this could also possibly be fixed by ensuring that terminals from the same imported grammar are imported to the namespace only once (in `new_td`)?

PJCampi:
fixes #348

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -725,7 +725,7 @@ class GrammarLoader:
         rule_defs = [options_from_rule(*x) for x in rule_defs]
 
         # Execute statements
-        ignore = []
+        ignore, imports = [], {}
         for (stmt,) in statements:
             if stmt.data == 'ignore':
                 t ,= stmt.children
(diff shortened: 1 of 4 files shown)


erezsh:
Looks good. Thanks!

PR #363, opened 2019-04-16
Refactor code accordence to PEP8 and use newer str.format method

Asocia:
I replaced all % formats with newer str.format method as is supported since python2.6+

___________________________________ summary ____________________________________
  py27: commands succeeded
SKIPPED:  py34: InterpreterNotFound: python3.4
  py35: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
  pypy: commands succeeded
  pypy3: commands succeeded
  congratulations :)


erezsh:
Sorry,  but I don't really see the point..

Asocia:
If you are asking for string formatting, I would use it because it is newer, more capable than % formatting and less probable to be deprecated in future versions of python. Other than that the only difference is adding/removing newlines to follow the guideline as explained here [https://www.python.org/dev/peps/pep-0008/#blank-lines](https://www.python.org/dev/peps/pep-0008/#blank-lines)

Asocia:
@erezsh What's your thoughts about this PR? Are you planning to merge this?

erezsh:
Sorry, this doesn't seem like a valuable change. I hope you'll find a way to contribute that improves the functionality or accessibility of the library.

Asocia:
OK, thanks

PR #365, opened 2019-04-16
Bugfix/364 reconstruction update start symbol of parser instead of recreating it

### LINKED ISSUE #364: reconstructor recreates a new parser at each iteration
Hi,

At present, the reconstructor recreates a new earley parser at each step of the reconstruction. This is an expensive operation because the same grammar is re-analysed every single time.

Would it be possible to create the parser at instantiation of the reconstructor and only change the `start` value of the parser afterwards? 
Something like:
```
   def _reconstruct(self, tree):
        self.parser.parser_conf.start = tree.data
        unreduced_tree = self.parser.parse(tree.children)   # find a full derivation
        assert unreduced_tree.data == tree.data
        ...
```
I have tested it on my grammar and it improves performance markedly. However there may be a reason against it that I missed?

Let me know if this solution works and I'll make a PR for it.

PJCampi:
fixes #364

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -725,7 +725,7 @@ class GrammarLoader:
         rule_defs = [options_from_rule(*x) for x in rule_defs]
 
         # Execute statements
-        ignore = []
+        ignore, imports = [], {}
         for (stmt,) in statements:
             if stmt.data == 'ignore':
                 t ,= stmt.children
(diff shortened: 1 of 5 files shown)


erezsh:
Why is there a commit for imports? Didn't I already merge this before?

erezsh:
Thanks!

PR #367, opened 2019-04-18
Update JSON Tutorial Code for Python 3

copperlight:
Thanks for writing such a helpful tutorial!

The existing code snippets which demonstrate the Transformer class rely
on tuple parameter unpacking, which was removed as of PEP-3113, which
landed in Python 3.0.

https://www.python.org/dev/peps/pep-3113/

The class methods don't make use of any internal data, so they can be
marked with @staticmethod.

The updated versions of these code snippets continue to work as expected
in Python 2.7.16, noting the difference in how unicode strings are
marked.

The code snippets copy and paste cleanly into the Python repl.

### DIFF
--- a/docs/json_tutorial.md
+++ b/docs/json_tutorial.md
@@ -6,13 +6,14 @@ In this tutorial we will write a JSON parser in Lark, and explore Lark's various
 
 It has 5 parts.
 
-  1. Writing the grammar
-  2. Creating the parser
-  3. Shaping the tree
-  4. Evaluating the tree
-  5. Optimizing
+1. Writing the grammar
+2. Creating the parser
+3. Shaping the tree
+4. Evaluating the tree
+5. Optimizing
 
 Knowledge assumed:
+
 - Using Python
 - A basic understanding of how to use regular expressions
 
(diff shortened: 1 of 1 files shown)


erezsh:
Thank you for the update.

Regarding the tuple unpacking, Lark provides a utility for that  `@v_args(inline=True)`.

I think it's more readable, so it'd be great If you could use that instead.

copperlight:
Updated examples to use the `@v_args` decorator.

Updated the benchmark example at the bottom.

Verified copy-paste in both the Python 3.7.3 and 2.7.16 repls.

drunkwcodes:
Does it mean that rename `v_args(inline=True)` and use `@inline` decorator instead?

copperlight:
@erezsh What do you think of the changes?

The `TreeToJson` transformer could be made more brief, by doing the following:

```
null = None
true = True
false = False
```

However, I am not sure if this helps or hinders understanding.

erezsh:
The changes are good, but the code is a little verbose. After all, this is supposed to be a tutorial for beginners, so being clear and simple is very important, especially in the code. I'll have a look at it soon.

PR #369, opened 2019-04-24
Add vaild_parse method for Lark class.

KmolYuan:
For more simpler statement with `Lark.vaild_parse` to discard tracking exceptions:

```python
tree = g.vaild_parse(expr)
if tree is None:
    return
vaild_operation(tree)
```

Python 3.8 style:

```python
if (tree := g.vaild_parse(expr)) is not None:
    vaild_operation(tree)
```


erezsh:
I'm against adding this method to Lark. It's not idiomatic, and if you really want this style, it's easy to do in your own code.

But thank you for submitting it.

PR #371, opened 2019-05-06
Fix DeprecationWarning in lalr_analysis.py

pvinci:
Under python 3.3+, logging.warn is deprecated.
Use logging.warning instead.

Fixes: /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/lark/parsers/lalr_analysis.py:87: DeprecationWarning: The 'warn' function is deprecated, use 'warning' instead


erezsh:
Thanks!

PR #373, opened 2019-05-09
Allow any graph attribute in `pydot__tree_to_png`.

fbindel:
I needed to increase the dpi of the pydot graph, so I added `**kwargs` to `pydot__tree_to_png` in `tree.py`, as adding a single additional attribute seemed silly to me. To keep compatibility, `rankdir="LR"` is still an explicit default value in the function signature. The docstring has been updated accordingly.


erezsh:
Looks good, thanks!

PR #393, opened 2019-05-29
Clarify handling of filtered terminals

davaya:


erezsh:
Thanks for the PR, I can understand why that example can be useful.

However, this isn't a case of missing documentation: PNAME on a whole is a terminal, so of course it will appear fully, including all its fragments.

If you can rewrite it to reflect this fact, for example as a separate note, with a description such as "Terminals always show their whole match, without omitting any part.", then I will accept this pull request. (and keep your example of course)

Cheers,
Erez

davaya:
Updated with a note.

erezsh:
Looks good, thanks!

PR #401, opened 2019-06-27
Updated parameter syntax in Evaluating the tree section of json_tutorial.md

dmoini:
Parameters in string and number functions in TreeToJson classes in sections 4 and 5 of json_tutorial.md are incorrect. Fixed the examples to have the correct syntax.

### DIFF
--- a/docs/json_tutorial.md
+++ b/docs/json_tutorial.md
@@ -251,9 +251,9 @@ Also, our definitions of list and dict are a bit verbose. We can do better:
 from lark import Transformer
 
 class TreeToJson(Transformer):
-    def string(self, (s,)):
+    def string(self, s):
         return s[1:-1]
-    def number(self, (n,)):
+    def number(self, n):
         return float(n)
 
     list = list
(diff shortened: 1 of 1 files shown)


MegaIng:
This will not work correctly. These functions take a list as parameters, unless they are annotated with `@v_args(inline=True)`. There are two ways to fix this: annotate the class with `v_args`, but then one would have to rework the class. I thing the best option is to just extract the first element of this list. This could be done in one of these ways:

```python
class TreeToJson(Transformer):
    def string(self, children):
        s, = children
        return s[1:-1]

    def number(self, children):
        n, = children
        return float(n)
```
or

```python
class TreeToJson(Transformer):
    def string(self, children):
        return children[0][1:-1]

    def number(self, children):
        return float(children[0])
```

I personally prefer the first style, since it is closer to the original syntax.

erezsh:
I also prefer the first one.

dmoini:
Thank you for the clarification.

Issue #389, opened 2019-05-27
Allow visitors to visit terminals (transformers too)


erezsh:
Provide a way to use transformers & visitors to iterate over terminals, in addition to rules.

It can be used as an alternative to lexer_callbacks.

bilderbuchi:
Is this something that can be implemented quickly? 
Otherwise, if I have to use Earley and its default lexer, the best approach is to convert terminals I have to visit to rules, right?

erezsh:
>Is this something that can be implemented quickly?

Probably yes

> the best approach is to convert terminals I have to transform to (helper) rules

Yep. Also, you don't really have to preserve the terminal if  you know its value.

erezsh:
Lark now contains that option for transformers. You can use it with

    Transformer(visit_tokens=True).transform(...)

bilderbuchi:
Great, thanks!

erezsh:
Can you think of a use case in which it makes sense for Visitors?

I think Transformers are the only mechanic where it is useful.

bilderbuchi:
I can't, but I have not used Visitors yet. 
We are creating Python code from another language (https://gitlab.com/librecube/prototypes/python-pluto), there this is useful to do simple modifications to terminals so that valid Python gets passed on during transformation (e.g. `TRUE` to `True`).

erezsh:
A new issue exists (and an implementation for this is already in master)

Issue #76, opened 2018-01-29
Detect collision in regular expressions when creating a lexer


erezsh:
When using a standard or contextual lexer, colliding regular expressions may produce an incorrect parser, that only produces an error when encountering the obscured terminal in the input text.

For example, the following definitions:

    VAR: /\w+/
    FILENAME: /[\w.]+/

Will cause the standard lexer to always choose one over the other, but will only produce an error when it makes the wrong choice in run-time (and in some cases, will even produce an unintended parse without producing an error).

Ideally, this would be the correct behavior: If regexp1 contains regexp2 completely, use secondary lexing (i.e. the "unless" mechanism that already exists for strings), else throw an error.

Presumably, the solution lies in running an intersection on the NFAs.

Zac-HD:
Happily, there is already a library for this, https://github.com/qntm/greenery, although:

- It doesn't exactly match Python's regex syntax.  I once tried to support that in qntm/greenery#37, but moved on when it became clear that I'd spend more time on adding valiation logic than I wanted to for an evening fun project.  Definitely tractable though.
- Python's regex, and "PCRE" generally, don't always describe regular languages.  Fortunately it's trivial to detect these features from the compiled regex's parse tree, but they still need to be handled by raising an error or bailing out of the attempt to detect collisions.

erezsh:
I found greenery about a year ago, and was excited to use it for Lark. But it had two major problems:

1) As you said, it doesn't work with Python's regexp syntax. And moreso, it doesn't support common additions, such as lookaheads, which I believe are regular.

2) I was very discouraged by its performance. To be fair, greenery tries to construct a new regexp that matches the intersection, while all I need to know is if an intersection exists. Still, non-trivial regexps could take full seconds to evaluate, which is a no-go.

Maybe I misevaluated it, either from a technical or conceptual standpoint, so feel free to correct me. But I'm not interested in adding a dependency unless it can handle, say, over 95% of my users' grammars.

Btw, my interest in this feature isn't just for better error handling. Having this feature in place paves the road to a stable Earley+LALR integration, which, similarly to GLR+LALR integration, can create a huge performance boost for many Earley grammars.

Zac-HD:
I think you're entirely right, which means this might take a new library to solve :confused:

MegaIng:
[...]

PR #416, opened 2019-07-27
Added basic support for TokenVisitor and TokenTransformer (#389)

MegaIng:


erezsh:
Thanks for this PR, and sorry it took me a while to get to it.

It looks really good, except for:

1) There is no test to prove that it actually works, and to try to catch edge cases

2) I think a better interface would be to include it in the regular `Transformer` and `Visitor`, and allow to initialize them with `tokens=True`. It will be less lines of code, will minimize the amount of classes the user needs to know or import, and better allow for future support in providing that transformer to Lark (useful for performance reasons)

MegaIng:
It's ok, I am currently ignoring my responsibilities over at [regex-intersections](https://github.com/lark-parser/lark/issues/76)...

To your points:

1. Yes, I will add a few tests. What edge cases can you thing of? I can currently only think of `%ignored` and '_IGNORED' cases, which are both kinds of token that can not possibly be handled. 
2. What do you mean by initilization? `__init__`? I would not do that because transformer might be initialized at different points in your code. * We could just integrate the API completely into the Transformer. Because we require UPPERCASE function-names, they can not collide with rules. We will just have to  think about the subclasses.

\* This would also be the fourth way how you customize your Transformer/Visitor:
1. Which superclass do you choose?
2. Which function to you implement? (including `__default__`)
3. Which class/function-decorators do you use?
4. (new) Which arguments do you pass to `__init__`?

I already think there are enough places to search for the feature you need. (and there are still some missing, e.g. python keywords as rule names)

erezsh:
1. Just empty trees, throwing Discard, that sort of stuff. There's probably not a lot of room for edge cases here.

2. I don't see why having 4 different "options" for visitors make it worse than 3. That's like saying the Python's class has inheritance, globals, methods, decorators, etc., so let's not add the super() keyword

I would unite Visitor and Transformer if I could, but they operate very differently. But callbacks for Tokens are a minor addition, and breaking it into new classes seems more clunky to me.

MegaIng:
My argument for 2. was that we don't need an option for Token: since we use uppercase-function names, we can just integrate them.

erezsh:
@MegaIng You're right, but

1) It might degrade performance, because it will require an attribute check for every Token, and there would be a lot of them (probably 3 to 5 times as much as tree branches, on average)

2) It might break transformers for existing users, who used upper-case methods because they are "safe", but happened to match a token name

So my plan is to add it as an option, and in the next major version (which isn't far away) to swap the defaults, so those who care about performance can turn it off.

erezsh:
Added the ability to visit tokens in transformers into master. I used your PR as inspiration, so thanks!

PR #418, opened 2019-07-31
Proper LALR parser

Raekye:
The existing parser calculates lookaheads for reductions by looking at the FOLLOW set of the grammar, making it actually an SLR(1) parser. LALR(1) parsers have the same number of states as SLR(1) parsers, but parse a strict superset of SLR(1) grammars, so they are generally preferred.

~~These commits implement the algorithm described in Compilers: Principles, Techniques, and Tools (2nd edition) to (supposedly) efficiently calculate LALR(1) lookaheads.~~ This now implements [DeRemer and Pennello's LALR lookahead algorithm][1], based on PLY and The Theory and Practice of Compiler Writing. It does _not_ generate LR(1) states (which may be orders of magnitudes larger) to merge into LALR(1) states. **Update**: it is now much much faster than the old (dragon book) algorithm.

At a high level, the algorithm:

1. Generates LR(0) itemsets/states (similar to the existing parser, see below for minor change)~~
1. ~~Calculates lookaheads~~
1. ~~Attaches lookaheads to the LR(0) kernels to form LALR(1) kernels~~
1. ~~Generates the LR(1) closure of the LALR(1) kernels to obtain the LALR(1) states (same number as LR(0) states)~~
1. Does some stuff I can't really explain 😅

I tried my best to maintain Lark's coding style and I believe the code is ~~well documented~~ documented wherever I thought possible. My main concern is the naming scheme, particularly the `lr0_*` member variables introduced to `GrammarAnalyzer`. The LALR(1) algorithm I implemented does not use an `$END` token for the root rules. However, I did not want to overwrite the existing member variables as it would affect the Earley parser. **Update**: I can't really explain _why_ the algorithm works, so I couldn't add many insightful comments unfortunately.

The LALR(1) parser can be tested with the following grammar/program:

```
from lark import Lark

grammar = r'''
EQUALS:  "="
STAR: "*"
ID: "id"

s: l EQUALS r | r
l: STAR r | ID
r: l
'''

parser = Lark(grammar, parser='lalr', lexer='standard', start='s')
text = '*id=id'
print(parser.parse(text).pretty())
```

The older parser (with debug enabled) logs a shift/reduce conflict. While shifting produces the right behaviour in this example (which is the default conflict resolution mechanism), LALR(1) parsers are known to be more powerful than SLR(1) parsers. As it is, users may see shift/reduce or reduce/reduce conflicts with proper LALR(1) grammars, if the grammars are not SLR(1).

I have run the tests and only 3 tests in `tests.test_nearley.test_nearley.TestNearley` fail, which were failing for me before my changes (they have a `FileNotFoundError`, so I assume I'm just missing some test files). However, I have not profiled these changes. Theoretically, parsing speed should be the same because the parsing table format and algorithm were only minimally changed (removed the artificial `$END` token appended to root rules. Now, a successful parse ends in a reduce action). Generating the parser may be slower as the lookahead calculation is more complex compared to SLR(1).

[1]: https://dl.acm.org/citation.cfm?id=357187

### DIFF
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -3,13 +3,12 @@ from .utils import Serialize
 ###{standalone
 
 class Symbol(Serialize):
-    __slots__ = ('name', '_hash')
+    __slots__ = ('name',)
 
     is_term = NotImplemented
 
     def __init__(self, name):
         self.name = name
-        self._hash = hash(self.name)
 
     def __eq__(self, other):
         assert isinstance(other, Symbol), other
@@ -19,7 +18,7 @@ class Symbol(Serialize):
         return not (self == other)
 
     def __hash__(self):
-        return self._hash
+        return hash(self.name)
 
     def __repr__(self):
         return '%s(%r)' % (type(self).__name__, self.name)
@@ -28,13 +27,12 @@ class Symbol(Serialize):
 
 
 class Terminal(Symbol):
-    __serialize_fields__ = 'name', 'filter_out', '_hash'
+    __serialize_fields__ = 'name', 'filter_out'
 
     is_term = True
 
     def __init__(self, name, filter_out=False):
         self.name = name
-        self._hash = hash(self.name)
         self.filter_out = filter_out
 
     @property
@@ -44,7 +42,7 @@ class Terminal(Symbol):
 
 
 class NonTerminal(Symbol):
-    __serialize_fields__ = 'name', '_hash'
+    __serialize_fields__ = 'name',
 
     is_term = False
 
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -3,10 +3,13 @@ from .utils import Serialize
 ###{standalone
 
 class Symbol(Serialize):
+    __slots__ = ('name', '_hash')
+
     is_term = NotImplemented
 
     def __init__(self, name):
         self.name = name
+        self._hash = hash(self.name)
 
     def __eq__(self, other):
         assert isinstance(other, Symbol), other
@@ -16,7 +19,7 @@ class Symbol(Serialize):
         return not (self == other)
 
     def __hash__(self):
-        return hash(self.name)
+        return self._hash
 
     def __repr__(self):
         return '%s(%r)' % (type(self).__name__, self.name)
@@ -31,6 +34,7 @@ class Terminal(Symbol):
 
     def __init__(self, name, filter_out=False):
         self.name = name
+        self._hash = hash(self.name)
         self.filter_out = filter_out
 
     @property
@@ -69,7 +73,7 @@ class Rule(Serialize):
         expansion : a list of symbols
         order : index of this expansion amongst all rules of the same name
     """
-    __slots__ = ('origin', 'expansion', 'alias', 'options', 'order', '_hash')
+    __slots__ = ('origin', 'expansion', 'alias', 'options', 'order', '_hash', '_rp')
 
     __serialize_fields__ = 'origin', 'expansion', 'order', 'alias', 'options'
     __serialize_namespace__ = Terminal, NonTerminal, RuleOptions
@@ -81,6 +85,7 @@ class Rule(Serialize):
         self.order = order
         self.options = options
         self._hash = hash((self.origin, tuple(self.expansion)))
+        self._rp = None
 
     def _deserialize(self):
         self._hash = hash((self.origin, tuple(self.expansion)))
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -71,7 +71,7 @@ class Rule(Serialize):
         expansion : a list of symbols
         order : index of this expansion amongst all rules of the same name
     """
-    __slots__ = ('origin', 'expansion', 'alias', 'options', 'order', '_hash', '_rp')
+    __slots__ = ('origin', 'expansion', 'alias', 'options', 'order', '_hash')
 
     __serialize_fields__ = 'origin', 'expansion', 'order', 'alias', 'options'
     __serialize_namespace__ = Terminal, NonTerminal, RuleOptions
@@ -83,7 +83,6 @@ class Rule(Serialize):
         self.order = order
         self.options = options
         self._hash = hash((self.origin, tuple(self.expansion)))
-        self._rp = None
 
     def _deserialize(self):
         self._hash = hash((self.origin, tuple(self.expansion)))
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -28,7 +28,7 @@ class Symbol(Serialize):
 
 
 class Terminal(Symbol):
-    __serialize_fields__ = 'name', 'filter_out'
+    __serialize_fields__ = 'name', 'filter_out', '_hash'
 
     is_term = True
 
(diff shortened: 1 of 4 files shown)


erezsh:
Great PR! I like how you maintained the coding style, and wrote code that's clean and understandable.

The only thing missing is a test case, that would that it does indeed improve the parser, and prevent future degradation.

There is also another problem, of performance. It's probably not noticeable in tiny grammars, but for medium grammars like `examples/python3.lark`, it's very noticeable. Here's the performance of the latest master:

```bash
$ time python3 -m examples.python_parser
test_python_lib (170 files), time: 31.67557168006897 secs

real    0m36.733s
user    0m33.656s
sys     0m0.422s
```
That means that compiling the grammar took about 3 seconds. Still much more than desirable, but somewhat acceptable.

Here is the performance of your PR:

```bash
test_python_lib (170 files), time: 34.119346380233765 secs
real    1m16.430s
user    1m11.047s
sys     0m0.734s
```

Compiling the grammar took 42 seconds!! This is serious degradation, which I would be very cautious to spring on users.

My point is, perhaps there are simple optimizations you could use to bring this down to single-digits?  After all, if PLY can do it, then I'm sure it's not that hard :)

Raekye:
For sure I'll look into it! I was mindful of avoiding unnecessary work/stuff that had obviously bad time complexity, but I don't claim to be a performance guru and I wasn't specifically trying to optimize the code. I will profile the code and see what I can do

erezsh:
This is looking much much better!

I cleaned up the code a little (many of the hash tricks you added are no longer necessary, and I simplified what I could), and pushed it to the `true_lalr` branch. I'm going to let it sit there for a bit, maybe try to improve the code a little further (mostly just try to make it more readable), and I think soon it will be ready to merge into `master`.

erezsh:
If you feel like working on `true_lalr` and improving it, either by simplifying, documenting, or any other way, feel free to send a PR for it. I can also just give you push permissions for that branch, which would be a little simpler as a workflow.

Raekye:
I remembered something/thought of something minor to improve the code last night while falling asleep haha. Instead of checking `s not in self.rules_by_origin` to see if it's a terminal, you can just check `.is_term`. Lines [190](https://github.com/lark-parser/lark/blob/true_lalr/lark/parsers/lalr_analysis.py#L190) and [205](https://github.com/lark-parser/lark/blob/true_lalr/lark/parsers/lalr_analysis.py#L205) in `lalr_analysis.py`.

I'd be happy to keep looking at it, to improve it. Push permissions for that branch would be nice, though I don't mind updating this/sending another PR

erezsh:
@Raekye My bad, github doesn't allow such granularity. So if you don't mind sending PRs, let's stick with that for now :)

bilderbuchi:
What GH can do, though, if I remember correctly, is to "allow edits by maintainers" when you open a PR (there's a box to check), so @erezsh could edit the PR branch, which solves the present problem I think.

Raekye:
I have the "allow edits from maintainers" box checked. Although it might be awkward, since erezsh already merged this into a new branch and made commits to it

erezsh:
Okay, I added my cleanup commits. You can edit on top of them now.

evandrocoan:
You can get detailed performance per function call with this snippet. Edit the lark main and add the setup to the beginning and the result printing to the bottom. Optimally, lark could have a command line option which activates this on its main.

```python
import sys

import io
import pstats
import cProfile

profiller = cProfile.Profile()
profiller.enable()

# Core here

profiller.disable()
outputstream = io.StringIO()

profiller_status = pstats.Stats( profiller, stream=outputstream )
profiller_status.sort_stats( "time" )
profiller_status.print_stats()

sys.stderr.write( outputstream.getvalue()[:2000] + '\n' )
```

Running this solemnly will output:
```
         1 function calls in 0.000 seconds

   Ordered by: internal time

      ncalls     tottime     percall     cumtime     percall filename:lineno(function)
           1       0.000       0.000       0.000       0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

Adding a few cool stuff inside it like:
```python
import mmap
import json

# https://docs.python.org/3.9/library/mmap.html
# https://blog.askesis.pl/post/2019/02/mmap.html
memory = mmap.mmap( -1, 20, access=mmap.ACCESS_WRITE )
memory.write( b'{ "var": 10 }' )

end_of_the_file = memory.tell()
print( 'memory {:,}, end_of_the_file {}\n'.format( len( memory ), end_of_the_file ) )

memory.seek( 0 )
print( 'raw', memory.read( end_of_the_file ), file=sys.stderr )

memory.seek( 0 )
loaded = json.loads( memory.read( end_of_the_file ) )

print( 'json', loaded, file=sys.stderr )
print( 'loaded', memory[:end_of_the_file], file=sys.stderr )
```

Will output:
```
         3,031 function calls (2,921 primitive calls) in 0.006 seconds

   Ordered by: internal time

      ncalls     tottime     percall     cumtime     percall filename:lineno(function)
          17       0.001       0.000       0.001       0.000 {built-in method nt.stat}
        15/6       0.000       0.000       0.001       0.000 F:\Python\lib\sre_parse.py:475(_parse)
           4       0.000       0.000       0.000       0.000 {built-in method marshal.loads}
           4       0.000       0.000       0.000       0.000 <frozen importlib._bootstrap_external>:914(get_data)
           4       0.000       0.000       0.001       0.000 <frozen importlib._bootstrap_external>:793(get_code)
        25/6       0.000       0.000       0.000       0.000 F:\Python\lib\sre_compile.py:71(_compile)
           4       0.000       0.000       0.000       0.000 {method 'read' of '_io.FileIO' objects}
          14       0.000       0.000       0.000       0.000 F:\Python\lib\sre_compile.py:276(_optimize_charset)
           6       0.000       0.000       0.001       0.000 <frozen importlib._bootstrap_external>:1356(find_spec)
           4       0.000       0.000       0.000       0.000 {built-in method builtins.print}
         102       0.000       0.000       0.000       0.000 F:\Python\lib\sre_parse.py:233(__next)
       31/12       0.000       0.000       0.000       0.000 F:\Python\lib\sre_parse.py:174(getwidth)
           3       0.000       0.000       0.000       0.000 {built-in method builtins.__build_class__}
           2       0.000       0.000       0.000       0.000 {built-in method _imp.create_builtin}
         6/2       0.000       0.000       0.006       0.003 <frozen importlib._bootstrap>:978(_find_and_load)
         123       0.000       0.000       0.000       0.000 F:\Python\lib\sre_parse.py:164(__getitem__)
          39       0.000       0.000       0.000       0.000 {method 'format' of 'str' objects}
           1       0.000       0.000       0.000       0.00
```

erezsh:
Just merged it into `master`. After an appropriate waiting period I plan to release it with `0.8`

@Raekye thank you for a great contribution, and sorry it took so long!

erezsh:
P.S. I plan to mention you in the next release. Would like to add a few words?

Raekye:
@erezsh Hi! No worries, sorry for the late response myself. Would it be alright to add my username and email (libraekye at gmail dot com) to `lalr_analysis.py`? Totally minor thing, no worries if you prefer not. Other than that, I can't think of anything at this time. Thanks for your work on this project and time reviewing the PR!

erezsh:
@Raekye Yes, that's completely acceptable. And let me know if you think of anything else.

PR #423, opened 2019-08-09
Find newlines in ignored tokens

### LINKED ISSUE #420: Add \s to lark.lexer._regexp_has_newline
This is a trivial fix: _regexp_has_newline does not test for \s patterns in the regex, so if you use it in your grammar, Lark will report wrong line numbers. Should I send a PR?

fabiommendes:
Fix #420

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /lark_parser.egg-info/**
 tags
 .vscode
+.idea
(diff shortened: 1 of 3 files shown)


erezsh:
Close one, the test was incorrect, but I fixed it.

Thank you for contributing!

PR #427, opened 2019-08-18
Add debug=True|False flag to Earley and XEarley to allow dumping the SPPF to PNG.

night199uk:
This re-uses the debug=True|False flag used for LALR parser to enable dumping the SPPF to a PNG using pydot/graphviz, which is very useful in debugging Earley.  This could possibly be moved to a different flag later, but debug=X is currently unused by Earley so it seemed better than adding yet more options.


erezsh:
Looks good to me!

Issue #419, opened 2019-08-01
Collecting examples of inconsistent parsing by Earley (non-deterministic results)


jtrakk:
@erezsh [has indicated](https://github.com/lark-parser/lark/issues/343#issuecomment-517050980) it'd be helpful for us to provide examples of random parsing, and the simpler the better. I'll kick it off.

Please include:

- grammar
- input text
- parse trees
- Python version
- Lark version
- the values of PYTHONHASHSEED that generate different parse trees.

jtrakk:
This is a reduced version of my attempt at emulating [jo](https://github.com/jpmens/jo/blob/master/jo.md) with Lark.

```py
import lark

grammar = r"""
?start: sequence

?value: object
  | SIGNED_NUMBER      -> number
  | "null"             -> null

object : "{" [pair ("," pair)*] "}"
pair   : string ":" value
string : ESCAPED_STRING

?thing : value
       | eqpair
       | eqnull




name : NAME
eqpair : (value|name) "=" value

eqnull : (thing | name ) "="

sequence : thing+



%import common.ESCAPED_STRING
%import common.CNAME -> NAME
%import common.SIGNED_NUMBER
%import common.WS
%ignore WS


"""


parser = lark.Lark(grammar)

text = 'a=1'

print(parser.parse(text).pretty())

```
```bash

PYTHONHASHSEED=1 python example.py
sequence
  eqpair
    name	a
    number	1


PYTHONHASHSEED=3 python example.py
sequence
  eqnull
    name	a
  number	1

```

erezsh:
Another example is given at #388: 

```python
from lark import Lark

numbergrammar = """
start: number*
number: hex_number | decimal_number
hex_number.100: size? HEX_BASE HEX_VALUE
decimal_number: UNSIGNED_NUMBER
size.100: NON_ZERO_UNSIGNED_NUMBER
NON_ZERO_UNSIGNED_NUMBER: /[1-9][0-9_]*/
FIXED_POINT_NUMBER: /([0-9][0-9_]*\.[0-9][0-9_]*)/
UNSIGNED_NUMBER: /([0-9][0-9_]*)/
HEX_VALUE: /([0-9a-fA-FxXzZ][0-9a-fA-F_xXzZ]*)/
HEX_BASE: /'([sS])?[hH]/

%import common.WS
%ignore WS
"""

testparser = Lark(numbergrammar)

test_numbers = """
16'habcd
"""

[...]

PR #428, opened 2019-08-18
Fix Earley determinism

night199uk:
The way we were calculating the tie-breaker for Earley was wrong. Rule.order should be the index of the specific rule expansion amount all of the expansions, not the index of the rule within the file. This allows the grammar author to easily control the tie-breaker within a rule:
a: b
  | c

(b will always be preferred over c if neither b nor c have priorities assigned). Priorities trump expansion ordering.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -511,12 +511,12 @@ class Grammar:
 
         simplify_rule = SimplifyRule_Visitor()
         compiled_rules = []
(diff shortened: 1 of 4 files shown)


night199uk:
Fixes #419.

erezsh:
Such a simple oversight! Even I should have caught it.

Anyway, fantastic, thanks for the fix!

Issue #372, opened 2019-05-09
priorities in LALR and Earley


PJCampi:
Hey,

Is it correctly understood that priority can be set on rules only for earley and on terminals only for lalr? 

I expected to be able to switch seamlessly between the two algorigthms for my unambiguous and non-recursive grammar but I get a lot of parsing errors when switching from earley to larl. I think that most of them are due to priorities on rules not understood by lalr.

Also btw there may be some inconsistency in the index of priority between the two algorithm.

This grammar parses '1' to b with earley:
```
%import common.INT

start: a | b

a: INT
b.1: INT
```

while it resolves to A with lalr:
```
%import common.INT

start: A | B

A: INT
B.1: INT
```

One has to increment to B.2 to get the grammar to parse '1' to B.

erezsh:
Yes, there is a difference in how the two algorithms handle priority. It's not impossible to fix, it's just a lot of work.

Regarding your example, `B.1` is the same as `B`, because 1 is the default priority. So yeah, you need to use `B.2` or higher.

This is the same for Earley. The reason Earley resolves this as `a`, is because when priorities are even (that is, both 1), it determines according to the order of appearance. In this case, `a` before `b`.

PJCampi:
Ok thanks for the info.

Note however that earley resolves as `b` if both priorities are equal, not `a`. That's why I believed that the index of a rule without priority was 0.

night199uk:
That's odd. The logic for the Earley prioritisation is done through the sort_key function on PackedNode. Essentially, each ambiguity in an ambiguous derivation is held under a PackedNode child of the SymbolNode or IntermediateNode. I don't have time to dig in right now, but here are some breadcrumbs for anyone troubleshooting: 

The PackedNode children of a SymbolNode are sorted before returning them:
```
    @property
    def children(self):
        return sorted(self._children, key=attrgetter('sort_key'))
```

sort_key on PackedNode is this:
```
    @property
    def sort_key(self):
        """
        Used to sort PackedNode children of SymbolNodes.
        A SymbolNode has multiple PackedNodes if it matched
        ambiguously. Hence, we use the sort order to identify
        the order in which ambiguous children should be considered.
        """
        return self.is_empty, -self.priority, -self.rule.order
```

The default priority is -inf, and self.rule.order is the ordering of children of the rule (in this case a = 1, b = 2, I think).  
Hence, in your case a = 1, b = 2. So the sort key is:
a: False, inf, -1
b: False, inf, -2

[...]

PR #430, opened 2019-08-20
Tweak the Earley ambiguity algorithm to correctly prefer earlier branches

night199uk:


night199uk:
Fixes the Earley comments on #372.

erezsh:
Thanks :)

PR #431, opened 2019-08-21
fix custom lexer handling for lalr parser + test

michaelheyvaert:
Fixes broken custom lexer construction. The custom lexer code in the examples folder worked because it didn't had an _ _ init _ _ method.

### DIFF
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -118,7 +118,7 @@ class LALR_ContextualLexer(LALR_WithLexer):
 
 class LALR_CustomLexer(LALR_WithLexer):
     def __init__(self, lexer_cls, lexer_conf, parser_conf, options=None):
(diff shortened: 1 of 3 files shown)


erezsh:
Looks good. You could make it a little nicer by doing

    (CustomLexer, 'lalr'),

and then just checking `is CustomLexer` in the `SkipIf`

michaelheyvaert:
That was my original approach, but it introduced more complexity in the test code.

All test conditions would need to be made aware of the fact that LEXER has type Union[str, Lexer], so they need to handle both cases (ie. LEXER.startswith('dynamic') becomes not isinstance(Lexer, type) and LEXER.startswith('dynamic')).

I could wrap those conditions in a few helper functions if you prefer that direction.

erezsh:
Okay, that's a good point. Thanks for the PR!

PR #432, opened 2019-08-22
Fix for staticmethods in Transformers + examples, docs

night199uk:
If you use the pattern x = staticmethod(lambda: True) or similar
with v_args annotations at a class level; the staticmethod may not
be correctly wrapped; due to:
https://stackoverflow.com/questions/45375944/python-static-method-is-not-always-callable


erezsh:
This PR will have to be submitted again (if it's not solved already)

PR #452, opened 2019-09-22
Fix SyntaxError with Python 3 in JSON tutorial

### LINKED ISSUE #403: Syntax errors in docs (json tutorial)
https://github.com/lark-parser/lark/blob/master/docs/json_tutorial.md
There are syntax errors in function signatures like so:
`def string(self, (s,)): ...`
Looks like someone is programming in a language with pattern matching) Or am I missing something?

wataash:
Unpacking a tuple argument is invalid in Python 3.

```py
>>> def foo(a, (b, c)):
  File "<stdin>", line 1
    def foo(a, (b, c)):
               ^
SyntaxError: invalid syntax
```

Fixes #403

### DIFF
--- a/docs/json_tutorial.md
+++ b/docs/json_tutorial.md
@@ -230,7 +230,8 @@ from lark import Transformer
 class MyTransformer(Transformer):
     def list(self, items):
         return list(items)
-    def pair(self, (k,v)):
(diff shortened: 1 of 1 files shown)


erezsh:
Thanks

PR #471, opened 2019-11-03
page for transformer_and_visitor

swuecho:
I find the doc of transformer_and_visitor are very helpful, besides the grammar reference. However, the doc hide from me.  

also, I believe http://blog.erezsh.com/how-to-write-a-dsl-in-python-with-lark/ should be included in the doc directly. links tend to become obsolete and outdated (assuming no copyright problem)


erezsh:
Thanks, it was a nice idea!

My blog post is already in the docs, on the main page (index)

PR #482, opened 2019-11-18
Allow comments in rule definitions

mrob95:
Fixes a bug whereby adding comments in the middle of multi line rule definitions causes a parsing error.

Example:
```
from lark import Lark

grammar = """
start: example+

example: "a"
  // Comment
  | "b"
"""

l = Lark(grammar)
```

Gives:
```
lark.exceptions.UnexpectedToken: Unexpected token Token(_OR, '|') at line 6, column 3.
```


erezsh:
Thanks for the PR!

Issue #523, opened 2020-02-13
PEP 561 typed package


KmolYuan:
Hello! Do you want Lark-parser to have type annotations?
A typed API will helpful to developing more convenience. (like PyCharm)
There will be some generic types instead of `Any` in the function signatures.

Some "public" names can be supported by **stub files** (`pyi`) that will not change your source code directly.
And MyPy checker will be able to executed in latest Python to ignore syntax and module compatibility.

erezsh:
If you are suggesting to send a PR to add pyi files to Lark, then I'm all for it.

PR #524, opened 2020-02-13
Annotations for public API

KmolYuan:
#523

The generic type syntax `Class[T]` is supported by `__class_getitem__` magic method but generate `Class[T] <=> Class` during runtime. (since Python 3.7)

The subclass of `_Decoratable` and `VisitorBase` are supported generic typing. `Transformer` can be typed as:

```python
class T(Transformer[float]):
    ...

t = T()
f = t.transform(tree)  # type: float
```

Some methods not mentioned on documentation are not recorded.

Requires Python 3.8.


lark/tree.pyi
```diff
@@ -0,0 +1,39 @@
...
+from .lexer import Token
+
+
+class Tree:
+
+    data: str
+    children: List[Union[str, Tree]]
+    meta: Token
+
+    def __init__(self, data: str, children: List[Tree], meta: Optional[Token] = None):
```
erezsh: `children` should accept the same type as `Tree.children`

`meta` is not a `Token`, it's `tree.Meta`.

lark/tree.pyi
```diff
@@ -0,0 +1,39 @@
...
+    def find_data(self, data: str) -> Iterator[Tree]:
+        ...
+
+    def iter_subtrees(self) -> Iterator[Tree]:
+        ...
+
+    def iter_subtrees_topdown(self) -> Iterator[Tree]:
+        ...
+
+    def __eq__(self, other: object) -> bool:
```
erezsh: Why is this necessary? No one calls `__eq__` directly. Perhaps there's a reason I'm not aware of?

lark/exceptions.pyi
```diff
@@ -0,0 +1,42 @@
...
+
+
+class LexError(LarkError):
+    pass
+
+
+class UnexpectedInput(LarkError):
+    pos_in_stream: int
+
+    def get_context(self, text: str, span: int = 40):
```
erezsh: Isn't it a bad idea to write the default here? It may change in the future.

Or is it unused?

lark/lark.pyi
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(
+        self,
+        grammar: Union[str, IO[str]],
+        *,
+        start: _Start = ...,
+        parser: Literal["earley", "lalr", "cyk"] = ...,
+        lexer: Optional[Lexer] = ...,
+        transformer: Optional[Transformer] = ...,
+        postlex: Optional[Literal["standard", "contextual"]] = ...,
```
erezsh: `postlex` is a callable object (a lexer proxy)

However
```python
  lexer: Optional[Literal["standard", "contextual", "dynamic", "dynamic_complete"]]
```

lark/exceptions.pyi
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+from typing import Dict, Iterable, Callable
+from .tree import Tree
+
+
+class LarkError(Exception):
```
erezsh: Why are the empty classes required? Does mypy gain any information from this?

lark/lexer.pyi
```diff
@@ -0,0 +1,64 @@
...
+        ...
+
+
+class TerminalDef:
+    name: str
+    pattern: Pattern
+    priority: int
+
+
+class Token(str):
```
erezsh: Missing `Token.value`

KmolYuan:
> Also, is it possible to put all the pyi files in their own folder? I would like the keep the code namespace clean and for code only.

Yes, I can pack them into a dummy module called "`lark-stubs`".
It can work with both development and users.

Sorry about missing some members in stubs, because there are not mentioned in docs.
I'll add them as soon as you instruct.

lark/exceptions.pyi
KmolYuan: MyPy will check inheritance of the classes. Subclass instance is allow to passed as a superclass instance. ([ref](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#class-types))
Is this exception public? If some one want to catch all Lark exceptions.

lark/exceptions.pyi
KmolYuan: Default value can show as `span: int = ...` if it might changed.
Or just hide it if it's a developer option.

lark/lexer.pyi
KmolYuan: Docs not mentioned. I thought it was used to represent itself in `str(self)`.

lark/tree.pyi
KmolYuan: Well, in this case it's unnecessary. Or represent a override method?

lark/tree.pyi
KmolYuan: Sorry, because I saw [doc](https://lark-parser.readthedocs.io/en/latest/classes/#tree) mentioned it has some attributes not listed.
Should I add in `tree.Meta` (as dynamic attributes) or implement as a [`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol)?

erezsh: I'm not sure how Protocol will help. I think it's fine to write `Meta`'s attributes as if they are there, since they usually will be (unless empty is True)

lark/tree.pyi
erezsh: What do you mean?

lark/exceptions.pyi
erezsh: Can't mypy take it from the actual function? Or is there a way to use a CONST from the code (i.e. `DEFAULT_CONTEXT_SPAN`, so there will be one source of truth?

lark/lark.pyi
KmolYuan: What is `postlex` looks like?
I found an `Indenter` instance was used in there.
It's not exposed yet in my work.

lark/exceptions.pyi
erezsh: `LarkError` should be public. So you can either fix that in your patch, or I'll do it in a different patch.

The same is true for most exceptions (most of them are already public)

lark/lexer.pyi
erezsh: It's better to look at the code.

erezsh:
Great, sounds good. `stubs` is the idiomatic word? Because `type-annotations` sounds more informative.

KmolYuan:
@erezsh Yes, a package contains only `pyi` files are named `*-stubs`.
It will be installed with original package and can be found by MyPy.

lark/lark.pyi
erezsh: postlex is a Protocol with the method `process(self, stream: Iterator[Token]) -> Iterator[Token])`

lark/exceptions.pyi
KmolYuan: The actual value is not important in static type checking.
It at most makes the IDE display its preset value.

lark/exceptions.pyi
KmolYuan: I'll patch it in `lark/__init__.py`.

lark/tree.pyi
KmolYuan: Since superclass `object` already have it, unless it's a abstract method, I will remove it.

lark/exceptions.pyi
erezsh: My point is that if a value will be displayed, it has to be the correct one, and change if the code changes. Otherwise, it's better to not have a value.

KmolYuan: I think the default value is just to remind the user that this is an optional parameter of default behavior.
Regardless the implementation, they should keep refer to docstring or documentation if they are not truly understand the function.

erezsh: A quick search revealed the correct way to write it:

    for arguments with a default, use ... instead of the actual default;

https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#conventions

lark-stubs/exceptions.pyi
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+from typing import Dict, Iterable, Callable
+from .tree import Tree
+
+
+class LarkError(Exception):
```
erezsh: So these empty exceptions can be removed now, right?

lark-stubs/lexer.pyi
```diff
@@ -0,0 +1,202 @@
...
+    mres: _MRes
+
+    def __init__(self, mres: _MRes):
+        ...
+
+    def __call__(self, t: Token) -> Token:
+        ...
+
+
+class CallChain:
```
erezsh: Why did you add all these internal classes? I don't think 99% of the users will ever even see them..

KmolYuan: Oh, That's the note when I'm tried to deriving member types in some methods, same as `_Lex`.
I can remove them.

lark-stubs/exceptions.pyi
KmolYuan: The public exception should be listed to avoid import problem in MyPy, because `pyi` will replace `py` module when doing type check.

erezsh: Why would there be an import problem though?

lark-stubs/lexer.pyi
erezsh: The thing is that some of these classes might change in the future, because they aren't part of the API.

How does it help to have these? (For example with _Lex?)

lark-stubs/exceptions.pyi
KmolYuan: For users checking their code. MyPy will check the annotated dependencies either.

lark-stubs/lexer.pyi
KmolYuan: I left it behind when assuming the types of some members, because they are generated from some non-public methods.

KmolYuan: Like `CallChain` and `UnlessCallback` can be simplified as `Callable` but I don't known at first time.

lark-stubs/exceptions.pyi
KmolYuan: Would it be better to identify the public name if you add a `__all__` list? ([PEP8](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names))

erezsh: If it's necessary and will allow to remove these re-definitions, that I'll consider it.

Is the an import problem currently? I don't understand why..

KmolYuan: Do you mean that empty inheritance like "LarkError" should be removed? If it is a public name, it can't, and its existence is necessary. If not, other subclasses can inherit from `Exception` directly.

erezsh: My question is why do exceptions need to be included at all? There is no additional information presented in them, and they are already accessible by importing exceptions.py?

KmolYuan: No, pyi will replace py files if exist.

<https://mypy.readthedocs.io/en/stable/stubs.html#creating-a-stub>

> If a directory contains both a .py and a .pyi file for the same module, the .pyi file takes precedence.

erezsh:
Thanks!

PR #533, opened 2020-03-12
Fix lexer_callback arg in stub for Lark constructor

jnwatson:
Mypy was giving me a "error: Missing named argument "lexer_callbacks" for "Lark"" because of the missing default None in the last parameter of the Lark constructor.


erezsh:
Thanks!

Issue #541, opened 2020-03-26
Flag to make all terminals case insensitive


rdeoliveira:
**Feature request**
As a developer using lark-parser
I would like to set a global flag (when instantiating the parser?) to treat all terminals as case insensitive
So that I don't have to add `i` to all my terminals, if case doesn't matter for any of them

MegaIng:
I am currently working my way through the source code to add this feature, but I have a few question regarding the design of the source code. @erezsh How do I best ask these? This is in regards to a few places where I am not sure why it was designed the way it is. (Example: Why don't all Lexers directly take the LexerConf object and unpack it themselves?).

erezsh:
@MegaIng You can ask me on gitter or through mail.

Regarding LexerConf, I really don't remember. Maybe it's possible to just pass it instead of unpacking it, it doesn't seem like a significant difference.

MegaIng:
@rdeoliveira Does the PR fulfill your wishes? If yes, you can close this issue.

rdeoliveira:
I suppose so. I suppose the new version is not yet available via pip install?

Best,
Rodrigo

MegaIng:
No, sadly only @erezsh can create a new release. you can manually download and `setup.py install` the newest 'development' version in this repo.

rdeoliveira:
Not to worry. I'll wait for until this can be pip installed. Will be easier given our CI/CD pipeline.

I can close this now. Thanks @MegaIng for the quick reaction!

erezsh:
@rdeoliveira Luckily, I can create a new release. `0.8.3` is now out with this feature.

rdeoliveira:
@erezsh This is awesome! Thanks.

PR #543, opened 2020-03-26
Added global_flags option to Lark (+ test)

MegaIng:
This is an attempt at a simple solution for #541. It adds a `global_flags` keyword argument to the `Lark` class that is distributed to all `re.compile` calls, essentially applying these flags globally. You have to import the `re` modules flags (e.g. re.I), strings wont work.

I am not sure how to add this to the docs. I think I have never looked into them.


erezsh:
That looks alright. But please use a different name. `global_re_flags`, or `g_regex_flags`.

erezsh:
Thanks!

PR #544, opened 2020-03-28
Template rules

MegaIng:
A basic implementations of Template Rules and test case (all pass). A few TODOs are still left:

- More better error messages
- potentially change value -> atom in arguments
- potentially add TERMINAL templates

### DIFF
--- a/examples/lark.lark
+++ b/examples/lark.lark
@@ -4,8 +4,11 @@ _item: rule
      | token
      | statement
 
-rule: RULE priority? ":" expansions _NL
-token: TOKEN priority? ":" expansions _NL
+rule: RULE rule_params priority? ":" expansions _NL
+token: TOKEN token_params priority? ":" expansions _NL
+
+rule_params: ["{" RULE ("," RULE)* "}"]
+token_params: ["{" TOKEN ("," TOKEN)* "}"]
 
 priority: "." NUMBER
 
@@ -27,9 +30,12 @@ name_list: "(" name ("," name)* ")"
 
 ?atom: "(" expansions ")"
      | "[" expansions "]" -> maybe
-     | STRING ".." STRING -> literal_range
-     | name
-     | (REGEXP | STRING) -> literal
+     | value
+
+?value: STRING ".." STRING -> literal_range
+      | name
+      | (REGEXP | STRING) -> literal
+      | name "{" value ("," value)* "}" -> template_usage
 
 name: RULE
     | TOKEN
--- a/examples/template_lark.lark
+++ b/examples/template_lark.lark
@@ -1,0 +1,56 @@
+start: (_item | _NL)*
+
+_item: rule
+     | token
+     | statement
+
+_rule_or_token: RULE
+              | TOKEN
+rule: RULE rule_params priority? ":" expansions{_rule_or_token} _NL
+token: TOKEN priority? ":" expansions{TOKEN} _NL
+
+rule_params: ["{" RULE ("," RULE)* "}"]
+
+priority: "." NUMBER
+
+statement: "%ignore" expansions{TOKEN} _NL         -> ignore
+         | "%import" import_path{_rule_or_token} ["->" _rule_or_token] _NL   -> import
+         | "%import" import_path{_rule_or_token} name_list{_rule_or_token} _NL     -> multi_import
+         | "%declare" TOKEN+                        -> declare
+
+!import_path{name}: "."? name ("." name)*
+name_list{name}: "(" name ("," name)* ")"
+
+?expansions{name}: alias{name} (_VBAR alias{name})*
+
+?alias{name}: expansion{name} ["->" RULE]
+
+?expansion{name}: expr{name}*
+
+?expr{name}: atom{name} [OP | "~" NUMBER [".." NUMBER]]
+
+?atom{name}: "(" expansions{name} ")"
+           | "[" expansions{name} "]" -> maybe
+           | value{name}
+
+?value{name}: STRING ".." STRING -> literal_range
+            | name
+            | (REGEXP | STRING) -> literal
+            | name "{" value{name} ("," value{name})* "}" -> template_usage
+
+_VBAR: _NL? "|"
+OP: /[+*]|[?](?![a-z])/
+RULE: /!?[_?]?[a-z][_a-z0-9]*/
+TOKEN: /_?[A-Z][_A-Z0-9]*/
+STRING: _STRING "i"?
+REGEXP: /\/(?!\/)(\\\/|\\\\|[^\/\n])*?\/[imslux]*/
+_NL: /(\r?\n)+\s*/
+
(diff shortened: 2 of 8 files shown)


MegaIng:
Btw. why are you not running a check that the tests are passing successfully?

erezsh:
I tried playing around with it. Can you explain why I'm getting this exception?

```python
            g = _Lark(r"""
                       start: "a" maybe{"b"} "b"
                       maybe{x}: x?
                       %ignore " "
                       """)
            g.parse('ab')

lark.exceptions.UnexpectedToken: Unexpected token Token($END, '') at line 1, column 2.
Expected one of:
        * B
```

Also, you should update `examples/lark.lark`

erezsh:
I had another thought while reviewing your code.

Right now, you're treating rules and templates as two separate and distinct entities. But isn't a rule just a template with 0 params?

If we treat them as the same thing, it opens up some options that seem to make sense. 

For example, that means templates, just like rules, can accept `_` and `?` to change their "tree signature", where the name of the tree node is the name of the rule/template without the params.

But if templates are always inline, it means you'll often need "helper rules" to create a nice structure.

It also lets us share a lot of the code for both them, and the grammar too.

Of course, we can still have nicer error messages, calling them template (by checking if `len(params)>0`). But even then, it feels like a solid win.

What do you think?

MegaIng:
I also had that train of thought. I came to a conclusion that this would not work. But I can't remember why. (Btw. all-tough I didn't test it, templates should be able to take `?`, `!` and priorities. My reasoning for always inlining them is to avoid auto generated names in the resulting tree. But we could just give all rules based on the same template the same alias)

erezsh:
Well, see if you can remember :)

I'm fairly sure that it would. And as you said, we can use aliases. In a way, that's how rules work already. If there is no explicit alias, it just uses the rule name as the alias.

MegaIng:
Ok, I will just try to refactor everything and see if I hit a roadblock. (and also make the alias-thing)

MegaIng:
Ok, unification is now successfully done. I also think I have the dealing with alias correctly, but currently no tests for that exist.

MegaIng:
> ```
>            g = _Lark(r"""
>                       start: "a" maybe{"b"} "b"
>                       maybe{x}: x?
>                       %ignore " "
>                       """)
>            g.parse('ab')
>
>lark.exceptions.UnexpectedToken: Unexpected token Token($END, '') at line 1, column 2.
>Expected one of:
>        * B
>```

I don't no why this error happens, you explain it to me:
```python
>>> g = lark.Lark(r"""
...            start: "a" _maybe "b"
...            _maybe: "b"?
...            %ignore " "
...            """, parser='lalr')
... 
>>> g.parse('ab')
lark.exceptions.UnexpectedToken: Unexpected token Token($END, '') at line 1, column 2.
Expected one of: 
	* B
```

(Its a bug in the lalr parser. I have no idea about the inner workings.)

MegaIng:
Ok, if you take a look at `examples/template_lark.lark`, you can see an interseting usecase of templates.

erezsh:
Everything looks much nicer now imho. I'll look at it soon.

> Btw. why are you not running a check that the tests are passing successfully?

It used to run on travis. I think that adding mypy check through "github actions" made it stop running the `.travis.yml`.

> Its a bug in the lalr parser

Ah, my bad. It's not a bug though, just a shift-reduce collision.

erezsh:
> Btw. why are you not running a check that the tests are passing successfully?

Fixed it now in master.

MegaIng:
I found a problem that I am not sure how to work around: If the ?question mark modifier is used, it conflicts with the usage of alias that I am generating. This code:

```
start: expr{b}
b: "B"
?expr{t}: "A" t+
```
get turned into:

```
start: expr_b
b: "B"
?expr_b: "A" b+ -> expr
```
(not quite. The name of the generated rule is different)

if we now parse `AB`, we don't get the expected `Tree('start', [Tree('b', [])])` (as one would think from the original expression), but `Tree('start', [Tree('expr', [Tree('b', [])])])`, since we have an alias. to fix this we have to remove the alias:


```
start: expr_b
b: "B"
?expr_b: "A" b+
```
would correctly return `Tree('start', [Tree('b', [])])`, but would incorretly return ``Tree('start', [Tree('expr_b', [Tree('b', []), Tree('b', [])])])`` for `ABB`.

To you have a solution on how to fix this?

erezsh:
The line that makes this happen is in `parse_tree_builder.py`
```python
                (expand_single_child and not rule.alias) and ExpandSingleChild,
```

It works under the logic that an alias means that the `?` flag should be ignored. But that logic is for the user-interface, and doesn't *have* to apply. It's not part of the core logic.

So I think the way around it is to maintain the difference between a user alias and an automated alias create by templates.

Maybe in `RuleOptions.auto_alias`?

MegaIng:
Ok, this helped to fix it and actually improve the system: `RuleOptions` now has on attribute more `template_source` which saves the name of the template from which the rule was created. This is used to create the alias when no other names were found.

erezsh:
Nice, that turned out reasonably elegant.

Are you thinking of adding terminal templates as well? (I see you prepared the grammar for it). It might be pretty easy, just some mix of `ApplyTemplates` and `PrepareAnonTerminals`.

MegaIng:
Yes, I am working on it.

erezsh:
Don't know if it's an edge-case, but how about passing templates as arguments?

```python
l = Lark('''start: a{b}
            a{t}: t{"a"}
            b{x}: x''')
```

It would be pretty cool if implemented.

If we decide to not support it for now, we should at least have a better error in place.

Right now it's

    Template 't' used but not defined (in rule a)

But really `a{b}` should have raised the error, saying you can't pass templates.

MegaIng:
My brain can not reason about what tree this grammar should produce. It shouldn't be hard to implement (In fact, I think it is just a to strict check before any actual replacing gets done).

I think the correct tree is this: `Tree('start', [Tree('a', [Tree('b',[])])])`, isn't it?

erezsh:
Yes, that's how I read it too.

MegaIng:
Ok, here a version where meta templates should work. (And I am still working on TokenTemplates)

erezsh:
Very nice! I'm certainly glad we took the "rule is template{}" route.

erezsh:
I played around with it , and noticed that one of the biggest obstacles for my refactoring so far, is that the rules are only inlined at the tree level (`_rule`), and not the rule level, and that creates more reduce collisions (in LALR ofc). So I think adding this feature of "inlined at compile" could be useful in general, but really help the use of templates.

I was thinking about the `__rule` syntax for it. Not very imaginative, but recognizable from python, and it's better than another strange symbol like `@` or `&`.

MegaIng:
I think this is a good idea and interesting feature, but might be hard to explain to the end users. But we don't have to explain it. (But it should be mentioned in the docs, and should kept in mind to solve questions)

PR #553, opened 2020-04-03
Added hook-lark.py and the required lines in setup.py

MegaIng:
Note: tests could still be added, and someone else should check if this correct.

### DIFF
--- a/lark/__pyinstaller/hook-lark.py
+++ b/lark/__pyinstaller/hook-lark.py
@@ -1,0 +1,14 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2017-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


lark/__pyinstaller/hook-lark.py
```diff
@@ -0,0 +1,14 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2017-2020, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
```
Legorooj, line 10: Can you change the copyright header? We only put these on hooks in our repo to stop someone suing us. I'd change to the standard lark header.

erezsh: We can probably just remove it. Maybe I'm naive, but I'm not worried about being sued, and Lark's MIT license already covers this file.

Legorooj: @erezsh your repo, your choice. You're pretty safe - we don't actually need to have the copyright header, but it's that little bit more security in a hypothetical case.

MegaIng: I copied the file over from the other PR. I didn't notice the header was still in there. I would say remove is the best option.

PR #572, opened 2020-05-03
Fixed links, link to stand-alone tutorial, renamed index.md

phorward:
Renaming the README.md is only a proposal because it directly renders on GitHub and you don't have to choose index.md first to get an overview. The links in how_to_use.md pointed to 404s, I wasn't sure to which direction they pointed initially.


erezsh:
Thanks for the PR.

Won't removing docs/index.md interfere with readthedocs ?

erezsh:
Also, it's referred to by `mkdocs.yml`

phorward:
@erezsh  you're right, I moved it back and added a dummy README.md with a link!

erezsh:
Per my previous comment, the link to root `/` works well in readthedocs, which your changes will break. It also works in `mkdocs --serve`, so only github doesn't like it (for obvious reasons).

I'll only merge a solution that supports readthedocs, since it's the main view for reading the documentation.

erezsh:
To be a bit more helpful: You should try replacing them with relative links to index (`../index.md`)

phorward:
Ok then its better to let this unchanged.

PR #574, opened 2020-05-03
Rewrite iter_subtrees to more efficient version

BlankSpruce:
Using OrderedDict provides these properties:
- given subtree is yielded only once since it's stored in OrderedDict
only once (even though it may be put there multiple times) so no need
to double check if subtree was already seen
- order of iteration is preserved as it was previously when subtrees
to iterate over were stored in a list

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -4,6 +4,7 @@ except ImportError:
     pass
 
 from copy import deepcopy
+from collections import OrderedDict
 
 
 ###{standalone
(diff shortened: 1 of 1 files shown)


BlankSpruce:
I've amended commit to avoid the issue with `OrderedDict.values()` being not usable with `reversed` in Python 3.4.

MegaIng:
I would suggest adding a `del queue` statement before the second for-loop to reduce the memory footprint.

Also, this algorithm is less efficient since if trees appear more than ones, both copies are added to the queue, despite all the children already being in the queue. The first repetition check is still needed.

BlankSpruce:
Done.

erezsh:
Cool, thanks!

PR #576, opened 2020-05-04
Remove meta property in favor of member access

BlankSpruce:
It seems that check for presence of `self._meta` doesn't have to be
performed every time `self._meta` is about to be fetched if it can be
just default constructed on `Tree` initialization. Avoiding this check
removes unnecessary performance penalty for using this member

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -27,33 +27,39 @@ class PropagatePositions:
     def __call__(self, children):
         res = self.node_builder(children)
 
+        # local reference to Tree.meta reduces number of presence checks
         if isinstance(res, Tree):
+            res_meta = res.meta
             for c in children:
-                if isinstance(c, Tree) and not c.meta.empty:
-                    res.meta.line = c.meta.line
-                    res.meta.column = c.meta.column
-                    res.meta.start_pos = c.meta.start_pos
-                    res.meta.empty = False
-                    break
+                if isinstance(c, Tree):
+                    child_meta = c.meta
+                    if not child_meta.empty:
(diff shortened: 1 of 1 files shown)


MegaIng:
It increases the memory footprint, and the performance gain should be minimal.

Did you measure it, or is this a blind optimization?

BlankSpruce:
Not a blind optimization but rather low hanging fruit. I've been recently profiling my project to see what I can do better but couldn't stop my curiosity about internals of lark itself. No doubt, it's quite targetted optimization when `propagate_positions` is used but nevertheless in my case most of the gains aren't even observed in my code but inside `PropagatePositions.__call__`. Proposed change reduces time spent there by around a third which leads to around 12-15% less time spent inside `lalr_parser:45:parse` which I'd say is worth considering.

MegaIng:
I would suggest modifying `PropagatePositions.__call__` instead. It will have almost the same gains, but without the memory footprint increase when `propagate_positions=False`. (this can be a lot when the trees are huge)

BlankSpruce:
Thanks for advice. I've given a try in limiting calls to `Tree.meta` inside `PropagatePositions.__call__` instead as suggested and the gains aren't as close but not that far either. In the same profiling scenario gain is around 11-12% which sounds good enough. It seems like a fair compromise since worrying about memory footprint is certainly justified. I've amended the commit and commit message to reflect what's actually reworked.

I'm aware that `propagate_positions=True` might not be for the best in the first place but that doesn't mean such use-case can't be improved.

erezsh:
Looks good, thanks!

Issue #590, opened 2020-05-14
Use regex in addition to re?


julienmalard:
**Suggestion**
While using `Lark` I came across an already-documented bug in the builtin `re` module, whereby abugidas with vowel marks (that is, most of South Asia and Oceania) fail to match the `\w` directive (e.g., as in `re.match("^[\w\s][\w\s]*", "किशोरी", re.UNICODE)`, as taken from this question [here](https://stackoverflow.com/questions/12746458/python-unicode-regular-expression-matching-failing-with-some-unicode-characters)).

It seems that the recommended "solution" is to simply used the `regex` module instead of `re`. However, this would mean adding a dependency to `Lark`, so I thought I would ask whether you would look favourably upon such a suggestion before submitting a pull request. I was thinking of proposing something along the lines of:
```
try:
    import regex as re
except ImportError:
    import re
```
so that `regex` is not required and nothing breaks if the dependency is not installed.

**Describe alternatives you've considered**

An alternative, which I am currently using as a workaround, is to generate the toxens manually. For instance, for a Python parser, I can use:
```
ID_START = ''.join([
    c for c in map(chr, range(sys.maxunicode))
    if category(c) in ['Lu', 'Ll', 'Lt', 'Lm', 'Lo', 'Nl']]) + '_'

ID_CONTINUE = ID_START + ''.join([
    c for c in map(chr, range(sys.maxunicode))
    if category(c) in ['Mn', 'Mc', 'Nd']]) + '·'
```
and then define my `NAME` token as `ID_START ID_CONTINUE*`. However, I think that it would be much nicer to be able to use `\w` according to Unicode standard. A side benefit of using `regex` is that projects using Lark and specifying `regex` as a dependency could also then explicitly use Unicode categories in their regexes as well.

Please do let me know!

erezsh:
Hi,

I'm certainly open to supporting the `regex` library, if it's useful.

Is it fully-compatible with `re`, or are there differences beyond bugfixes?

How does it compare in terms of performance?

I think it might be nice to support something like

    pip install lark-parser[regex]

Not sure about making it the default though.

julienmalard:
Hello,
Thank you! I am not sure about performance; if you have a few recommended benchmarks for common regexes in grammars I could check quite quickly though.
I am quite sure that it is fully compatible. Apart from bugfixes, it also allows for unicode categories (see comments [here](https://stackoverflow.com/questions/1832893/python-regex-matching-unicode-properties)).

[...]

PR #593, opened 2020-05-14
Added regex module option.

julienmalard:
Added regex module option (optional alternative to re module).

I believe we should also add a test configuration to make sure all tests pass with `pip install lark-parser[regex]`, but was unsure of where to add it in the `tox.ini` file; please do let me know if you have any pointers. I could then also add a conditional test with a grammar containing a `regex` module-specific terminal definition that only activates when the [regex] optional dependency is enabled.

Please let me know what you think!


julienmalard:
Will close #590.

erezsh:
Well, it's good that the tests pass.

If you want to add regex-specific tests, the place to do it is in the `tests` folder. Just don't run the tests if the import fails (look at the nearley test).

MegaIng:
@erezsh I think what @julienmalard means that we add an additional pull-request check that also installs the `regex` modules.

julienmalard:
@MegaIng Yes, I think that that is what I meant. I was referring to adding a test matrix option that runs all tests with regex installed, so that we can check that everything works when regex is used instead of re (the tests run now run without regex installed, so, effectively, with no changes as compared to the master branch).

erezsh:
If regex already have tests that ensure that it's fully backwards-compatible with re, I'm not sure there's a point of also doing those tests on our end. At least not more than once in a long while.

julienmalard:
All right, sure then! Should I include `regex` module-specific tests, or is all fine?

erezsh:
Yes, you should include module-specific tests.

julienmalard:
I have tried adding tests as `test_regex.py`. I do think that I need help for a few points:
1. I am not too sure how to tell tox to install `regex`. I added a `regex-requirements.txt` file and also
2. There is a problematic line in utils.py: `sre_parse.parse(regexp).getwidth()`. The problem is that `sre_parse` will not accept character classes from `regex` (the `\w` bugfix works a charm, though!). The effect of that, for the moment, is that any grammars including character class regexes will fail to compile (which at least is not a regression from current behaviour).
3. I wasn't sure how to make sure the `test_regex.py` tests are run. Is including the file in the tests directory enough, or do I need to specify it somewhere?

Many thanks!

erezsh:
I'm worried about adding this as the default, as users might experience performance degradation. It would be much better, imho, if this was opt-in rather than "always on"

julienmalard:
Hello,
Sure, I think I could change that quickly enough. Do you think it would be best to have it as a global option, e.g., 

```
import lark
from lark import Lark

lark.use_regex = True
parser = Lark('my grammar.lark', 'lalr')
```

Or as an option specific to each parser:
```
from lark import Lark
parser = Lark('my grammar.lark', 'lalr', regex=True)
```

erezsh:
Option #2 sounds good.

erezsh:
That looks good. Just make sure that the tests pass.

julienmalard:
Thanks! It was a missing tox dependency for `regex`. Testing it again now.

julienmalard:
Looks like this will take a bit of trial and error! Do you know how I can tell tox to install `regex` prior to running tests? All works well locally, but not on GitHub builds. Thanks!

erezsh:
See `.github/workflows/tests.yml`

Basically copy whatever we did for nearley-requirements. Maybe we can even merge them (for the tests)

julienmalard:
Thanks! I updated stubs and the README to match changes. Please let me know if all was done well!

erezsh:
There is a reason that all the parser tests sit in `test_parser.py`. There is an automated process there that runs each test for every parser and lexer. For example, your tests fail when `parser="lalr"`.

Put your tests there and make sure that they pass.

Also, you forgot to update `tox.ini`

And lexer.py still imports `regex` for some reason, although it shouldn't (and probably shouldn't import `re` either)

julienmalard:
Thanks! Hope this works. I changed the tests, fixed the problems with `lalr`, and updated `tox.ini`. I also removed the `regex` import from `lexer.py`, but line 48 still requires `re`, so I left that one in.
Please let me know how it looks now!

erezsh:
Notice that pypy3 fails for some reason. Any chance you can figure out why?

erezsh:
Hmm looks like most of the tests failed..

julienmalard:
That is odd! It seems that pypy3 failed and that the rest aborted for some reason (without printing out error messages). Is it possible that the pypy3 failure caused the rest to stop running?
I am not sure what problems is occuring with pypy3 (all runs well locally with regular python), so I'll unfortunately probably have to debug remotely here with lots of commits and print statements...I hope you wouldn't mind?

erezsh:
It shouldn't be that difficult to install - https://www.pypy.org/download.html#python-3-6-compatible-pypy3-6-v7-3-1

julienmalard:
I think it worked!

erezsh:
Congrats on a great PR 👍

julienmalard:
Many thanks for your help and guidance throughout the process!

PR #597, opened 2020-05-30
Use token type equality as a fallback when matching error examples

aleh-arol:
Enhance UnexpectedInput example matching by using token type equality as a fallback from full token equality (instead of just parser state equality). So with this change best match selection is 3 stage process:
* If parser state match and full token match - select candidate
* If parser state match and token types match - favour this candidate over others matching just parser state
* If parser state match only - select candidate as so far best option
Example of where it is useful is when tokens to match are not fixed strings.

### DIFF
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -32,7 +32,7 @@ class UnexpectedInput(LarkError):
         after = text[pos:end].split('\n', 1)[0]
         return before + after + '\n' + ' ' * len(before) + '^\n'
 
-    def match_examples(self, parse_fn, examples):
+    def match_examples(self, parse_fn, examples, token_type_match_fallback=False):
         """ Given a parser instance and a dictionary mapping some label with
             some malformed syntax examples, it'll return the label for the
             example that bests matches the current error.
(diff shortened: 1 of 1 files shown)


erezsh:
This is a potentially backwards-incompatible change, so I have to be careful about merging it in.

Can you include a concrete example where this would be useful?

Also, how would it affect this? https://github.com/lark-parser/lark/blob/master/examples/error_reporting_lalr.py

aleh-arol:
I will try:
overall idea (which is obvious from the commit) is to favour unexpected token type instead of "selecting the first provided error example matching parser state as a best candidate for error". So when you have several potentially similar error types (same context) but which differ with the unexpected token type only this change helps to select the one that matches the problem better.

e.g. the problem is you got unexpected token because you're expecting "A" or "B" followed by "C", then depending on the data (actually the order of the keys in the dict) of error examples you provide to match_examples the one matching parser context which will be the first found will be the candidate. 

So you have 2 error examples in your examples dict 1) "Unexpected A" 2) "C without B" (latter error is more specific case of another in terms of parser context, and you want to provide a specific one to users - e.g. context based error) - in case unexpected token was "C" you want error 2 otherwise error 1. The only difference between error examples is unexpected token type (becuase parser context is the same).

aleh-arol:
As for backwards compatibility this can be an option of the match_examples which is off by default.

erezsh:
Okay. If you change it to be off by default, I'll accept it.

aleh-arol:
Done

PR #606, opened 2020-06-30
Reconstruction keep terminals

julienmalard:
This pull request ports over some commits made on the `recons_work2` branch some time ago, which has since fallen behind `master`. This part fixes problems with tokens that should be kept in the parse tree for reconstruction.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -307,6 +307,7 @@ class PrepareAnonTerminals(Transformer_InPlace):
         self.term_set = {td.name for td in self.terminals}
         self.term_reverse = {td.pattern: td for td in terminals}
         self.i = 0
+        self.rule_options = None
 
 
     @inline_args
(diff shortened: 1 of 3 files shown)


erezsh:
Hi, sorry it took me a while to get to this PR!

Can you explain what `rules_for_root` are for? And why you're creating a new reconstruction parser for each rule?

And why you had to change how `keep_all_tokens` was handled?

(Sorry if I already asked these before) 

Also, you import `unicodedata` but don't do anything with it.

julienmalard:
Hello Erez,

No problem! 

Those are difficult questions, since those are mostly ports from the `recons_work2` branch and I have forgotten most of the reasons why since. If I remember correctly, I think you helped me out quite a bit and did some of those changes yourself. 
I think that the majority of the conversation was at https://github.com/lark-parser/lark/pull/199, if that helps any. Please let me know if you'd like me to look more deeply into this again.

Regarding `unicodedata`, I had made a small change to fix an issue with `isalnum` in the reconstructor but had saved that commit for a future pull request (and apparently forgot to remove the import before committing).
-Julien

erezsh:
Well, looks like I was fine with it before, and everything works, so I'll accept it on faith. 

Thanks for the PR! 👍

julienmalard:
Thank you!

Issue #604, opened 2020-06-26
Avoid using root logger for debug


pwwang:
Using the root logger will interrupt other loggers.

Instead of using:
```python
import logging
logging.basicConfig(...)
```

We probably should do:
```python
import logging
logger = logging.getLogger('lark') # or other fancy name
# configure logger

# and use the logger instead of root logging to keep it independent
```

MegaIng:
Could you make a pull request?

PR #607, opened 2020-06-30
Avoid using root logger

pwwang:
#604

### DIFF
--- a/docs/how_to_use.md
+++ b/docs/how_to_use.md
@@ -30,13 +30,13 @@ Use the reference pages for more in-depth explanations. (links in the [main page
 
 ## LALR usage
 
-By default Lark silently resolves Shift/Reduce conflicts as Shift. To enable warnings pass `debug=True`. To get the messages printed you have to configure the `LOGGER` beforehand. For example:
+By default Lark silently resolves Shift/Reduce conflicts as Shift. To enable warnings pass `debug=True`. To get the messages printed you have to configure the `logger` beforehand. For example:
 
 ```python
 import logging
-from lark import Lark, LOGGER
+from lark import Lark, logger
 
-LOGGER.setLevel(logging.DEBUG)
+logger.setLevel(logging.DEBUG)
 
 collision_grammar = '''
 start: as as
--- a/docs/how_to_use.md
+++ b/docs/how_to_use.md
@@ -30,12 +30,13 @@ Use the reference pages for more in-depth explanations. (links in the [main page
 
 ## LALR usage
 
-By default Lark silently resolves Shift/Reduce conflicts as Shift. To enable warnings pass `debug=True`. To get the messages printed you have to configure `logging` framework beforehand. For example:
+By default Lark silently resolves Shift/Reduce conflicts as Shift. To enable warnings pass `debug=True`. To get the messages printed you have to configure the `LOGGER` beforehand. For example:
 
 ```python
-from lark import Lark
 import logging
-logging.basicConfig(level=logging.DEBUG)
+from lark import Lark, LOGGER
+
+LOGGER.setLevel(logging.DEBUG)
 
 collision_grammar = '''
 start: as as
--- a/lark/__init__.py
+++ b/lark/__init__.py
@@ -1,3 +1,4 @@
+from .common import LOGGER
 from .tree import Tree
 from .visitors import Transformer, Visitor, v_args, Discard
 from .visitors import InlineTransformer, inline_args   # XXX Deprecated
--- a/lark/__init__.py
+++ b/lark/__init__.py
@@ -1,4 +1,4 @@
-from .common import LOGGER
+from .common import logger
 from .tree import Tree
 from .visitors import Transformer, Visitor, v_args, Discard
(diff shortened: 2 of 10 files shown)


MegaIng:
And one more general comment: Might be overkill, but the official logging documentation suggest using the hierarchical structure of the module also in the loggers, e.g. using `getLogger(__name__)` for each individual module. Since they are all children of `lark`, the global config would be applied to all children. (I think?)

But really nice PR. I don't think I would have even though of testing this xD.

pwwang:
Have thought about it actually. Using `__name__` to manage the hierarchical structure of loggers is great, especially in terms of giving the flexibility to configure different sub-loggers. However, (sub-)loggers in lark are mainly used in different parsers (early and lalr, as far as I can see from the source code); and they are mutually exclusive, which is controlled by the `Lark` constructor options, meaning there is only one sub-logger working at the same time in such case. 

Using a unique name isn't that bad for now since we don't have complex logger hierarchy at the moment. And this is easy to be extended to a hierarchical structure in the future if we are planing for heavy loggings.

But feel free to ask me to switch that hard-coded logger name (`LARK`) to something like (`__name__`). It doesn't matter for now, because only the logger itself (`LOGGER`) is exposed to the users.

MegaIng:
I would suggest at least changing the hard coded name to `lark`, all lowercase. That is what others are expecting, and we don't really have a reason here to go against the norm. And my online primary concern with sub-loggers was for future work, but I do not really have a idea why would really need it. So we don't need it, I agree.

pwwang:
The other idea is that we could actually collapse the 2 settings into 1 for debug logging. By 2 settings I mean this:
```python
import logging
from lark import Lark, LOGGER

LOGGER.setLevel(logging.DEBUG) #1

Lark(grammar, debug=True)      #2
```

Instead, we could set up the logger level inside `Lark` constructor according to the debug option value, leaving the `LOGGER` untouched, unless specific configurations needed (for the `LOGGER`).

What we need to do to achieve this, I think, is to select a good logging level being set for the `LOGGER` inside `Lark` constructor. Accordingly, we should use higher (or equal) levels for the messages we want to show when `debug` is on, and lower ones when it's off.

For example, in `Lark` constructor:
```python
import logging
from .common import LOGGER

class Lark ...
  def __init__(...):
    # ...
    if debug:
      LOGGER.setLevel(logging.DEBUG)
    else:
      # when debug is off, only show messages at/above ERROR level
      LOGGER.setLevel(logging.ERROR)

# Then this LOGGER can be used anywhere else after a `Lark` object is created.
# If debug is on, messages with DEBUG and above will be showing,
# otherwise, only those with ERROR and above are showing.
```

Then one should be able to do this without touch the LOGGER and logging stuff, unless he wants to customize it for special use:
```python
Lark(grammar, debug=True) # debug messages showing
```

BTW, hierarchical sub-loggers are super useful for big projects where people want to control the behavior of loggers from submodules (ie. different logging levels, handlers and filters perhaps).

pwwang:
For the above idea, however, I don't want to change your code until we make some agreements on the loglevel that should be set when debug is on or off. Because we might also need to change the levels of messages that you used originally, of which I have no idea about the importance.

MegaIng:
One problem I could imagine with the debug=True also setting the Logger level is that you can not activate you debug without also getting the messages, potentially spamming chat. The debug option also generates a graphical overview if I remember correctly, so there is a reason why you might not want the see the debug messages. (And all of this happens before the constructor returns, so you can't manually set it back.

pwwang:
> One problem I could imagine with the debug=True also setting the Logger level is that you can not activate you debug without also getting the messages, potentially spamming chat. The debug option also generates a graphical overview if I remember correctly, so there is a reason why you might not want the see the debug messages. (And all of this happens before the constructor returns, so you can't manually set it back.

Oh, never mind. I haven't realized that there are things to be done other than logging when debug is on.

MegaIng:
One thing I just realized: This is breaking backwards compatibly. Although probably not relevant, `logging.basicConfig()` will no longer work correctly.

erezsh:
Please don't test for specific log messages. The wording is likely to change.

Also, I don't like the `LOGGER` convention. It's better to just call it `logger`.

pwwang:
@erezsh No problem with that being changed. For the test, what about just testing there should be something outputting instead of specific messages? If should work if those messages will not be eliminated in the future in that test case.
 
@MegaIng This is from the python doc:
```
logging.basicConfig(**kwargs)
Does basic configuration for the logging system by creating a StreamHandler with a default Formatter and adding it to the 
root logger. The functions debug(), info(), warning(), error() and critical() will call basicConfig() automatically 
if no handlers are defined for the root logger.
```
Since we are not using root logger, it is supposed to not work anymore. However, I am not sure how backward compatibility should be handled in such case.

erezsh:
You can just test if errors were made or not.

Or, maybe it's possible to add some code / id to the error messages, so that you can look for that id in the messages. But it has to be done in way that won't confuse the users or obfuscate the message too much.

erezsh:
Thanks for the PR!

Issue #626, opened 2020-07-19
Support 'bytes' for input and Tokens, in addition to 'str'


ctrlcctrlv:
**Describe the bug**

https://github.com/lark-parser/lark/blob/b5abf2d7afcd70af2be6c57aa454e532c8a99a0b/lark/lexer.py#L99

This merely masks what in 2020 is almost always a bug. If you must do something other than raise an error, decoding with `backslashreplace` is better. If this is a workaround because you're sometimes allowing Token to be a `bytes`...just allow its `value` to be a `bytes`...

erezsh:
Can you provide an example in which it poses an issue?

ctrlcctrlv:
Given how you responded to me in #624, I don't know if I should even bother posting this, but I already wrote it, so...if you don't want my help, just say so, and I'll go away. I gave you a simple example that causes a crash there. There are plenty of projects I can give my time to in a day...doesn't need to be yours... 😄 

Anyway...It is quite common, actually, to use LALR parsers to parse binary formats which have regular grammars and don't require more advanced state machines. If I'm not mistaken, BMP files (Windows bitmap) are simple enough to be parsed with LALR parser.

But I don't have time to implement one. So one other simple binary regular format is, actually, UTF8 itself! Here's a grammar I wrote:

**`utf8.lex`**
```lex
start: BOM? char*
BOM: "\xef\xbb\xbf"
char: CHAR1 | CHAR2 | CHAR3 | CHAR4
CONTINUATION_BYTE: "\x80" .. "\xbf"
CHAR1: "\x00" .. "\x7f"
CHAR2: "\xc0" .. "\xdf" CONTINUATION_BYTE
CHAR3: "\xe0" .. "\xef" CONTINUATION_BYTE CONTINUATION_BYTE
CHAR4: "\xf0" .. "\xf7" CONTINUATION_BYTE CONTINUATION_BYTE CONTINUATION_BYTE
```

This works fine, and other than the BOM, doesn't look half bad. Sadly, it is ugly to use, because Latin-1 seems to be the sanctioned Lark "way of doing binary", instead of Python's way, which is bytestrings:

**`utf8.py`**
```python3
# Example script for UTF8 grammar utf8.lark
# License: Apache 2.0
# Copyright 2020 Fredrick R Brennan

import lark
import regex
import sys
sys.stdout = open(sys.stdout.fileno(), mode="w", encoding="utf8", buffering=1)
import os

[...]

PR #630, opened 2020-07-23
Bytes support

MegaIng:
This is a start of implementing support for byte string as suggested in #626. This is still WIP.

My idea is that the grammar is still string, but you pass the `use_bytes=True` flag, make the patterns to be compiled with `bytes`. If you need to use match bytes that are not compatible with whatever encoding is used, you can just escaped them. They will be unescaped later.

TODO:

- [X] Add `use_bytes` to make regex compile as bytes
- [x] Add tests (essentially, everything needs to be tested again, but with bytes)
- [x] Find and check edge cases

@ctrlcctrlv, does this fix your use case?


MegaIng:
Btw. @erezsh The type checking test is configured wrong. It doesn't fail despite reporting errors. And how strongly are you against dropping support for python2.7? It would simplify this a lot.

ctrlcctrlv:
Hey @MegaIng. Very neat, nice progress! I am glad you were inspired by my idea. Yes, this was pretty much exactly what I had in mind. :smile::+1:

MegaIng:
@ctrlcctrlv I would like to use your code that you posted in the issue as test/examples. How would I best do that regarding the license? I have zero experience with that.

ctrlcctrlv:
@MegaIng That line came from another file, I didn't notice.

I edited it so it now reads

```
# Example script for UTF8 grammar utf8.lark
# By Fredrick R Brennan, 2020
# This file is released into the public domain. I dedicate the work to the public domain by waiving all of my rights to the work worldwide under copyright law, including all related and neighboring rights, to the extent allowed by law. You can copy, modify, distribute and perform the work, even for commercial purposes, all without asking permission. 
```

erezsh:
I'm very much against dropping 2.7. 

But this patch doesn't have to support it. i.e. `use_bytes` on python2 can just throw `NotImplementedError`

MegaIng:
Ok, I decided to skip the tests for python2.7. While the parse works in general, I would call it 'undefined behavior'. I will not be bother to fix this.

MegaIng:
@erezsh I think this PR is finished. Should I try to reduce the commit count?

erezsh:
Yeah, that would be nice!

Also, I think it's best to throw NotImplementedError for Python2 if the behavior is unpredictable. You can add a note saying it *mostly* works, but I wouldn't want a user to stumble into it by mistake, thinking it should work.

MegaIng:
@erezsh Like this?

erezsh:
The commit division seems okay (I didn't look too deeply)

But ideally the commit messages should describe what they contain.

erezsh:
Also maybe you could rename `source_grammar` into `grammar_source`? It would be more in line with Lark's naming convention.

MegaIng:
I could have sworn I already added better commit messages...

MegaIng:
Sitenote: I just used coverage.py to check how good the tests are. Total 84%, with a lot of small stuff in `__repr__` and similar. More interesting is that a lot of `early.py` is not covered. Only 64% is covered. I sadly have no idea how it works, so I can't say anything to it's importance.

erezsh:
Thanks @MegaIng  and @ctrlcctrlv for the pull request.

I made two changes you should be aware of:

* I removed the use of DualLark. This feature is too minor to justify doubling the test times. But I left the class there, so we can still use it whenever we make changes to the `use_bytes` feature.

* I changed the terminal encoding from utf8 to latin-1. It doesn't make sense to me to allow matching characters outside of `\x00-\xFF` for a bytes input, or to assume the encoding if always utf8. But maybe you can convince me otherwise.

MegaIng:
It doesn't matter, but it should be `ascii`. We are already making sure that there are no non-ASCII characters, to prevent exactly this problem.

MegaIng:
@erezsh At least the documentation has to be changed. latin1 will throw errors.

erezsh:
I thought you're already making sure there are no non-ascii characters?

MegaIng:
@erezsh Yes, but that is not reflected in the documentation.

erezsh:
I'm not sure I understand. Has the behavior changed? Or was it just not documented in the first place?

MegaIng:
You wrote this in the documentation:

> use_bytes - Accept and parse an input of type bytes instead of str. Grammar should still be specified as str, and terminal values are assumed to be latin-1.

This is wrong

erezsh:
Oh, hold on. Why not allow latin-1?

MegaIng:
To keep it simple and obvious. Don't make it encoding depended, latin1 doesn't support all possible byte values, so you still have to use `\x` anyway. Make it obvious how stuff get's loaded from files.

ctrlcctrlv:
>latin1 doesn't support all possible byte values

Really?

MegaIng:
Well, ok it does in python, but a lot are not mapped to usable characters. (Like it [should](https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Code_page_layout))

erezsh:
Let's take it from the start. There are two separate questions:

1) Do we want to support `\x00-\xFF`? I think the answer has to be yes, I don't see a reason to only support 7bit ascii, when we can support 8bits just as easily.

2) What's the best encoding to make it work? It seems like latin-1 does the trick, but if there's a better encoding that's fine.

Unless I'm missing something?

MegaIng:
Of course I want to support `\x00-\xFF`. But to reduce confusing, my suggestion is to do this explicitly with the escape sequence instead of with an encoding.

erezsh:
This PR is causing an error for Python 3.4: https://travis-ci.org/github/lark-parser/lark/jobs/716103753

That's something we'll have to fix before release.

Issue #629, opened 2020-07-23
Locally disable %ignore


Caagr98:
It is common that languages have some sections, such as strings or floats, which have completely different lexical syntax (w.r.t. whitespace and such) than everything else.

The current approach appears to be to define long unwieldy regexes and capturing the entire string as one token, as in (python3.lark)[https://github.com/lark-parser/lark/blob/master/examples/python3.lark#L173]. This is difficult to understand, and requires heavy postprocessing. It also only works if the grammar is regular, which makes string interpolation difficult to parse. (I think cPython works around this by parsing out {} sections in a second pass, which heavily restricts nesting.) One possible workaround would be using four different regexes (quote-to-quote, quote-to-splice, splice-to-splice, and splice-to-quote), but this is not a satisfactory solution.

It is of course possible to define a string as
```
string: "\""  strchar* "\""
strchar: /[^"\\]/
       | "\\\"" -> escape_quote
       | "\\\\" -> escape_backslash
```
But that allows %ignore rules into the string, which is very much unwanted. Would it be possible to add some kind of notation to prohibit %ignores inside particular sections? Some options would be surrounding a set of rules with some pragma such as `%ignore_off` … `%ignore_on` or `%protect` … `%unprotect`, or adding some prefix for space-sensitive rules, such as `~string: …`.

One thing that needs to be considered if this were to be added is how to deal with post-lexers' `always_accept` and similar features. Most usecases I can think of are similar to %ignore and should thus be restricted, but there might be cases I haven't thought of.

Another consideration is maybe not just having a blanket "stop all ignore rules", having some kind of more fine-grained control. It's not unthinkable that one could want a whitespace-sensitive language, but containing sections of lisp or something, which has a *different* set of ignore rules.

erezsh:
I can see how more fine grained control of %ignore might be useful.

I can suggest adding the following syntax:

```ruby
%ignore WS, COMMENT {
  // list of rules

  %ignore COMMENT2, -WS {
	// list of rules, ignoring only COMMENT and COMMENT2
  }
} 
```

I don't think `always_accept` would pose a big hindrance. Actually, adding a 'scoped ignore' feature might be a way to get rid of this hack.

However, that's all in theory. In practice, I'm not sure 'scoped ignore' would be an easy feature to add.

But I'll be happy to accept such a PR, and help in any way I can.

Caagr98:
[...]

PR #631, opened 2020-07-24
Scoped ignore

Caagr98:
Allows different `%ignore` rules across different parts of the grammar, as per #629.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -778,7 +778,7 @@ class GrammarLoader:
     def __init__(self, re_module):
         terminals = [TerminalDef(name, PatternRE(value)) for name, value in TERMINALS.items()]
 
-        rules = [options_from_rule(name, None, (), x) for name, x in  RULES.items()]
+        rules = [options_from_rule(name, None, ['WS', 'COMMENT'], x) for name, x in  RULES.items()]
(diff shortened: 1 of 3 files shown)


MegaIng:
Note for tests: All test that affect all parsers (which I think this should do?) Should be in test_parser.py, since those are automatically tested for every lexer/parse combinations. You can just add `skipIf` if it doesn't work for some parser/lexer combinations.

lark/load_grammar.py
```diff
@@ -214,7 +219,7 @@ def expr(self, rule, op, *args):
         assert False, op
 
     def maybe(self, rule):
-        keep_all_tokens = self.rule_options and self.rule_options.keep_all_tokens
```
MegaIng, line 217: Why was this change made?

Caagr98:
Yeah, I'll put them in there later. This way is much faster while developing though, since I don't need to re-run the full suite all the time.

As for that line you asked about (can't find the comment for some reason), I changed `rule_options` so it can't be `None` (since I needed to be able to pass around the ignore sets), so that check is unnecessary.

MegaIng:
You can run selective tests via `python -m tests -k whatever`. This will only run tests that contain whatever. Add `-v` for better output and `-f` to exit on first fail.

erezsh:
Closing for now. Feel free to reopen if you make any progress on this.

PR #636, opened 2020-07-28
Fixes caching when custom lexers are used

### LINKED ISSUE #613: Lark cache fails when custom lexers are used
**Describe the bug**

In parser_frontends.py, the passed lexer is wrapped using functools.partial: 
```
return partial(LALR_CustomLexer, lexer)
```

but this results in a functools.partial object, not a lexer object, so when deserialize is called later, it fails with
```
AttributeError: 'functools.partial' object has no attribute 'deserialize'
--> 338         self.parser = self.parser_class.deserialize(data['parser'], memo, self._callbacks, self.options.postlex, self.re)
```

**To Reproduce**
I've modified the custom lexer example

```
from lark import Lark, Transformer, v_args
from lark.lexer import Lexer, Token

class TypeLexer(Lexer):
    def __init__(self, lexer_conf, re_=None):
        pass

    def lex(self, data):
        for obj in data:
            if isinstance(obj, int):
                yield Token('INT', obj)
            elif isinstance(obj, (type(''), type(u''))):
                yield Token('STR', obj)
            else:
                raise TypeError(obj)

grammar = """
        start: data_item+
        data_item: STR INT*
        %declare STR INT
        """
parser = Lark(grammar, parser='lalr', lexer=TypeLexer, cache=True)

# Fails
parser = Lark(grammar, parser='lalr', lexer=TypeLexer, cache=True)
```

adefazio:
Fixes #613

### DIFF
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -55,6 +55,7 @@ class Lark:
         propagate_positions: bool = False,
         maybe_placeholders: bool = False,
         lexer_callbacks: Optional[Dict[str, Callable[[Token], Token]]] = None,
+        cache: Union[bool, str] = False,
         g_regex_flags: int = ...
     ):
         ...
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -1,5 +1,3 @@
-from functools import partial
-
 from .utils import get_regexp_width, Serialize
 from .parsers.grammar_analysis import GrammarAnalyzer
 from .lexer import TraditionalLexer, ContextualLexer, Lexer, Token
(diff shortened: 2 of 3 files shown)


erezsh:
The patch looks good.

But for some reason I was sure we already merged something like this.. I wonder why

MegaIng:
I also thought the same, but apparently not.

erezsh:
@adefazio would it be okay if I asked you to rebase this into two commits? Things like "newline" kind of clutter the log.

I can do it myself if you prefer, but I think that might mess with github's references.

adefazio:
> @adefazio would it be okay if I asked you to rebase this into two commits? Things like "newline" kind of clutter the log.
> 
> I can do it myself if you prefer, but I think that might mess with github's references.

Ok, I'll do that.

adefazio:
Commits are now flattened

erezsh:
Thanks!

PR #638, opened 2020-08-03
For Nearley: add argparse, extras install, allow es6 compiling

TMiguelT:
Some improvements to the Nearley converter:

* Use argparse instead of manually parsing `argv`. This gives us a `--help` output, and better error detection for free
* Include `js2py` as an "extras" dependency (installed using `pip install lark-parser[nearley]`)
* Add a `--es6` flag to this command, which enables the experimental ES6 support Js2Py
* The changes to the interface are documented in the readme

Todo:
* [ ] Add an explanation of `nearley_lib` in argparse. I don't actually know what this parameter does, so I can't document it without help

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,11 @@ Using Lark? Send me a message and I'll add your project!
 
 Lark comes with a tool to convert grammars from [Nearley](https://github.com/Hardmath123/nearley), a popular Earley library for Javascript. It uses [Js2Py](https://github.com/PiotrDabkowski/Js2Py) to convert and run the Javascript postprocessing code segments.
 
+First, ensure you have Lark installed with the `nearley` component included:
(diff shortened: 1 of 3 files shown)


erezsh:
Thanks!

The `nearley_lib` is just there to support importing builtins.

TMiguelT:
Thanks for the merge!

> The nearley_lib is just there to support importing builtins.

I'm not quite sure what that means. Do you think you could add a sentence to the argparse parser to clarify? I think that would be quite helpful.

erezsh:
Done

erezsh:
Nearley comes with these builtins: cow.ne  number.ne  postprocessors.ne  string.ne  whitespace.ne

So we want to support including their most recent version

Issue #300, opened 2018-12-24
Support the verbose flag for regexps


erezsh:
Example test:

```python
         def test_token_flags_verbose(self):
             g = _Lark(r"""start: NL | ABC
                           ABC: / [a-z] /x
                           NL: /\n/
                       """)
             x = g.parse('a')
             self.assertEqual(x.children, ['a'])
```

erezsh:
Supported!

PR #645, opened 2020-08-10
Support multiline regular expressions if they are marked with `x` (verbose) flag

decorator-factory:
Now you can do this:
```make
rule: /
      a     b  c d   # Foo bar baz 
     e   f  g h
     /x
```


decorator-factory:
Implementation of https://github.com/lark-parser/lark/issues/300

erezsh:
Looks good!

Just squash the first 3 commits together, and remove the unused import.

decorator-factory:
I hope I didn't screw up during the rebase wizardry. It seems that I didn't.

erezsh:
Looks good enough!

Thanks for the PR 👍

PR #647, opened 2020-08-12
Error handling

### LINKED ISSUE #644: Enhancement: Higher Importance for custom error handling.
Lark's current error handling method loops through a dictionary, with the first items in the dict having higher importance than the last. 
Instead of this current method, implement a new method which allows users to set a higher importance for errors. It helps make code much more organized and easier to understand.

**But, where would this come in play?**
Let's say I have a grammar where all expressions would end in a semicolon.
Here is the grammar:
```py
exc_class = u.match_example( parser.parse, {
        MissingBracket : [
          '( 1 + 1',
          '( 1 + 1 ;',
          '1 + 1 )'
        ],
        MissingSemicolon : [
          '1 + 1',
          '1 + 1;;'
        ],
        })
```
But when I parse this code: `1 + 1`, it would raise a MissingBracket instead of a MissingSemicolon
Currently, the only way to fix this is by having the less important error matches pushed to the back like so:
```py
exc_class = u.match_example( parser.parse, {
        MissingBracket : [
          '( 1 + 1',
          '( 1 + 1 ;',
        ],
        MissingSemicolon : [
          '1 + 1',
          '1 + 1;;'
        ],
        MissingBracket : [
          '1 + 1 )'
        ]
        })
```
However, through this way you would have less organized code, and it would be much simpler having some sort of way to set a error example with more importance.

### LINKED ISSUE #646: Incorrect Expected Token
In the list of Expected Tokens of the `UnexpectedToken`, some tokens appear that aren't valid

```python

from lark import Lark, Token

grammar = """
start : term ";"
term : "a"
      | "(" term ")"
"""

parser = Lark(grammar, parser='lalr')  # lark parser

parser.parse("(a")
```

raises

```
  File "C:\Users\tramp\PycharmProjects\lark\lark\parsers\lalr_parser.py", line 60, in get_action
    return states[state][token.type]
KeyError: '$END'

During handling of the above exception, another exception occurred:

  File "C:/Users/tramp/PycharmProjects/lark/test.py", line 106, in <module>
    parser.parse("(a")
  File "C:\Users\tramp\PycharmProjects\lark\lark\lark.py", line 400, in parse
    return self.parser.parse(text, start=start)
  File "C:\Users\tramp\PycharmProjects\lark\lark\parser_frontends.py", line 127, in parse
    return self._parse(token_stream, start, set_parser_state)
  File "C:\Users\tramp\PycharmProjects\lark\lark\parser_frontends.py", line 58, in _parse
    return self.parser.parse(input, start, *args)
  File "C:\Users\tramp\PycharmProjects\lark\lark\parsers\lalr_parser.py", line 36, in parse
    return self.parser.parse(*args)
  File "C:\Users\tramp\PycharmProjects\lark\lark\parsers\lalr_parser.py", line 112, in parse
    _action, arg = get_action(token)
  File "C:\Users\tramp\PycharmProjects\lark\lark\parsers\lalr_parser.py", line 67, in get_action
    raise UnexpectedToken(token, expected, state=state, puppet=puppet)
lark.exceptions.UnexpectedToken: Unexpected token Token($END, '') at line 1, column 2.
Expected one of: 
	* SEMICOLON
	* RPAR
```
Despite `SEMICOLON` not being valid.

Same behavior for `puppet.choices()`

MegaIng:
Closes #646 and closes #644. 

Has two big changes:

- Allows both dictionaries and any iterable of tuple as example for `match_example`
- Adds an `accepts` list to `UnexpectedToken` and uses it in `match_example` to restrict the context more precisely. (Might be a 'breaking' change if we consider the behavior clearly defined.)

### DIFF
--- a/lark-stubs/exceptions.pyi
+++ b/lark-stubs/exceptions.pyi
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from typing import Dict, Iterable, Callable, Union, TypeVar, Tuple
+from typing import Dict, Iterable, Callable, Union, TypeVar, Tuple, Any, List, Set
 from .tree import Tree
 from .lexer import Token
 
@@ -25,7 +25,10 @@ T = TypeVar('T')
 
 
 class UnexpectedInput(LarkError):
+    line: int
+    column: int
     pos_in_stream: int
+    state: Any
 
     def get_context(self, text: str, span: int = ...):
         ...
@@ -41,12 +44,14 @@ class UnexpectedInput(LarkError):
 
 
 class UnexpectedToken(ParseError, UnexpectedInput):
-    pass
-
+    expected: List[str]
+    considered_rules: Set[str]
+    puppet: Any
+    accepts: List[str]
 
 class UnexpectedCharacters(LexError, UnexpectedInput):
-    line: int
-    column: int
+    allowed: Set[str]
+    considered_tokens: Set[Any]
 
 
 class VisitError(LarkError):
--- a/lark-stubs/exceptions.pyi
+++ b/lark-stubs/exceptions.pyi
@@ -3,7 +3,7 @@
 from typing import Dict, Iterable, Callable, Union, TypeVar, Tuple, Any, List, Set
 from .tree import Tree
 from .lexer import Token
-
+from .parsers.lalr_puppet import ParserPuppet
 
 class LarkError(Exception):
     pass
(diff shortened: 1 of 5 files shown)


erezsh:
I'm assuming this is a draft PR, because of the prints.

I imagined `accepts` as a method of `ParserPuppet`.

Also maybe make the use of `accepts` in `match_examples` optional for now. It is a breaking change after all.

MegaIng:
Nope, not a draft: the prints allow the grammar creator to more precisely know what is going in inside `match_examples`.

The reason I added `accepts` to the `UnexpectedToken` is that I also use it to improve the output the `Expected` list. But of course I can either remove the attribute or also add a method to the puppet, depending on what you want.

Yeah, ok.

MegaIng:
I'll move the debug prints to the logger. That is a better idea.

MegaIng:
Is there still a problem with #607? I was trying to base my log message on it. I though it was already merged.

erezsh:
I just didn't get around to test it yet. I wanted to make sure it doesn't mess with the global logger in any way, or contain other breaking changes.

MegaIng:
> The reason I added accepts to the UnexpectedToken is that I also use it to improve the output the Expected list. But of course I can either remove the attribute or also add a method to the puppet, depending on what you want.

Opinion?

erezsh:
Yes, I agree, you should also make it accessable through UnexpectedToken

    def accepts(self):
       return self.puppet.accepts()

PR #650, opened 2020-08-13
Make transformer work with tokens in standalone parser, fixes #648

### LINKED ISSUE #648: Transformer is not applied for tokens in standalone parser
**Describe the bug**
Standalone parser doesn't invoke methods on tokens defined in supplied transformer

**To Reproduce**
Given:
- grammar `example.lark`:
```
start: IDENTIFIER "(" text ")"
IDENTIFIER: /[A-Za-z_][A-Za-z0-9_]*/
text: IDENTIFIER
```
- standalone parser generated with: `python3 -m lark.tools.standalone example.lark > example.py`

Then:
- this program:
```python
import lark

class DowncaseIdentifiers(lark.Transformer):
    def IDENTIFIER(self, token):
        return lark.Token("IDENTIFIER", token.lower())

    def text(self, children):
        return lark.Tree("text", [self.IDENTIFIER(children[0])])

parser = lark.Lark.open(
    grammar_filename="example.lark", transformer=DowncaseIdentifiers(), parser="lalr"
)
to_parse = "FOO(BAR)"
result = parser.parse(to_parse)
print(result.pretty())
```
produces:
```
start
  foo
  text  bar
```
- this program:
```python
import example

class DowncaseIdentifiers(example.Transformer):
    def IDENTIFIER(self, token):
        return example.Token("IDENTIFIER", token.lower())

    def text(self, children):
        return example.Tree("text", [self.IDENTIFIER(children[0])])

parser = example.Lark_StandAlone(transformer=DowncaseIdentifiers())
to_parse = "FOO(BAR)"
result = parser.parse(to_parse)
print(result.pretty())
```
produces:
```
start
  FOO
  text  bar
```

Expected:
Both programs produce the same output, namely:
```
start
  foo
  text  bar
```

BlankSpruce:
I'm not 100% certain if this is the proper way to recreate lexer callbacks so any advice would be much appreciated.

### DIFF
--- a/lark/common.py
+++ b/lark/common.py
@@ -17,9 +17,6 @@ class LexerConf(Serialize):
         self.skip_validation = skip_validation
         self.use_bytes = use_bytes
 
-    def _deserialize(self):
-        self.callbacks = {} # TODO
-
 ###}
 
 class ParserConf:
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -344,7 +344,14 @@ class Lark(Serialize):
         self.rules = [Rule.deserialize(r, memo) for r in data['rules']]
         self.source = '<deserialized>'
         self._prepare_callbacks()
-        self.parser = self.parser_class.deserialize(data['parser'], memo, self._callbacks, self.options.postlex, re_module)
+        self.parser = self.parser_class.deserialize(
+            data['parser'],
+            memo,
+            self._callbacks,
+            self.options.postlex,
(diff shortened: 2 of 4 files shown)


BlankSpruce:
Removed not needed import.

erezsh:
Thanks for the PR!

erezsh:
P.S. Just so you know, I added a link to gersemi in the README

Issue #455, opened 2019-10-02
Incorrect production of _ambig nodes when ambiguity="explicit"


FrancescoManfredi:
Here what happens (it only happened in this case and I can't imagine why): the parser returns a parsing three containing a different phrase than what was given as input to the .parse method. Moreover there is an ambiguity in the input phrase and this isn't represented. I'm fairly sure the second thing is causing the first one.

I'm attaching a test to reproduce this behavior:
https://pastebin.com/yJW66hu6

Thanks for building this and thanks in advance if you take this issue into consideration.

MegaIng:
I am not the maintainer, but here a bit of general advice to make it easy for everyone involved:
- Post the difference between what you get and what you want. What do you mean with 'different phrase than what was given as input to the .parse method.'? Please shown an example of this. 
- Your grammar is really big, so it would really help if you try to reduce it to an minimal example. 
- If you don't manage to do this, it would also help if you point out where the ambiguity you know of is in the grammar.

FrancescoManfredi:
Thanks for you answer. You are right, I can be more clear. Sorry but this is the first time I submit an issue on GH.
I'm going to add more details:

When I parse the sentence: 
**"c'è probabilità di nevicate domani"**

```
I get the following parse tree:
start
  weather_question
    weather_question_int_13
      question_snow
        question_snow_int_53
          motion_loc
            motion_loc_int_1	c'è
          chance
            chance_int_0
              probabilità
              di
          query_snow
            query_snow_int_4	nevicate
          motion_loc
            motion_loc_int_1	c'è
          chance
            chance_int_0	probabilità
          query_snow
            query_snow_int_4
              di
              nevicate
          future_point_in_time
            future_point_in_time_int_0	domani
```
which would be correct if I parsed the sentence: "**c'è probabilità di nevicate c'è probabilità di nevicate domani**".

About the **ambiguity**: 
**chance_int_0** has the ability to match "**probabilità di**";
**query_snow_int_4** has the ability to match "**di nevicate**";
The expression "**chance_int_0 query_snow_int_4**" (not directly present in the grammar but achievable as "chance query_snow") can therefore match "**probabilità di nevicate**" in two different ways depending on which one produces the "**di**" part.
The parse tree shown above somehow shows that there is an ambiguity, but it **does not introduces an _ambig node in the point where the ambiguity arises**.

I'm sorry about the size of the grammar, but the original one was even bigger and I managed to extract just the minimum required to run the test.
[...]

PR #651, opened 2020-08-13
Prevent invalid tree with ambiguity="explicit"

chanicpanic:
This mitigates #455 by making sure that Lark.parse() with ambiguity="explicit" returns a valid tree even if it cannot process all ambiguities in the SPPF. A warning of this occurrence has been added when debug=True.

Example:

```python
from lark import Lark

grammar = """
start: ab bc d?
!ab: "A" "B"?
!bc: "B"? "C"
!d: "D"
"""

p = Lark(grammar, ambiguity="explicit", debug=True)

tree = p.parse("ABCD")
print(tree.pretty())
```

One possible current output (invalid tree):
```
start
  ab
    A
    B
  bc    C
  ab    A
  bc
    B
    C
  d     D
```

One possible new output (valid tree):
```
WARNING:root:Ambiguous intermediate node in the SPPF: (start ::= ab bc* d, 0, 3, 2). Lark does not currently process these ambiguities: (resolving with the first derivation)
start
  ab    A
  bc
    B
    C
  d     D
```

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -314,7 +314,7 @@ class Parser:
 
         # Perform our SPPF -> AST conversion using the right ForestVisitor.
         forest_tree_visitor_cls = ForestToTreeVisitor if self.resolve_ambiguity else ForestToAmbiguousTreeVisitor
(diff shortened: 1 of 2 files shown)


erezsh:
Thanks for the PR!

I added slightly different code instead, but with the same effect.

See the warning using `lark.logger.setLevel(0)`

PR #652, opened 2020-08-14
ContextualLexer/UnexpectedInput puppet

MegaIng:
This is the start of a solution for the bug report that came up in the gitter.im chat. Note that this does really solve it! It is still impossible (I think) to actually skip over UnexpecedCharacters (or the Tokens the ContextualLexer creates from them).

Also, the tests fail with a for me not understandable error message.

### DIFF
--- a/lark-stubs/exceptions.pyi
+++ b/lark-stubs/exceptions.pyi
@@ -29,6 +29,7 @@ class UnexpectedInput(LarkError):
     column: int
     pos_in_stream: int
     state: Any
+    puppet: ParserPuppet
 
     def get_context(self, text: str, span: int = ...):
         ...
@@ -46,7 +47,6 @@ class UnexpectedInput(LarkError):
 class UnexpectedToken(ParseError, UnexpectedInput):
     expected: Set[str]
     considered_rules: Set[str]
-    puppet: ParserPuppet
     accepts: Set[str]
 
 class UnexpectedCharacters(LexError, UnexpectedInput):
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -4,6 +4,8 @@ from typing import (
     TypeVar, Type, List, Dict, IO, Iterator, Callable, Union, Optional,
     Literal, Protocol,
 )
+
+from .exceptions import UnexpectedToken
 from .visitors import Transformer
 from .lexer import Token, Lexer, TerminalDef
 from .tree import Tree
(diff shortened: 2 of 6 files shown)


erezsh:
Regarding the design, what if we define a "Resumeable" interface that both the lexer and parser adhere to?

Something like

```python
class Resumable:
    def get_resumable_capsule(self):
        "Returns a capsule of data for resuming self"
        pass

    @classmethod
    def resume(cls, capsule):
        "Resume execution from given point"

...

class ContextualLexer(Resumable):
    ...

class LalrParser(Resumable):
    ...
```

Something like that. What do you think?

MegaIng:
But the intersting aspect would be how to modify this resumable State. For the parser, we can just use the Puppet with a few modifications (maybe scrolling back might be interesting? For discarding open brackets or something).

The Lexer State should also be relativly simple to create an interface for.

But yes, the seems like a great idea.

erezsh:
Perhaps a similar LexerPuppet with methods like "feed_text", "move_cursor", etc.

erezsh:
Context:

**MegaIng @MegaIng Aug 14 13:18**

```python
from lark import Lark

grammar = """
start: A+
A: "a"
"""

def handle_undefined(e):
    return True


parser = Lark(grammar, parser='lalr')

tree = parser.parse("aabab", on_error = handle_undefined)
print(tree)
```
Will simple stop parsing without any kind of error message
(and only return the tree till the first b)
(assuming you are using my PR)

**Erez Shinan @erezsh Aug 14 13:37**

Right, because the UnexpectedCharacters exception interrupts the lexer loop, the stream to the parser ends. There is no mechanism to tell the lexer to continue lexing from where it stopped
So it seems that we have to either add a "resume" mechanism to the lexer, or transmit the error to the parser without raising (by yielding the exception?)
Resuming shouldn't be too hard, I think. You just have to save the LineCounter and pass it back to the lexer when it resumes. (Unless I'm forgetting something)
But both (resuming vs. yielding error) seems like reasonable and actionable solutions to me

**MegaIng @MegaIng Aug 14 14:06**

The problem with yielding is that I am not sure how to best pass information about skipped characters. As example, we want to skip the rest of the line, how do we pass that information to the lexer?
This should be simple for the developer to modify, depending on all available information (state, value_stack, accepted, ...)
(And I am still thinking about revamping a bit of the lexer infrastructure to allow custom contextual lexer. This could also be used to allow lexer that return an object, and not a function generator, allowing for inspection/overriding of the lexer state)

MegaIng:
Better progress in #700.

PR #653, opened 2020-08-14
Add missing elements in standalone parser

BlankSpruce:
Add:
- missing imports
- missing `Enumerator` class
- missing `_serialize` function
- `__version__` variable

Additionally regenerated json parser example


erezsh:
Why do you need to serialize from the Standalone parser?

BlankSpruce:
Serialization might be still needed for multiprocessing purposes. That being said I'll check if, at least in my case, it could be somehow mitigated.

BlankSpruce:
I've amended the change. Lack of `_serialize` method is not an issue which can't be easily mitigated. Thanks for hint.

erezsh:
Thanks for the PR!

PR #657, opened 2020-08-16
Postproc option for reconstruct, and fixed isalnum bug with unicode identifiers

julienmalard:


erezsh:
Why not

```python
if post_proc:
    x = post_proc(x)
```
?

erezsh:
Also, where can I find a list of all available categories? Python docs aren't being helpful.

julienmalard:
@erezsh : Good idea! For the identifier unicode class categories, they are buried deep in a PEP from 2007: https://www.python.org/dev/peps/pep-3131/#specification-of-language-changes

erezsh:
Thanks!

julienmalard:
Thank you!

PR #661, opened 2020-08-17
Handle ambiguous intermediate nodes in the SPPF

### LINKED ISSUE #238: Ambiguity error for simple grammar
Hi,

I recently stumbled upon an error, where ``lark`` is missing some ambiguous solutions. I tried to create a minimal example to make the problem easier to understand. The code is the following:

#### Code

```python
import lark

simple_string = 'aaab'
simple_grammar = '''
    s: as? as? b
    as: a a?

    ?a: /a/
    ?b: /b/
    '''
simple_parser = lark.Lark(simple_grammar, start='s', ambiguity='explicit')
print(simple_parser.parse(simple_string).pretty())
```

There is an ambiguity in the first rule for matching **aaab**. Either the first non-terminal ``as`` is resolved to the terminals **aa** or the second non-terminal ``as`` is, leaving the other one to be resolved to a single **a** and resulting in two solutions for this string-grammar combination. However, the output of ``lark`` is the following indicating that there is no ambiguity:

#### Output

```
s
  as
    a
  as
    a
    a
  b
```

Whereas the expected output is the following:

#### Expected Output

```
_ambig
  s
    as
      a
    as
      a
      a
    b
  s
    as
      a
      a
    as
      a
    b
```

I verified the ambiguity of this simple example using a different parser. Finally, if I change the ``b`` in the first rule from a non-terminal to the corresponding terminal, I get the right output.

chanicpanic:
This allows ambiguous intermediate nodes to be properly unpacked from the SPPF. Thus, parsing with earley will return an ambiguous parse tree representing all possible derivations of the input string (assuming that the number of derivations is finite). This fixes #238 and #455 with respect to the earley parser.

**Summary of Approach**

The `CompleteForestToAmbiguousTreeVisitor` class has been added as an improved version of the `ForestToAmbiguousTreeVisitor` class. During the forest walk, auxiliary nodes are added to the tree to represent ambiguous intermediate nodes and their children. These auxiliary nodes are processed into a single '\_ambig' node for their nearest symbol node ancestor on the walk up via  the callbacks passed to `ForestToTreeVisitor`. 

The callback is implemented in the `AmbiguousIntermediateExpander` class in  `parse_tree_builder.py`. It has been added as the first class in the node builder chain for explicit ambiguity parses. After flattening the ambiguous intermediate nodes, it calls the next builder in the chain to take care of other tree shaping.

**Tests**

Tests have been added to the earley test class to check the structure and contents of parse trees resulting from SPPFs with ambiguous intermediate nodes.

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -195,6 +195,86 @@ def maybe_create_ambiguous_expander(tree_class, expansion, keep_all_tokens):
     if to_expand:
         return partial(AmbiguousExpander, to_expand, tree_class)
 
+class AmbiguousIntermediateExpander:
+    """
+    Propagate ambiguous intermediate nodes and their derivations up to the
+    current rule.
+
+    In general, converts
+
+    rule
+      _iambig
+        _inter
+          someChildren1
+          ...
+        _inter
+          someChildren2
+          ...
+      someChildren3
+      ...
+
+    to
+
+    _ambig
+      rule
+        someChildren1
+        ...
+        someChildren3
+        ...
+      rule
+        someChildren2
+        ...
+        someChildren3
+        ...
+      rule
+        childrenFromNestedIambigs
+        ...
+        someChildren3
+        ...
+      ...
+
+    propagating up any nested '_iambig' nodes along the way.
+    """
+
+    def __init__(self, tree_class, node_builder):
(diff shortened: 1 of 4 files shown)


erezsh:
Wow, that's awesome, thanks!!

Great PR too 👍

PR #664, opened 2020-08-19
corrected caret placement in error messages due to tabs

### LINKED ISSUE #663: caret indicator in error messages doesn't take into account tabs in line
run this code and look at the output error message:
```python
from lark import Lark

grammar = '''
%import common.WS
%ignore WS

start: "a"
'''

p = Lark(grammar)

p.parse("\tb")
```

for me, this outputs:
```
...
lark.exceptions.UnexpectedCharacters: No terminal defined for 'b' at line 1 col 2

        b
 ^

Expecting: {'A'}
```

note that the caret is not properly lined up with the incorrect character "b"
there are 8 spaces in front of the b in the output, but, 1 space in front of the caret
(8 is probably because it's the tab size for cmd.exe, and i'm sure the number varies per environment)

if spaces are used instead, the caret lines up properly.

if you replace the `\t` in `p.parse("\tb")` with an embedded tab character:
```
p.parse("	b")
```
you get the same result

i think lark is treating the tab as a one-width character, when it should be expanded to some number of spaces.

I'm on windows 10, 64-bit
Python 3.8.3, 64-bit
Lark 0.9.0

starwarswii:
fixes #663

### DIFF
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -49,11 +49,11 @@ class UnexpectedInput(LarkError):
         if not isinstance(text, bytes):
             before = text[start:pos].rsplit('\n', 1)[-1]
             after = text[pos:end].split('\n', 1)[0]
-            return before + after + '\n' + ' ' * len(before) + '^\n'
+            return before + after + '\n' + ' ' * len(before.expandtabs()) + '^\n'
         else:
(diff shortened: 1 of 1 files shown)


erezsh:
Looks good, thanks!

PR #667, opened 2020-08-19
Refactored reconstructor out into tree_matcher

erezsh:
Functionality should stay the same.

### DIFF
--- a/lark/tree_matcher.py
+++ b/lark/tree_matcher.py
@@ -3,10 +3,10 @@
 import re
 from collections import defaultdict
 
-from lark import Tree, Token
-from lark.common import ParserConf
-from lark.parsers import earley
-from lark.grammar import Rule, Terminal, NonTerminal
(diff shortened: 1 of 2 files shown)


lark/tree_matcher.py
```diff
@@ -0,0 +1,177 @@
...
+            for alias in rule_aliases:
+                yield make_recons_rule_to_term(origin, NonTerminal(alias))
+            yield make_recons_rule_to_term(origin, origin)
+
+    def match_tree(self, tree, rulename):
+        """Match the elements of `tree` to the symbols of rule `rulename`.
+
+        Args:
+            tree (Tree): the tree node to match
+            rulename ([type]): [description]
```
MegaIng, line 144: ...

erezsh: Whoops. Will fix, thanks!

PR #670, opened 2020-08-19
Added tree_validator.py, which allows validating an AST against a grammar

erezsh:


MegaIng:
The tests aren't being run.

erezsh:
Should work now.

MegaIng:
You a lot of TODO: ambiguity. What does this do with ambiguity?

erezsh:
Well, if the grammar is ambiguous, there's more than one way to validate it. So to support it, we'll have to take every possible path.

(unless we can think of a more efficient way)

MegaIng (approved):
I would suggest adding an example, when possible with the python grammar and a relatively complex program as a stress test.

erezsh:
Good idea, I did some more testing and figured out it's incorrect.

I might give it another go some day, maybe with a different approach.

Issue #673, opened 2020-08-21
Designing a new forest visitor interface


chanicpanic:
I noticed in the docs that there are plans to make processing the SPPF more user-friendly. After becoming familiar with the current implementation, I am interested in taking this on. Of course, I would like some input on what a good interface might look like.

**Notes on the current implementation**

1. `ForestVisitor` is an abstract base class for performing a depth-first walk of the SPPF with the `visit` method. 
2. Subclasses of `ForestVisitor` receive events from the super class via  specific method overrides. Namely,

  - `visit_token_node`
  - `visit_symbol_node_in`
  - `visit_symbol_node_out`
  - `visit_packed_node_in`
  - `visit_packed_node_out`

3. The walk is controlled by the return value of the `visit_symbol_node_in` and  `visit_packed_node_in` methods.

  - If `None` is returned, the visitor will begin to backtrack.
  - If a node is returned, it will be visited next.
  - If an iterator over nodes is returned, they will be visited in order.

4. The return value of `visit` is the value of `self.result`. Subclasses are expected to manipulate this value during the visit.

5. When an attempt is made to visit a node that is already being visited (i.e. a cycle), an exception is thrown by `ForestVisitor`. Recovering from this as a subclass is effectively impossible.

**My thoughts**

i. Visiting by subclassing is the right approach. This is consistent with  the Visitors and Transformers for trees.

ii. Handling nodes by overriding certain methods seems reasonable. The set  of methods is also functional. It could be expanded by adding separate methods for intermediate nodes (more reflective of the structure of the  SPPF). Or, an entirely different/higher level interface could be used.

iii. Controlling the walk could certainly be simplified/reworked, although I  am not sure what a good approach would be. At the very least, I would  add support for returning an iterable of nodes.

iv. The current way of setting a return value does not feel quite right. I have not thought of a good alternative at the moment.

v. I think the superclass should notify the subclass of a cycle. Perhaps through another method. That way, the subclass can decide how to proceed. 

vi. Does the current implementation do too much? Right now, it has aspects of all of Interpreter, Visitor, and Transformer for trees. Should these functionalities be separated?

erezsh:
I'll be happy to accept a PR that simplifies that interface. Generally I agree with what you wrote, but I don't feel like I have anything smart to say at the moment.

Anything that's more in-line with the Transformer/Interpreter of the Tree class seems like the right direction.

If you come up with some ideas for the alternative interface, perhaps I'll be able to comment and advise.

[...]

PR #675, opened 2020-08-24
New forest utilities

chanicpanic:
This improves the `ForestVisitor` and adds other Forest utilities. See #673.

**Summary of Changes**

- Small improvements and updates have been made to the existing `ForestVisitor` class. The most significant ones being the removal of a visit return value and improved handling of cycles.
- A new class, `ForestTransformer`, has been added as an extension of `ForestVisitor` to be the base class for bottom-up forest transformations.
- A specialized `ForestTransformer`, `ForestToParseTree`, replaces the old `ForestToTreeVisitor`, `ForestToAmbiguousTreeVisitor`, and `CompleteForestToAmbiguousTreeVisitor`.
- `TreeForestTransformer` has been written as a `ForestTransformer` with  an interface similar to the existing tree `Transformer`. The major difference is additional methods for ambiguity handling. While designed for controlling forest to tree transformations, clever subclass implementations may be able to perform arbitrary transformations.


lark/parsers/earley_forest.py
```diff
@@ -208,25 +234,130 @@ def visit(self, root):
...
+
+    def __init__(self):
+        # results of transformations
+        self.data = dict()
+        # used to track parent nodes
+        self.node_stack = deque()
+
+    def transform(self, root):
+        """Perform a transformation on a Forest."""
+        self.node_stack.append('result')
```
erezsh, line 296: What is the advantage of having `data['result']` rather than just `self.result`?

chanicpanic: In `visit*node_out`, transformations are appended to the list of data for the parent node. By using `data['result']` and pushing `'result'` onto `node_stack`, I don't have to check whether the stack is empty in all those methods (whether the node is `root`).

lark/parsers/earley_forest.py
```diff
@@ -208,25 +234,130 @@ def visit(self, root):
...
+
+    def visit_packed_node_in(self, node):
+        self.node_stack.append(id(node))
+        self.data[id(node)] = []
+        return node.children
+
+    def visit_token_node(self, node):
+        try:
+            self.data[self.node_stack[-1]].append(self.transform_token_node(node))
+        except Discard:
```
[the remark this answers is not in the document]
chanicpanic: Do you mean put the transformation attempt under `try` and the append under `else`?

chanicpanic:
A new PR with a cleaner history has been opened. #681

Issue #498, opened 2019-12-18
Raw graphviz dot output


felixvuo:
I think it would be nice to have the option to dump a raw graphviz DOT file instead of the sppf.png for a few reasons:

* the file size is a lot smaller
* it's nice to be able to "do things" with the resulting tree in its structured form, rather than navigate the rendered bitmap
* it's only one extra (or no extra) line and could be optional, e.g. if you've turned on "debug" mode.

Any thoughts?

erezsh:
I support the idea

PR #678, opened 2020-08-24
Graphivz dot output

omega16:
Now you can choose between write a png, write a dot file or get a pydot graph from your tree

### DIFF
--- a/examples/fruitflies.py
+++ b/examples/fruitflies.py
@@ -33,9 +33,13 @@ sentence = 'fruit flies like bananas'
 def make_png(filename):
     tree.pydot__tree_to_png( parser.parse(sentence), filename)
 
+def make_dot(filename):
+    tree.pydot__tree_to_dot( parser.parse(sentence), filename)
+
 if __name__ == '__main__':
     print(parser.parse(sentence).pretty())
     # make_png(sys.argv[1])
+    # make_dot(sys.argv[1])
 
 # Output:
 #
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Lark is great at handling ambiguity. Here is the result of parsing the phrase "f
 
 ![fruitflies.png](examples/fruitflies.png)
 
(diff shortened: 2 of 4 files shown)


omega16:
This solve issue #498

MegaIng:
You can just refactor all tree methods to the tree class, in the part that is not included in the standalone parser.

This will make it clearer: These functions aren't useful without the Tree. Just make sure to keep the original method for backwards compatibility.

Unless @erezsh disagree and says that these functions are for debugging, and should therefore be kept separated from the `Tree` class.

omega16:
I think more or less the same, only problem is backwards compatibility, so, marking pytdot_tree_to_png as deprecated, leave that function (for now).

new function names could be one of :


~~~python
write_dot
write_png
get_pydot_graph
get_graphviz_graph

save_dot
save_png

to_dot
to_png
to_graphviz_graph

tree_to_dot
tree_to_png
tree_to_graphviz

~~~

examples/fruitflies.py
```diff
@@ -33,9 +33,13 @@
 def make_png(filename):
     tree.pydot__tree_to_png( parser.parse(sentence), filename)
 
+def make_dot(filename):
+    tree.pydot__tree_to_dot( parser.parse(sentence), filename)
+
 if __name__ == '__main__':
     print(parser.parse(sentence).pretty())
     # make_png(sys.argv[1])
+    # make_dot(sys.argv[1])
```
erezsh, line 42: Why two comments?

erezsh (commented):
I'm ambivalent about it. If you want to move it to Tree, that's okay by me.

examples/fruitflies.py
omega16: It was to expose the use of dot output, but now it seems that defining `make_do` is enougth.

erezsh:
Thanks for the PR!

PR #680, opened 2020-08-28
Make Trees reconstructable from reprs

TG-Techie:
This request makes the strings returned from `repr(some_token_or_tree)` valid python code by changing the repr output to include quotes around a tree's 'data' or a token's 'type'.

### For example:
Let's take a possible parsed version of a binary-op for some language.
(shown with `.pretty()` for layout)
```
binop_expr
    expr
        var_expr    a
    binop
        add
    expr
        var_expr    b
```

Without the proposed changes the `repr` of the above Tree would look like:
```
Tree(binop_expr, [Tree(expr, [Tree(var_expr, [Token(NAME, 'a')])]), Tree(binop, [Tree(add, [])]), Tree(expr, [Tree(var_expr, [Token(NAME, 'b')])])])])
```

The change proposed would alter this to look like:
```
Tree('binop_expr', [Tree('expr', [Tree('var_expr', [Token('NAME', 'a')])]), Tree('binop', [Tree('add', [])]), Tree('expr', [Tree('var_expr', [Token('NAME', 'b')])])])])
```
now `Tree('binop_expr',` and `Token('NAME',` now have quotes around the data and type.

This change has is beneficial in two ways:
1. Primarily, It will allow trees to be stored as text in files without having to pickle the Tree Object. 
2. This change brings the `__repr__` methods of Token and Tree into compliance with what is suggested for the repr function. 

The repr spec"
https://docs.python.org/3.8/reference/datamodel.html?highlight=__repr__#object.__repr__


TG-Techie:
I've been unable to reproduce the build fail on 3.5 locally, I will continue to try to debug it. if there are any suggestions on how to fix it they would be greatly appreciated.

MegaIng:
@TG-Techie This error is not your fault, but @erezsh pushed without looking at test results. Is fixed in most recent master.

TG-Techie:
@MegaIng thank you. pulled the fix in

erezsh:
@MegaIng The tests passed just fine. It's an inconsistent error that stems from random ordering inside dicts pre-3.7.

erezsh:
@TG-Techie Any chance you can add a test for it at `test_trees.py`?

TG-Techie:
gladly!

TG-Techie:
@erezsh thank you for suggesting it, I wasn't sure how.

erezsh:
Thanks! 👍

PR #681, opened 2020-08-28
New forest utilities (2)

chanicpanic:
A better version of #675 ready for review.

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -75,6 +75,7 @@ class LarkOptions(Serialize):
             - "resolve" - The parser will automatically choose the simplest derivation
                         (it chooses consistently: greedy for tokens, non-greedy for rules)
             - "explicit": The parser will return all derivations wrapped in "_ambig" tree nodes (i.e. a forest).
+            - "forest": The parser will return the root of the shared packed parse forest.
 
     **=== Misc. / Domain Specific Options ===**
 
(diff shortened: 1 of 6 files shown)


erezsh:
Looks good! There's just one thing that's bothering me. `data` keeps being used as a tuple, which leads to "magic code" like:

```python
                children.append(data[0])
            if len(data) > 1:
                children.append(data[1])
```

and it's not really clear what's going on there. It seems to me that putting it in a dedicated class, and giving it a significant name, might help improve the code. What do you think?

chanicpanic:
Good idea. I have pushed changes for this.

erezsh:
@chanicpanic 

What does `make_get_path()` do?

It seems like its result isn't being used right now. Is it in preparation?

But also regarding the interface, why do you have to send a function? Why not just send `path` and iterate over it later if necessary?

chanicpanic:
@erezsh 

`make_get_path(node)` creates a `get_path` function to be passed to the `on_cycle` callback. `get_path` returns a list of the nodes in the cycle where the first element is equal to the node passed to `on_cycle`, and the last element is the node
visited most recently.

While `get_path` is not used in any current subclasses of `ForestVisitor`, I thought it might be useful in some cases to have the path of the cycle available. I opted to pass a function to compute it on demand as most of the time it would not be needed. However, as you point out, passing `path` would be simpler. Plus, it would give more information.

I am leaning toward this change. One thing, should a copy, `path[:]`, be passed to protect the original?

erezsh:
I think it would be a shame to needlessly copy the lists every time just in case someone subclasses it in the future. In general, changing arguments in-place is a bit "rude" in Python, so adding a small note about it should be enough. (probably in the on_cycle docstring? and ofc with a mention of the utility function `get_path_in_cycle` or similar)

chanicpanic:
Good to know. I will pass `path` instead of `get_path` and add a note to `on_cycle`.

chanicpanic:
The changes to `on_cycle` have been applied.

erezsh:
How did the tests pass Python2.7 while containing nonlocal? That's a real mystery..

erezsh:
Thanks for the great PR!

PR #688, opened 2020-09-15
Add print parameter to lark.tools.standalone.main

rec:


rec:
So you can call l.t.standalone.main() and capture the output.

erezsh:
Nice idea, thanks!

MegaIng:
Suggestion: rework tests. I will do this latter if noone else does.

rec:
I'd be willing to add a test for this - why not?

There didn't seem to be tests covering this but it would be easy enough to write one now...

erezsh:
Look in `lark\tests\test_tools.py`

PR #689, opened 2020-09-15
Test print argument to lark.tools.standalone.main

rec:


rec:
Oops, moment pls.

erezsh:
Nice, well done!

PR #691, opened 2020-09-18
updated lark.lark with two small fixes

starwarswii:
fixed lark.lark not accepting negative priorities
fixed lark.lark not accepting non-newline-terminated grammars

includes tests in examples/lark_grammar.py


examples/lark.lark
```diff
@@ -1,20 +1,20 @@
-start: (_item | _NL)*
+start: (_item _NL | _NL)* _item?
```
omega16, line 1: It seems like it could be as before 
```
start : (_item | _NL)*
```
And it would work with the changes since _item can't end on _NL

omega16: I have tested to leave **start** as in original and it seems to work.

```python
from lark import Lark

#The new grammar with first rule changed to  _start : (_item | _NL)*
with open("grammar.lark", "r") as g:
    grammar = g.read()


#The old grammar (current grammar in repo)
with open("original_grammar.lark", "r") as g:
    original_grammar = g.read()

parser = Lark(
        grammar,
        )

parser_original = Lark(
        original_grammar,
        )

#Here I change the -1 to 1 or original grammar would complain
#about it instead of the end of line
test_grammar = """
start: r
r.-1: X
X: "a" | "b" """


#v1=parser_original.parse(test_grammar)
#just test if new grammar modified can still handle the EOF without \n
v2=parser.parse(test_grammar) 
v1=parser_original.parse(original_grammar)
v2=parser.parse(original_grammar)
print(v1==v2)
```

starwarswii: doesn't that allow multiple items to come right after one another, without newlines in between?
e.g. `_item _item`

starwarswii: for tinkering, I've been using [the online ide](https://lark-parser.github.io/lark/ide/app.html) for it's immediate feedback. here's my test grammar:

```lark
start: (_item | nl)*

_item: A

A: "a"

nl: _NL
_NL: /(\r?\n)+\s*/
```

testing with that, keeping the start as before allows "aaaaa", while the new start requires newlines between

erezsh: Didn't test it, but perhaps this would work

```ruby
start: _item (_NL _item?)*
```

examples/lark_grammar.py
```diff
@@ -14,6 +14,7 @@
 
 grammar_files = [
     examples_path / 'lark.lark',
+    examples_path / 'no_newline_at_end.lark',
```
erezsh, line 17: This is not the right path for it. Either put it in `tests/` or as an internal string in this file.

erezsh:
Please rebase it. Such a small change only needs to be a single commit, or a couple at most.

If don't mind doing it myself, but then github won't connect it to this pull request.

examples/lark.lark
starwarswii: The 'problem' with that is that then the grammar won't except an empty file. The original lark.lark does, should the new one? Or is that actually a bug?

Maybe a more pressing issue, I think that start does not allow leading new lines before the first item in the file

erezsh: Then

```ruby
start: _item? (_NL _item?)*
```

starwarswii: that looks like it's correct. is there any sort of state-explosion problem with the optional `?` being inside the repetition `*` like that? i was initially concerned with that, but i'm not that familiar with the lark internals.

you could also flip it around, so it's more like my original proposal: `start: (_item? _NL)* _item?`.
is there any benefit/preference for putting the repetition at the start vs at the end?

erezsh: I think either way is pretty much the same. 

Everything inside the `(...)*` will be put in a new rule, so I wouldn't worry about combinatoric effects. But if you want to be sure, you can create `_maybe_item: item?` and use that. That will ensure no hoodwinks are going on. Again, I think it doesn't matter much either way.

starwarswii: alright. in that case, i think i will go ahead with `start: (_item? _NL)* _item?`
perhaps it could be more optimized, maybe even taking into account that `_NL eats multiple newlines, so after a `_NL` can only follow an `_item`, but i think this is good enough.

starwarswii:
fixing commit history now

starwarswii:
should be good to go

erezsh:
Great! Thanks for the PR

PR #693, opened 2020-09-19
Documentation and examples for earley_forest

chanicpanic:
Here is API reference for members of `earley_forest.py`.
Advanced examples have also been added to demonstrate the usage of `ForestVisitor` and `TreeForestTransformer`.


erezsh:
Awesome! Thanks for the PR!

PR #694, opened 2020-09-19
Reduce/Reduce conflict resolution hints

ldbo:
Resolving Reduce/Reduce conflicts with LALR can be tedious with complex grammar. This PR provides hints on potential rules generating the conflicts when they occur.

To determine a potential rule to be blamed for a conflict between two rules, the first common ancestor of these rules is found. This process is then repeated for each of the pairs of conflicting rules. This allows to add hints in the error message generated upon finding a conflict.

For example, with this grammar:

```
start: expression
expression: arithmetic_expression
          | call

arithmetic_expression: addition
addition: addition "+" primary
        | primary
primary: identifier
identifier: /[a-z]+/

call: ["Call"] identifier
```

the following error message is generated:

```
lark.exceptions.GrammarError: Reduce/Reduce collision in Terminal('$END') between the following rules:
		- <call : identifier>
		- <primary : identifier>

The conflict may come from
		- <call> and <primary>: <expression>
```

It helps to see that a lone `identifier` can be parsed as a `call` or as a `primary`, which raises a conflict when using both in `expression`.

### DIFF
--- a/lark/parsers/lalr_analysis.py
+++ b/lark/parsers/lalr_analysis.py
@@ -255,7 +255,7 @@ class LALR_Analyzer(GrammarAnalyzer):
                 actions[la] = (Shift, next_state.closure)
             for la, rules in state.lookaheads.items():
                 if len(rules) > 1:
-                    raise self.__build_reduce_reduce_error(la, rules)
+                    raise self.build_reduce_reduce_error(la, rules)
                 if la in actions:
                     if self.debug:
                         logger.warning('Shift/Reduce conflict for terminal %s: (resolving as shift)', la.name)
@@ -283,12 +283,12 @@ class LALR_Analyzer(GrammarAnalyzer):
             self.parse_table = IntParseTable.from_ParseTable(_parse_table)
 
     # Generation of conflict resolution hints
-    def __build_reduce_reduce_error(self, la, rules):
+    def build_reduce_reduce_error(self, la, rules):
         """Build a GrammarError with conflict resolution hints."""
         msg = "Reduce/Reduce collision in %s between the following rules:" % la
         msg += "".join("\n\t\t- %s" % rule for rule in rules)
 
-        common_ancestors = self.__find_closest_common_ancestors(rules)
+        common_ancestors = self.find_closest_common_ancestors(rules)
         msg += "\n\nThe conflict may come from"
 
         for rules, ancestors in common_ancestors.items():
@@ -298,7 +298,7 @@ class LALR_Analyzer(GrammarAnalyzer):
 
         return GrammarError(msg)
 
-    def __find_closest_common_ancestors(self, conflicting_rules):
+    def find_closest_common_ancestors(self, conflicting_rules):
         """
         Find the closest common ancestor of pairs of rules.
 
@@ -306,7 +306,7 @@ class LALR_Analyzer(GrammarAnalyzer):
         :returns dict: 2-frozenset of NonTerminal rule names as keys, set of
         NonTerminal ancestor names as values
         """
-        parents = self.__rules_parents()
+        parents = self.rules_parents()
         ancestors = {rule.origin: {rule.origin} for rule in conflicting_rules}
         oldest_ancestors = {rule.origin: {rule.origin}
                             for rule in conflicting_rules}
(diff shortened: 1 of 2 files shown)


lark/parsers/lalr_analysis.py
```diff
@@ -279,9 +282,97 @@ def compute_lalr1_states(self):
         else:
             self.parse_table = IntParseTable.from_ParseTable(_parse_table)
 
+    # Generation of conflict resolution hints
+    def __build_reduce_reduce_error(self, la, rules):
```
MegaIng, line 286: dunder `__names` are not used in Lark. There is not really a reason to used name mangling, and it will probably just create problems down the road with subclassing/testing.

lark/parsers/lalr_analysis.py
```diff
@@ -279,9 +282,97 @@ def compute_lalr1_states(self):
...
+        msg += "\n\nThe conflict may come from"
+
+        for rules, ancestors in common_ancestors.items():
+            a = ' or '.join("<%s>" % ancestor.name for ancestor in ancestors)
+            rule1, rule2 = sorted(rule.name for rule in rules)
+            msg += "\n\t\t- <%s> and <%s>: %s" % (rule1, rule2, a)
+
+        return GrammarError(msg)
+
+    def __find_closest_common_ancestors(self, conflicting_rules):
```
MegaIng, line 301: Again, I would prefer to not use `__dunder` names

lark/parsers/lalr_analysis.py
```diff
@@ -279,9 +282,97 @@ def compute_lalr1_states(self):
...
+        parents = self.__rules_parents()
+        ancestors = {rule.origin: {rule.origin} for rule in conflicting_rules}
+        oldest_ancestors = {rule.origin: {rule.origin}
+                            for rule in conflicting_rules}
+        common_ancestors = dict()
+        intersections_found = set()
+
+        n = len(conflicting_rules)
+        pairs_number = (n * (n - 1)) // 2
+        while len(common_ancestors) < pairs_number:
```
MegaIng, line 318: Is there a guarantee that we will always find that many pairs? Or could this turn into an infinite loop?

ldbo: 1. The number of pairs in an n-set is `comb(n, 2) = n! / (2! * (n-2)!) = (n * (n-1)) / 2`, so `pairs_number` is exactly the number of pairs in `conflicting_rules`
2. Shift/Shift conflict arise if two rules could be matched in the same context. In other words, the conflicting rules are descendants of the starting rules in the pseudo-tree representing the relations between the rules. In particular, every pair taken from `conflicting_rules` has at least one common ancestor: the starting rule.
3. `ancestors` will eventually contain all the ancestors of each of the conflicting rules. Indeed, at the beginning of each step `i >= 0`, all of the ancestors of `rule` at a distance  `<= i` of `rule` are contained in `ancestors[rule.origin]`
4. `ancestors_intersection` builds the sets of common ancestors of rules given as arguments

These points prove that there are always exactly `pairs_number` pairs (tell me if one of them is not obvious). However, point 2 assumes that it is proven that Lark provides a valid LALR parser. For a non-complying parser (for example one which would find a conflict for rules that are not effectively used, and which could then be in different connected components of the rules graph), there is no guarantee on the termination of this loop.

If you want me to support non-complying parsers, we could brutally majorise the number of steps with `len(conflicting_rules)`. This has the drawback of hiding potential flaws in other parts of the parser.

lark/parsers/lalr_analysis.py
ldbo: It's now fixed

lark/parsers/lalr_analysis.py
ldbo: Fixed as well

erezsh:
I tested it on a 200 line grammar. Sometimes it worked (but unfortunately the result wasn't useful for debugging). Other times it went into an infinite loop (I counted 20000 iterations and stopped it).

I'm afraid I can't accept this patch as is.

ldbo:
> I tested it on a 200 line grammar. Sometimes it worked (but unfortunately the result wasn't useful for debugging). Other times it went into an infinite loop (I counted 20000 iterations and stopped it).
> 
> Sorry but I can't accept this patch as is.

Could you give me the grammar which put it into an infinite loop ? I'm really interested because I've used it to debug a lot of conflicts, which spared me a lot of time, and it never went into an infinite loop.

erezsh:
I wish I could, but it's a grammar I'm doing for a client.

I'll try to find a smaller example for it that I can also share.

PR #698, opened 2020-09-21
Run tests against Python 3.9

michael-k:
3.9.0 final is expected on Monday, 2020-10-05, see https://www.python.org/dev/peps/pep-0596/#schedule


erezsh:
Good idea, thanks!

PR #700, opened 2020-09-22
Resumable - working towards completely resumable parser+lexer (from puppet)

erezsh:
This is the first refactoring in order to allow a resumable lexer in addition to the parser.

To support it, I had to completely change and refactor the interface between them (or it would end up pretty ugly). That's because the parser must have access to the lexer, so it can request its state, and later recreate a new `LexerThread` with the saved state.

I think I also improved a few things on the way, but it's not perfect yet.

But I want some input before I continue. Did I miss anything important? Maybe I should use different architecture or better names?

Notes:

The puppet mechanism is currently broken. Later on I will fix it, and it will be much shorter, because it will be able to reuse most of the code from `lalr_parser.py`.

It breaks the custom lexer interface, so maybe it will have to go into v1.0

I managed to keep the postlexer interface the same with a hack, but maybe it's better to change it too.

Despite the changes, it seems that performance hasn't suffered noticeably.

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -271,7 +271,7 @@ class Lexer(object):
     lex = NotImplemented
 
     def make_lexer_state(self, text):
-        line_ctr = LineCounter('\n')
+        line_ctr = LineCounter(b'\n' if isinstance(text, bytes) else '\n')
         return LexerState(text, line_ctr)
 
 
@@ -346,7 +346,8 @@ class TraditionalLexer(Lexer):
                 allowed = {v for m, tfi in self.mres for v in tfi.values()} - self.ignore_types
                 if not allowed:
                     allowed = {"<END-OF-FILE>"}
-                raise UnexpectedCharacters(lex_state.text, line_ctr.char_pos, line_ctr.line, line_ctr.column, allowed=allowed, token_history=lex_state.last_token and [lex_state.last_token])
+                raise UnexpectedCharacters(lex_state.text, line_ctr.char_pos, line_ctr.line, line_ctr.column,
+                                           allowed=allowed, token_history=lex_state.last_token and [lex_state.last_token])
 
             value, type_ = res
 
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -424,16 +424,10 @@ class ContextualLexer(Lexer):
         except EOFError:
             pass
         except UnexpectedCharacters as e:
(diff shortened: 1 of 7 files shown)


lark/lexer.py
```diff
@@ -409,25 +413,35 @@ def __init__(self, conf, states, always_accept=()):
...
+            root_match = self.root_lexer.match(lexer_state.text, e.pos_in_stream)
             if not root_match:
                 raise
 
             value, type_ = root_match
             t = Token(type_, value, e.pos_in_stream, e.line, e.column)
-            raise UnexpectedToken(t, e.allowed, state=e.state)
+            raise UnexpectedToken(t, e.allowed, state=parser_state.position)
+
+class LexerThread:
```
MegaIng, line 438: Why not add a `.lexer` attribute to `LexerState`? This is also makes is somewhat harder to completely build a `LexerState` by hand for whatever reason.

MegaIng (commented):
Something similar to `PostLexConector` for `CustomLexer` should be possible, maybe with an extended LexerState that includes the `Iterable` returned from `CustomLexer.lex` so that each call to `CustomLexerWrapper.lex` just calls `next` on it. Every Operation that isn't supported will just error.

We can just differentiate between them whether or not `CustomLexer.lex(ls, ps)` fails. Alternatively, a class attribute `new_style` has to be set to True, otherwise old style is expected. This can then just be flipped so that new style is expected by default.


I think that a lot of simplification in `parser_frontends.py` is possible, but that can also be done in a latter PR. (Change the system for `early` as well, and then get ride of a lot of unnecessary case distinctions)

lark/lexer.py
erezsh: I had a similar thought, but came to realize that they should be separate entities.

That way `ContextualLexer` can pass the `LexerState` instance to `TraditionalLexer` without confusion.

If it has a `.lexer` attribute, then it would have to be sometimes used and sometimes ignored.

erezsh:
Hmm how would you simplify it? It seems to me like these distinctions are necessary.

erezsh:
@MegaIng Added more changes.

I think the only tasks left are to:

1) Resume each parse with a new copy of lexer_state

2) Add some interface to puppet to let users control the lexer position

3) Fix compatibility with old CustomLexer interface

4) Refactor and cleanup in the aftermath (not sure if there's much left to clean up)

PR #702, opened 2020-09-23
Fix `debug` and  `cache` options incompatibility

ldbo:
The `debug` flag is not compatible with caching : on loading a cached parser, its `debug` option is always set to `False`. This PR solves this.


erezsh:
Why are you using a lower protocol version?

Here is how much more space that takes for the Python grammar:

```python
(Pdb) p len(pickle.dumps(d))             
299466
(Pdb) p len(pickle.dumps(d, protocol=2)) 
419512
```

ldbo:
> Why are you using a lower protocol version?
> 
> Here is how much more space that takes for the Python grammar:
> 
> ```python
> (Pdb) p len(pickle.dumps(d))             
> 299466
> (Pdb) p len(pickle.dumps(d, protocol=2)) 
> 419512
> ```

It is indeed a lower version for Python >= 3.0, but is actually higher for Python 2.7 which uses protocol 1 by default (and does not support versions > 2). The issue is that protocol 1 can't dump `RulePtr` objects because of the use of `__slots__`, so I had to use another protocol. I chose to use protocol 2 so that cached grammars are compatible with all the supported Python versions. That way, they can be used as precompiled resources, and shared between different Python versions.

Would you prefer to use `protocol = pickle.HIGHEST_PROTOCOL`?

erezsh:
> Would you prefer to use protocol = pickle.HIGHEST_PROTOCOL

Yep, that's the right thing to do in this case.

erezsh:
I don't think cached grammars should be shared between different versions. I can't really see a use-case for this.

ldbo:
Ok, it's fixed now

erezsh:
Thanks for the PR!

Issue #703, opened 2020-09-26
Handle resources when running from zipped archive


sinoroc:
For context, _Poetry_ depends on [_poetry-core_](https://pypi.org/project/poetry-core/) which itself uses _lark-parser_. I have been experimenting with running _poetry_ from a zipped archive (see [_`zipapp`_](https://docs.python.org/3/library/zipapp.html)). But apparently _lark_ relies a lot on things such as `__file__` and `__path__` to load resources (_package data_), and this is not possible when the code is running from a zipped archive.

The right way to do such loading of resources from _package data_ is to use something like [_`pkgutil.get_data()`_](https://docs.python.org/3/library/pkgutil.html#pkgutil.get_data) or now even better [_importlib-resources_](https://docs.python.org/3/library/importlib.html#module-importlib.resources).

To be honest I am not entirely sure yet, if there is anything that should be done in this project, or if the necessary changes can be contained to _poetry_. But my first impression is that some things have to be modified here.

**_References_**:

* https://github.com/python-poetry/poetry/issues/2965

MegaIng:
This is only a problem if you use `%import common.*` or similar to use the stdlib. This can be easily avoided by just not using it and instead implementing your own terminals. (But this should probably be changed)

The passage shown in the other passage is the part where Lark provides a helper method for loading grammars from files via `Lark.open`. This is not the responsibly of lark to change, since it isn't a dependency of Lark, but of the user.

erezsh:
I agree with @MegaIng, lark only uses __file__ when importing library files. 

How does zippapp work? Is it still possible open files within the package using `open()` and a path? And if so, why doesn't it provide `__file__` that matches the right path inside the zip?

MegaIng:
@erezsh Is there currently any way to customize from where grammars are imported? If yes, is there any documentation for it? If not, maybe we should add it to fix this, with a fallback on the current file based method. (I don't see how it would be possible to replace the current import mechanic with `importlib-resources`. Except `common.lark` which should probably have a fallback that works from within zipapps).

erezsh:
@MegaIng I don't think we have that in. This seems relevant: https://github.com/lark-parser/lark/issues/603

sinoroc:
> This is only a problem if you use `%import common.*` or similar to use the stdlib. This can be easily avoided by just not using it and instead implementing your own terminals. (But this should probably be changed)
> 
[...]

PR #704, opened 2020-09-26
Custom import sources

### LINKED ISSUE #603: Customization of base_paths for relative imports
Currently, the `base_paths` for relative grammar imports only base on following, if I understood correctly:

1. the directory where the grammar file(-like object) is (fetching from the grammar file path)
2. the directory where the main script is (fetching from `sys.modules['__main__'].__file__`

However, this limits the architecture of projects that are using `lark`.

Imagine we have the following architecture:

```
|- package/
    |- submodule1/
        |- parser.py            # with a string grammar wanting to import some 
                                # shared rules/terminals from the shared grammar
    |- submodule2/              # maybe I have another submodule that wants to 
                                # import some rules/terminals from the shared grammar
    |- shared/
        |- grammar/
            |- shared.lark      # some shared rules/terminals
                                # say we have a terminal `INT` defined for the example below
```

Well, we could probably use 2) listed above to do the relative imports?

```lark
// not sure if this works
%import .shared.grammar.shared.INT 
```

However, this doesn't run if the package is not running independently. For example, using `pytest`, since `sys.modules['__main__'].__file__` will become the path to `pytest`.

**Ideas/Proposals:**

1. Reading paths from an env variable, say `LARK_GRAMMAR_PATH` (just like `PYTHON_PATH`)
2. Opening the `source` argument from `Lark` construct to allow overridden from outside.
3. Allowing pre-loaded grammar by `load_grammar.load_grammar`, since this function already allows specification of `grammar_name`, where the `base_paths` can fetch from

    With the above architecture, say in `parser.py`, we can do:
    ```python
    from pathlib import Path
    SHARED_GRAMMER= str(Path(__file__).parent.parent.joinpath('shared', 'grammar', 'shared.lark').resolve())
    grammar = load_grammar.load_grammar(
        r"""
        start: INT

        %import .shared.INT
        """,
        SHARED_GRAMMER
    )
    parser = Lark(grammar, ...)
    ```
    
    Then in `Lark` construct, we need to tell if the grammar is pre-loaded:
    ```python
    from load_grammer import Grammar

    class Lark:
        def __init__(self, grammar, ...):
             if isinstance(grammar, Grammar):
                 self.grammar = grammar
                 self.source = '<pre-loaded>'
                 # may have to deal with caching separately in such case
             else:
                 # regular grammar loading
    ```

MegaIng:
Added a new option to `Lark`, `import_sources`.
This is a list of either:

- A string, representing a folder which will be directly searched for any grammars
- A callable of the form `(base_paths, grammar_path) -> (grammar_name, grammar_source)`
  - `base_paths`: A list of paths lark identified as source folders. Probably not useful in custom loaders
  - `grammar_path`: The joined relative path of the grammar file, including the extension
  - `grammar_name`: The name that will be used for the grammar. Should be the path, but doesn't have to.
  - `grammar_source`: The raw grammar source, not a file object.

(This branch also includes a change to use `pkgutil` for `common.lark`, closing #703)

This should also close #603.

Note: while not a draft, I would like for someone to check whether this works for zipapp before merging.

### DIFF
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -33,6 +33,13 @@ class LarkOptions:
     g_regex_flags: int
     use_bytes: bool
     import_sources: List[Union[str, Callable[[str, str], str]]]
+    source: Optional[str]
+
+
+class FromPackageLoader:
+    def __init__(self, pkg_name: str, search_paths: Tuple[str, ...] = ...): ...
+    
+    def __call__(self, base_paths: List[str], grammar_path: str) -> Tuple[str, str]: ...
 
 
 class Lark:
@@ -62,6 +69,7 @@ class Lark:
         g_regex_flags: int = ...,
         use_bytes: bool = False,
         import_sources: List[Union[str, Callable[[List[str], str], Tuple[str, str]]]] = ...,
+        source: Optional[str],
     ):
         ...
 
@@ -70,6 +78,10 @@ class Lark:
 
     @classmethod
     def open(cls: Type[_T], grammar_filename: str, rel_to: Optional[str] = None, **options) -> _T:
+        ...
+    
+    @classmethod
+    def open_from_package(cls: Type[_T], package: str, grammar_path: str, search_paths: Tuple[str, ...] = ..., **options) -> _T:
         ...
 
     def lex(self, text: str) -> Iterator[Token]:
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -50,7 +50,7 @@ class FromPackageLoader:
 
 class Lark:
     source_path: str
-    source_code: str
+    source_grammar: str
     options: LarkOptions
     lexer: Lexer
     terminals: List[TerminalDef]
@@ -75,7 +75,7 @@ class Lark:
         g_regex_flags: int = ...,
         use_bytes: bool = False,
         import_paths: List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]] = ...,
-        source_path: Optional[str],
+        source_path: Optional[str]=None,
     ):
         ...
 
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -2,7 +2,7 @@
 
 from typing import (
     TypeVar, Type, List, Dict, IO, Iterator, Callable, Union, Optional,
-    Literal, Protocol,
+    Literal, Protocol, Tuple,
 )
 from .visitors import Transformer
 from .lexer import Token, Lexer, TerminalDef
@@ -32,6 +32,7 @@ class LarkOptions:
     cache: Union[bool, str]
     g_regex_flags: int
     use_bytes: bool
+    import_sources: List[Union[str, Callable[[str, str], str]]]
 
 
 class Lark:
@@ -60,6 +61,7 @@ class Lark:
         cache: Union[bool, str] = False,
         g_regex_flags: int = ...,
         use_bytes: bool = False,
+        import_sources: List[Union[str, Callable[[List[str], str], Tuple[str, str]]]] = ...,
     ):
         ...
 
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -32,19 +32,19 @@ class LarkOptions:
     cache: Union[bool, str]
     g_regex_flags: int
     use_bytes: bool
-    import_sources: List[Union[str, Callable[[str, str], str]]]
-    source: Optional[str]
+    import_paths: List[Union[str, Callable[[Optional[str], str], Tuple[str, str]]]]
+    source_path: Optional[str]
 
 
 class FromPackageLoader:
     def __init__(self, pkg_name: str, search_paths: Tuple[str, ...] = ...): ...
     
-    def __call__(self, base_paths: List[str], grammar_path: str) -> Tuple[str, str]: ...
+    def __call__(self, base_paths: str, grammar_path: str) -> Tuple[str, str]: ...
 
 
 class Lark:
-    source: str
-    grammar_source: str
+    source_path: str
+    source_code: str
     options: LarkOptions
     lexer: Lexer
     terminals: List[TerminalDef]
@@ -68,8 +68,8 @@ class Lark:
         cache: Union[bool, str] = False,
         g_regex_flags: int = ...,
         use_bytes: bool = False,
-        import_sources: List[Union[str, Callable[[List[str], str], Tuple[str, str]]]] = ...,
-        source: Optional[str],
+        import_paths: List[Union[str, Callable[[Optional[str], str], Tuple[str, str]]]] = ...,
+        source_path: Optional[str],
     ):
         ...
 
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -32,14 +32,20 @@ class LarkOptions:
     cache: Union[bool, str]
     g_regex_flags: int
     use_bytes: bool
-    import_paths: List[Union[str, Callable[[Optional[str], str], Tuple[str, str]]]]
+    import_paths: List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]
     source_path: Optional[str]
 
+
+class PackageResource(object):
+    pkg_name: str
+    path: str
+    
+    def __init__(self, pkg_name: str, path: str):
 
 class FromPackageLoader:
     def __init__(self, pkg_name: str, search_paths: Tuple[str, ...] = ...): ...
     
-    def __call__(self, base_paths: str, grammar_path: str) -> Tuple[str, str]: ...
+    def __call__(self, base_path: Union[None, str, PackageResource], grammar_path: str) -> Tuple[PackageResource, str]: ...
 
 
 class Lark:
@@ -68,7 +74,7 @@ class Lark:
         cache: Union[bool, str] = False,
         g_regex_flags: int = ...,
         use_bytes: bool = False,
-        import_paths: List[Union[str, Callable[[Optional[str], str], Tuple[str, str]]]] = ...,
+        import_paths: List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]] = ...,
         source_path: Optional[str],
     ):
         ...
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 
 import sys, os, pickle, hashlib
 from io import open
-
+from warnings import warn
 
 from .utils import STRING_TYPE, Serialize, SerializeMemoizer, FS, isascii, logger
 from .load_grammar import load_grammar, FromPackageLoader
@@ -90,10 +90,10 @@ class LarkOptions(Serialize):
             Accept an input of type ``bytes`` instead of ``str`` (Python 3 only).
     edit_terminals
             A callback for editing the terminals before parse.
-    import_sources
+    import_paths
             A List of either paths or loader functions to specify from where grammars are imported 
-    source
-            Override the source of from where the grammar was loaded. Usefull for relative imports and unconventional grammar loading
+    source_path
+            Override the source of from where the grammar was loaded. Useful for relative imports and unconventional grammar loading
 
     **=== End Options ===**
     """
@@ -119,8 +119,8 @@ class LarkOptions(Serialize):
         'edit_terminals': None,
         'g_regex_flags': 0,
         'use_bytes': False,
-        'import_sources': [],
-        'source': None,
+        'import_paths': [],
+        'source_path': None,
     }
 
     def __init__(self, options_dict):
@@ -196,13 +196,13 @@ class Lark(Serialize):
             re_module = re
 
         # Some, but not all file-like objects have a 'name' attribute
-        if self.options.source is None:
+        if self.options.source_path is None:
             try:
-                self.source = grammar.name
+                self.source_path = grammar.name
             except AttributeError:
-                self.source = '<string>'
+                self.source_path = '<string>'
         else:
-            self.source = self.options.source
+            self.source_path = self.options.source_path
 
         # Drain file-like objects to get their contents
         try:
@@ -213,7 +213,7 @@ class Lark(Serialize):
             grammar = read()
 
         assert isinstance(grammar, STRING_TYPE)
-        self.grammar_source = grammar
+        self.source_code = grammar
         if self.options.use_bytes:
             if not isascii(grammar):
                 raise ValueError("Grammar must be ascii only, when use_bytes=True")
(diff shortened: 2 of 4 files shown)


MegaIng:
@sinoroc

This is how you could write a zipfile aware loader:
https://github.com/lark-parser/lark/pull/704/files#diff-ed9d66898a1188fc8f00f74cc941be36R1786

Does this help for your case?

erezsh:
You can rewrite `import_grammar` without the code duplication, by putting the `if callable()` inside the `with`.

sinoroc:
Thanks! I didn't expect anything so soon :) 

I will give it a try. I'm new to _Poetry_'s code base, so it will take me a while. Plus there is the additional difficulty of _lark_ being vendored in, so it will take me a bit to figure out how to test this. I will let you know...

MegaIng:
Ok, after thinking a bit more about this, I decided that this should actually still be a draft:

- I am pretty sure relative import currently don't work
- I would like to add a simple interface so that it is easy to have a custom loader that uses grammar from a specific package.

erezsh:
Don't we have tests for relative imports? Or do you mean it doesn't work for callable sources?

MegaIng:
Yes, for the normal imports it works. But I am pretty sure it doesn't quite work for the current 'model' implementation. (And yes, we have tests. But I honestly don't have an overview on how they are organized. Maybe you can write a few tests that test all possible edge-cases with relative imports for custom sources? I don't think we need tests for renaming though)

MegaIng:
Ok, with this commit,  I think everything works. I also tested it from inside a zip files and everything works (Both `common.lark` loading and custom package imports) Maybe a few more docs would be nice, but where?

MegaIng:
@sinoroc Also note that for your usecase it should now be as simple as calling `Lark.open_from_package(__name__, 'grammar_name.lark', ('grammar_folder', ))`

lark/load_grammar.py
```diff
@@ -648,17 +649,67 @@ def compile(self, start):
...
+        self.search_paths = search_paths
+        
+    def __repr__(self):
+        return "%s(%r, %r)" % (type(self).__name__, self.pkg_name, self.search_paths)
+
+    def __call__(self, base_paths, grammar_path):
+        if len(base_paths) == 0:
+            to_try = self.search_paths
+        else:
+            assert len(base_paths) == 1
```
erezsh, line 673: Why is `len(base_paths)` only 0 or 1? Maybe it should be `Optional[str]` instead then?

lark/load_grammar.py
```diff
@@ -648,17 +649,67 @@ def compile(self, start):
...
+            if not base_paths[0].startswith('<%s:' % (self.pkg_name,)):
```
erezsh, line 674: What is this?

lark/load_grammar.py
```diff
@@ -648,17 +649,67 @@ def compile(self, start):
...
-                with open(joined_path, encoding='utf8') as f:
-                    text = f.read()
-                grammar = load_grammar(text, joined_path, re_)
+                if callable(source):
+                    joined_path, text = source(base_paths, grammar_path)
+                else:
+                    joined_path = os.path.join(source, grammar_path)
+                    with open(joined_path, encoding='utf8') as f:
+                        text = f.read()
+            if text is not None:
```
erezsh, line 709: Why would `text` be None?

lark/lark.py
```diff
@@ -401,6 +410,28 @@ def open(cls, grammar_filename, rel_to=None, **options):
             grammar_filename = os.path.join(basepath, grammar_filename)
         with open(grammar_filename, encoding='utf8') as f:
             return cls(f, **options)
+    
+    @classmethod
+    def open_from_package(cls, package, grammar_path, search_paths=("",), **options):
+        """Create an instance of Lark with the grammar loaded from within the package `package`.
+        This allows grammar loading from zipapps.
+        
+        Will also create a `FromPackageLoader` instance and add it to the `import_sources` to simplify importing
```
erezsh, line 419: Why is this useful for the user to know?

lark/lark.py
```diff
@@ -115,6 +119,8 @@ class LarkOptions(Serialize):
         'edit_terminals': None,
         'g_regex_flags': 0,
         'use_bytes': False,
+        'import_sources': [],
```
erezsh, line 122: The common name for this is `path`, both in windows, linux, and Python with `sys.path`.

So maybe `import_paths` makes more sense?

lark/lark.py
```diff
@@ -115,6 +119,8 @@ class LarkOptions(Serialize):
...
+        'source': None,
```
erezsh, line 123: `source` is a little vague in this context. Maybe call it `source_name` or `source_path`.

erezsh (changes requested):
Left you some comments.

lark/lark.py
MegaIng: I thought that it would be useful to prevent users from manually specific the `import_sources`. `FromPackageLoader` is after all a public class.

lark/lark.py
MegaIng: Yes, true

lark/lark.py
MegaIng: I though so too. But those paths are almost always just file paths, not custom loaders. If you want, we can rename it, but I specifically decided against it.

lark/load_grammar.py
MegaIng: This is to recognize relative imports from within a package. When a grammar is loaded by `FromPackageLoader`, it gets a path in the form `<{pkg_name}:/{full_path}>` Where `full_path` is the argument that was passed to `pkgutil.get_data`. Here we unparse it to get the relative path that was specified in importing (allowing relative imports). I will write a comment for this.

lark/load_grammar.py
MegaIng: This is the design you choose in the implementation. I have nothing against changing this, but I imagine you at some point in the future might want to add additional elements to `base_paths`. If that is the case, this `assert` will trigger and we have to change something here. I could future proof this now if you want. (e.g. allow multiple `base_paths` entries.

lark/lark.py
erezsh: Hmm maybe it's better to say something like 

    Imports in the grammar will use the `package` and `search_paths` provided, through `FromPackageLoader`.

But I'll leave it to your judgement.

lark/lark.py
erezsh: Even so, `import_paths` seems more adequate, no?

lark/load_grammar.py
erezsh: When I used `base_paths`, I did not differentiate between base_paths and search_paths, and I did not consider relative imports vs absolute imports.

Now that we make the distinction, there shouldn't be more than one base_path (or so it seems to me). And search_paths has now taken its place for a list of paths. So I think it should be turned into an optional string.

lark/lark.py
MegaIng: Would you be ok with also renaming `Lark.source`? It is as vague. (Also I would rename `Lark.grammar_source` to `Lark.source_code`)

lark/load_grammar.py
erezsh: Why are we accepting this repr-like string? Just so we can support `__name__`?

lark/lark.py
erezsh: In principle I agree, but it would break compatibility. You can open an issue for this for v1.0

Fwiw, at least when you read the existing `source` property, you can understand what it is from its value.

lark/load_grammar.py
MegaIng: No, because there might not be a meaningful way of expressing the location of the grammar file as a normal file path the `open` accepts. This is similar to `<string>`, just that it also carries information for imports inside of it instead of implied/inferred at a latter stage. This is a format I designed to just be used internally by the `FromPackageLoader` and also be readable by the end user. Do you have an alternative suggestion?

lark/lark.py
MegaIng: I would of course add backwards compatibility via properties.

lark/load_grammar.py
erezsh: Well, if you already know the package name.. Why not just accept a regular path within the package?

MegaIng: Because that could cause problems when there are multiple packages from which we are loading. And it is not like these paths would actually be useful for anything else.

lark/load_grammar.py
MegaIng: Because the `with suppress()` statement can fail and then the previous value of text will stay. I assign `None` to it before hand. Alternatively, I can rework it into a `try except else` statement if you want

lark/lark.py
erezsh: Okay, I suppose there's no harm

lark/load_grammar.py
erezsh: Maybe I missing something, but this all seems a little messier than it has to be.

But if you insist on this hack, parse it with a regexp. It shouldn't take more than two lines.

And document it well.

lark/load_grammar.py
erezsh: Yes, that's why I loaded the text inside the suppress. But if you think it's better outside (I understand, imports and such), it's better to use `try.. except`

MegaIng: Read the next comment.

erezsh: I already said I understand. What did that comment add to *my* comment?

lark/load_grammar.py
MegaIng: If you have an alternative suggestion that doesn't run into problems when there are multiple files with the same name in different places, I am open to ideas.

lark/load_grammar.py
MegaIng: Sorry, I misread your comment.

lark/load_grammar.py
erezsh: So, I'm trying to think of the actual use cases. Help me understand.

The first use case, is when trying to load the main grammar from a package. But there is a method for it, and it's only one grammar, one package. No difficulty.

The second use case, is when trying to import a grammar from within a grammar. In that case, we currently don't allow to specify the package anyway. We only allow a dotted path.

I can't think of other use cases.

So, in what realistic situation will this feature be used?

lark/lark.py
MegaIng: Python uses `sys.path` and `sys.path_hooks` for this. I don't want to split it into multiple arguments.

Ok, I will rename it `import_paths`.

lark/load_grammar.py
MegaIng: Maybe one can create a package with grammar library with lots of different grammar files. Then you can add that package to your own module where you also have a somewhat complex grammar split over multiple files. (Or a plugin system). If you try to use relative imports, we have to prevent the grammars from one package to find the grammars in the other package.

Under the old file-based system, this wasn't a problem since we always specified an absolute file path that clearly defines which grammar file we are talking about.

But now we can't easily do that. If we just use `full_path` there might be ambiguity of both packages use the `grammars` folder. If we try to use an absolute path, this will break when we are inside of a zip file. So we somehow have to encode both the package and the path inside of the package, unless we don't want to support those usecases. And I don't see any real harm done by this.

erezsh: You are describing a situation that no one currently has.

Also, I still don't understand who will pass the `<package:path>` string, and where it will pass.

If both packages have a `grammars` folder in the search paths, that is still not a collision. Only if they have the same `whatever.lark` file it will collide, no? And then it will choose whatever is first on the list.

The harm is code bloat. Death by a million cuts. Every little feature we add, is something we will have to support. And one line after another end up being hundreds of lines, that force us to do things in a certain way, because we overengineered when we didn't have to.

Again, maybe if you can give a specific use case that it will solve, it will help me understand. If you can't explain it, show me the code from the user's perspective.

MegaIng: > Also, I still don't understand who will pass the <package:path> string, and where it will pass.
This is the end of the function. I added a few more comments. So it is just the `FromPackageLoader` talking to itself.

erezsh: I see. If this is an internal structure, and not something that is user-facing, then please let's use a dedicated class for it.

MegaIng: Well, it is also the name of the grammar to be used in error reporting (the source argument to `GrammarLoader.load_grammar`) The benefit of just using str is that we don't need to make any changes in `load_grammar`.

erezsh: That's what `__str__` is for. And if we need to add a `str()` call somewhere, it will be well worth it.

MegaIng: Well, I pushed another commit. I would call this more code bloat than before, but if you prefer this, ok.

lark-stubs/lark.pyi
```diff
@@ -32,14 +32,20 @@ class LarkOptions:
...
+    pkg_name: str
+    path: str
+    
+    def __init__(self, pkg_name: str, path: str):
+
 class FromPackageLoader:
     def __init__(self, pkg_name: str, search_paths: Tuple[str, ...] = ...): ...
     
-    def __call__(self, base_paths: str, grammar_path: str) -> Tuple[str, str]: ...
+    def __call__(self, base_paths: Union[None, str, PackageResource], grammar_path: str) -> Tuple[PackageResource, str]: ...
```
erezsh, line 48: This should be `base_path`

lark-stubs/lark.pyi
```diff
@@ -32,11 +32,25 @@ class LarkOptions:
...
+    def __init__(self, pkg_name: str, search_paths: Tuple[str, ...] = ...): ...
+    
+    def __call__(self, base_path: Union[None, str, PackageResource], grammar_path: str) -> Tuple[PackageResource, str]: ...
 
 
 class Lark:
-    source: str
-    grammar_source: str
+    source_path: str
+    source_code: str
```
erezsh, line 53: Sorry for not catching this earlier, but this should be `source_text`. Not everyone uses lark just to parse code.

lark/lark.py
```diff
@@ -450,6 +479,15 @@ def parse(self, text, start=None, on_error=None):
                         # Prevent infinite loop
                         raise e2
                     e = e2
+    
+    @property
+    def source(self):
+        warn("Lark.source attribute has been renamed to Lark.source_path", DeprecationWarning)
+        return self.source_path
+    
+    @source.setter
```
erezsh, line 488: This is read-only. A setter isn't necessary.

lark-stubs/lark.pyi
MegaIng: Well, it is the source code for the grammar. Maybe `source_grammar`?

erezsh: Hmm then my comment is irrelevant. But yes, maybe that's a better name anyway.

erezsh (approved):

erezsh:
`grammar_source` is removed, but needs property+warning for backward compatibility.

I think that's the last hold out.

erezsh:
Thanks for the PR!

lark/load_grammar.py
erezsh: To be fair, this whole zipapp feature was a code bloat :)

PR #707, opened 2020-10-01
Fix for maybe_placeholders when keep_all_tokens=True

### LINKED ISSUE #706: maybe_placeholders and anonymous terminal omits the None
Doc says [here](https://lark-parser.readthedocs.io/en/latest/grammar.html#rules) that `maybe_placeholders=True` makes ` rule : some1 [item] some2` to produce a `None` if item is not found, but anonymous token seems to not produce `None`.

Minimal reproduction: 
~~~python
from lark import Lark

grammar = r"""
UINT : /\d+/

math_atom : UINT 

LESS : "<"

math_script : math_atom ("_" ["<"] math_atom)*
"""

parser = Lark(
    grammar= grammar,
    parser="lalr",
    keep_all_tokens=True,
    maybe_placeholders=True,
    start="math_script"
     )

test_file= "1_1"
tree = parser.parse(test_file)
print(tree.pretty())
~~~

Produces: 
~~~
math_script
  math_atom     1
  _
  math_atom     1
~~~

If you replace `math_atom` to `UINT` it works. 
if you replace `["<"]` with `[LESS]` it works. 

~~~
math_script
  math_atom     1
  _
  None
  math_atom     1
~~~

I have both a global lark installed and a fresh environment with just lark

global install `Version: 0.9.0`
environment install  `Version: 0.10.0`

MegaIng:
Fixes #706.

Includes test

slight change in the design: now the GrammarLoader is reused when importing, which shouldn't make a difference.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -650,7 +650,7 @@ class Grammar:
 
 
 _imported_grammars = {}
-def import_grammar(grammar_path, re_, base_paths=[]):
+def import_grammar(grammar_path, loader, base_paths=[]):
     if grammar_path not in _imported_grammars:
         import_paths = base_paths + IMPORT_PATHS
         for import_path in import_paths:
@@ -658,7 +658,7 @@ def import_grammar(grammar_path, re_, base_paths=[]):
                 joined_path = os.path.join(import_path, grammar_path)
                 with open(joined_path, encoding='utf8') as f:
                     text = f.read()
-                grammar = load_grammar(text, joined_path, re_)
+                grammar = loader.load_grammar(text, joined_path)
                 _imported_grammars[grammar_path] = grammar
                 break
         else:
@@ -803,7 +803,7 @@ class GrammarLoader:
         ('%ignore expects a value', ['%ignore %import\n']),
     ]
 
-    def __init__(self, re_module):
+    def __init__(self, re_module, always_keep_all_tokens):
         terminals = [TerminalDef(name, PatternRE(value)) for name, value in TERMINALS.items()]
 
         rules = [options_from_rule(name, None, x) for name, x in  RULES.items()]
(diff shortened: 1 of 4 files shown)


lark/load_grammar.py
```diff
@@ -946,7 +947,10 @@ def load_grammar(self, grammar_text, grammar_name='<?>'):
         rules = rule_defs
 
         rule_names = {}
-        for name, params, _x, _o in rules:
+        for name, params, _x, option in rules:
+            if self.always_keep_all_tokens: # We need to do this somewhere. Might as well prevent an additional loop
```
erezsh, line 951: This comment isn't helpful.

erezsh:
Everything looks alright. I'd just remove that comment, or replace it with something that new readers might find useful.


Does it conflict with #710? I don't see how.

MegaIng:
It changes the placement of the usage of `keep_all_tokens`, meaning the `_load` does not have a chance to correct the rules I think. Previously it was only used in building the callbacks which is done in `_load`, but now it is done in `load_grammar`. I think the best solution for this is extending #710/creating a new PR that allows options that need to be specified early like `keep_all_tokens` (after this PR (#707)) or `maybe_placeholders` to be set from the command line.

erezsh:
Why would `_load` need to correct the rules? Aren't the `RuleOptions` saved and loaded together with the rules?

MegaIng:
Yes, but we have a difference between the options before and after loading. That is the core problem here. When the standalone parser is generated, `keep_all_tokens` is not set, therefore `load_grammar` does not set the attributes correctly and the wrong attributes are saved. When afterwards `keep_all_tokens` is set, this will not be reflected in the attributes of the `RuleOptions`. (And just modifying the RuleOptions afterwards once again doesn't help for `maybe_placeholders` and is just a hack)

erezsh:
I see. So that leaves us with two options:

1. Require `keep_all_tokens` to be set when creating the standalone, and forbid changing it afterwards

2. Revert to the original code where `keep_all_tokens` only affected the tree builder

Unless I missed something?

MegaIng:
> Revert to the original code where keep_all_tokens only affected the tree builder

Yes, but then we also need to redesign `maybe_placeholders`. While possible, this is probably something you should take on as I don't know how it really works.

erezsh:
Yes, I think 1 is the best option. It's also not such a popular switch in the first place.

MegaIng:
The cosmic rays decided that the tests shall not run.

erezsh:
The universe must have its reasons.

PR #709, opened 2020-10-03
Guard Lark.__doc__ assignment for tools.standalone

rec:
This change is necessary for standalones to work, otherwise you get errors like:

    kson/grammar/kson.py:2285: in Lark
        __doc__ += "\n\n" + LarkOptions.OPTIONS_DOC
    E   TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -310,7 +310,8 @@ class Lark(Serialize):
             with FS.open(cache_fn, 'wb') as f:
                 self.save(f)
 
-    __doc__ += "\n\n" + LarkOptions.OPTIONS_DOC
+    if __doc__:
+        __doc__ += "\n\n" + LarkOptions.OPTIONS_DOC
 
     __serialize_fields__ = 'parser', 'rules', 'options'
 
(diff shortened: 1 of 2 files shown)


MegaIng (approved):

MegaIng:
I could have sworn this guard was already there, but apparently it was only in `LarkOptions`. Also, if this is that broken, how did the tests pass? (Or did you run it with the `-O` flag?)

rec:
Something about the standaloning process strips away the `__doc__` comments...?

MegaIng:
Ok. We have in theory tests for this. But for some reason `exec` messes with the `__doc__` variable, so that despite the doc string being stripped away, it is still defined. But if you run it in an external file, it obviously doesn't work.

MegaIng:
Ok, can you update this PR with the following code in `tests/test_tools.py` in `TestStandalone._create_standalone`, replacing the 
```
context = {]
```
line with
```
context = {'__doc__': None}
```
?

This simulates the behavior better.

rec:
onnit!

rec:
All right, I force-pushed a new commit which passes the tests locally!

PR #710, opened 2020-10-03
Allowing the standalone parser to take more options

### LINKED ISSUE #565: Better way to set options of the stand-alone parser
**Suggestion**

I cannot set the options of `Lark_StandAlone`, but I find a weird way to set it.

```python
from json_parser import Lark_StandAlone, DATA

DATA['options']['maybe_placeholders'] = True
parser = Lark_StandAlone()
```

it looks better to do it by code below.

```python
from json_parser import Lark_StandAlone

parser = Lark_StandAlone(maybe_placeholders=True)
```

MegaIng:
Closes #565.

A subset of the options is specified that should be safe to pass after the parser has been created. I think that all of the once specified are safe, but @erezsh you should double-check. I also think that the restcan't be passed, but I am less sure about that. Either way, I didn't test everything.

(Note: conflicts with #707 I think)

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -165,7 +165,6 @@ class LarkOptions(Serialize):
 
 _LOAD_ALLOWED_OPTIONS = {'postlex', 'transformer', 'use_bytes', 'debug', 'g_regex_flags', 'regex', 'propagate_positions', 'keep_all_tokens',
                          'tree_class'}
-_LOAD_BLOCKED_OPTIONS = set(LarkOptions._defaults.keys()) - _LOAD_ALLOWED_OPTIONS
 
 
 class Lark(Serialize):
@@ -235,8 +234,10 @@ class Lark(Serialize):
 
             if FS.exists(cache_fn):
                 logger.debug('Loading grammar from cache: %s', cache_fn)
+                for name in (set(LarkOptions._defaults) - _LOAD_ALLOWED_OPTIONS):
+                    options.pop(name, None)
                 with FS.open(cache_fn, 'rb') as f:
-                    self._load(f, self.options)
+                    self._load(f, **options)
                 return
 
         if self.options.lexer == 'auto':
@@ -352,7 +353,7 @@ class Lark(Serialize):
         inst = cls.__new__(cls)
         return inst._load(f)
 
-    def _load(self, f, temp_options=None, **kwargs):
+    def _load(self, f, **kwargs):
         if isinstance(f, dict):
             d = f
         else:
@@ -363,15 +364,11 @@ class Lark(Serialize):
         assert memo
         memo = SerializeMemoizer.deserialize(memo, {'Rule': Rule, 'TerminalDef': TerminalDef}, {})
         options = dict(data['options'])
-        if _LOAD_BLOCKED_OPTIONS.intersection(kwargs.keys()):
+        if (set(kwargs) - _LOAD_ALLOWED_OPTIONS) & set(LarkOptions._defaults):
             raise ValueError("Some options are not allowed when loading a Parser: {}"
-                             .format(_LOAD_BLOCKED_OPTIONS.intersection(kwargs.keys())))
+                             .format(set(kwargs) - _LOAD_ALLOWED_OPTIONS))
         options.update(kwargs)
-        if temp_options is not None:
-            for o in _LOAD_ALLOWED_OPTIONS:
-                options[o] = temp_options.options[o]
         self.options = LarkOptions.deserialize(options, memo)
-        re_module = regex if self.options.regex else re
         self.rules = [Rule.deserialize(r, memo) for r in data['rules']]
         self.source = '<deserialized>'
         self._prepare_callbacks()
@@ -379,7 +376,6 @@ class Lark(Serialize):
             data['parser'],
             memo,
             self._callbacks,
-            re_module,
             self.options, # Not all, but multiple attributes are used
         )
         self.terminals = self.parser.lexer_conf.tokens
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -236,7 +236,7 @@ class Lark(Serialize):
             if FS.exists(cache_fn):
                 logger.debug('Loading grammar from cache: %s', cache_fn)
                 with FS.open(cache_fn, 'rb') as f:
-                    self._load(f, self.options.transformer, self.options.postlex)
+                    self._load(f, self.options)
                 return
 
         if self.options.lexer == 'auto':
@@ -352,7 +352,7 @@ class Lark(Serialize):
         inst = cls.__new__(cls)
         return inst._load(f)
 
-    def _load(self, f, **kwargs):
+    def _load(self, f, temp_options=None, **kwargs):
         if isinstance(f, dict):
             d = f
         else:
@@ -367,6 +367,9 @@ class Lark(Serialize):
             raise ValueError("Some options are not allowed when loading a Parser: {}"
                              .format(_LOAD_BLOCKED_OPTIONS.intersection(kwargs.keys())))
         options.update(kwargs)
+        if temp_options is not None:
+            for o in _LOAD_ALLOWED_OPTIONS:
+                options[o] = temp_options.options[o]
         self.options = LarkOptions.deserialize(options, memo)
         re_module = regex if self.options.regex else re
         self.rules = [Rule.deserialize(r, memo) for r in data['rules']]
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -163,6 +163,11 @@ class LarkOptions(Serialize):
         return cls(data)
 
 
+_LOAD_ALLOWED_OPTIONS = {'postlex', 'transformer', 'use_bytes', 'debug', 'g_regex_flags', 'regex', 'propagate_positions', 'keep_all_tokens',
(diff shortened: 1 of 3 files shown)


lark/lark.py
```diff
@@ -358,10 +363,10 @@ def _load(self, f, transformer=None, postlex=None):
         assert memo
         memo = SerializeMemoizer.deserialize(memo, {'Rule': Rule, 'TerminalDef': TerminalDef}, {})
         options = dict(data['options'])
-        if transformer is not None:
-            options['transformer'] = transformer
-        if postlex is not None:
-            options['postlex'] = postlex
+        if _LOAD_BLOCKED_OPTIONS.intersection(kwargs.keys()):
```
rec, line 366: You can remove `.keys()` in the previous line.  A `dict` when used as an iterable is its keys!

lark/lark.py
```diff
@@ -371,19 +376,17 @@ def _load(self, f, transformer=None, postlex=None):
             data['parser'],
             memo,
             self._callbacks,
-            self.options.postlex,
-            self.options.transformer,
             re_module,
-            self.options.debug
+            self.options, # Not all, but multiple attributes are used
```
rec, line 383: This comment raises more questions than it answers.  :-D

I'd move it from here entirely and put it up to right where you define `_LOAD_ALLOWED_OPTIONS`, with more explanation.  I tried to write that longer explanation but couldn't, which means it needs it.

lark/lark.py
```diff
@@ -163,6 +163,11 @@ def deserialize(cls, data, memo):
         return cls(data)
 
 
+_LOAD_ALLOWED_OPTIONS = {'postlex', 'transformer', 'use_bytes', 'debug', 'g_regex_flags', 'regex', 'propagate_positions', 'keep_all_tokens',
+                         'tree_class'}
+_LOAD_BLOCKED_OPTIONS = set(LarkOptions._defaults.keys()) - _LOAD_ALLOWED_OPTIONS
```
rec, line 168: `.keys()` is unnecessary here too.

rec (commented):
Really nice idea - `options` now captures all the setup information and you use `_LOAD_ALLOWED_OPTIONS` to decide which parts are actually used.

I am not sure I see where there's checking for completely wrong options that aren't blocked, thought?

lark/lark.py
MegaIng: I made the conscious decisions to use `.keys()`: explicit is better than implicit. But maybe my judgement was wrong on that. Lets hear @erezsh as a third opinion.

lark/lark.py
MegaIng: This has nothing to do with `_LOAD_ALLOWED_OPTIONS`, but with `parser_frontend.WithLexer.deserialize` which takes some options and assigns them to attributes. (Including as example `use_bytes`) Previously they were all passed as arguments, leading to like five extra arguments. But once again, maybe we should do it more explicit here.

MegaIng:
>I am not sure I see where there's checking for completely wrong options that aren't blocked, thought?

This is done like before inside the `LarkOptions.__init__` method. It is a very split up system.

lark/lark.py
MegaIng: On third through, my judgement was probably wrong. Using it is less pythonic.

lark/parser_frontends.py
```diff
@@ -82,16 +82,18 @@ def __init__(self, lexer_conf, parser_conf, options=None):
...
 
-        inst.postlex = postlex
-        inst.parser = LALR_Parser.deserialize(inst.parser, memo, callbacks, debug)
+        inst.postlex = options.postlex
+        inst.parser = LALR_Parser.deserialize(inst.parser, memo, callbacks, options.debug)
 
         terminals = [item for item in memo.values() if isinstance(item, TerminalDef)]
-        inst.lexer_conf.callbacks = _get_lexer_callbacks(transformer, terminals)
+        inst.lexer_conf.callbacks = _get_lexer_callbacks(options.transformer, terminals)
         inst.lexer_conf.re_module = re_module
```
erezsh, line 93: Why not fold `re_module` as well?

```python
inst.lexer_conf.re_module = regex if options.regex else re
```

erezsh, line 93: Why not fold `re_module` as well?

```python
inst.lexer_conf.re_module = regex if options.regex else re
```

erezsh, line 93: Why not fold `re_module` as well?

```python
inst.lexer_conf.re_module = regex if options.regex else re
```

lark/lark.py
erezsh: I don't have a strong opinion about that. But personally I'd prefer

```python
_LOAD_BLOCKED_OPTIONS & set(...)
```

Over intersection.

erezsh: Also, why do you even need `_LOAD_BLOCKED_OPTIONS`? Why not just

```python
if set(kwargs) - _LOAD_ALLOWED_OPTIONS:
   # raise error
```

erezsh: And why do we need `temp_options` if we have `kwargs`?

erezsh:
I like the idea behind this. See my individual comments.

lark/lark.py
MegaIng: temp_options to deal with cache. If we used cache, the illegal option of cache will get passed through, but I will take a look.

> Also, why do you even need _LOAD_BLOCKED_OPTIONS

I am not that proficient with set arithmetic, yes you are correct both suggestions are better.

MegaIng: Aha. One thing: if we `set(kwargs) - _LOAD_ALLOWED_OPTIONS` completely wrong options will get the same error message as for options that normally can be set. That is what I was avoiding with `_LOAD_BLOCKED_OPTIONS`. There are a lot of ways on how to solve this, this just seemed easiest.

lark/parser_frontends.py
MegaIng: It means that we have to import `re` and `regex` into `parser_frontend.py`.

lark/lark.py
erezsh: Well, you can change the error message based on if it's `in LarkOptions._defaults`

I agree it's not a huge difference, but it feels better to me.

lark/parser_frontends.py
erezsh: I don't know if it makes sense to pass a module as an argument only so we won't have to import it.

MegaIng: That is true

lark/lark.py
MegaIng: Removing temp_options on the other hand is quite the pain. We need to filter out the arguments that can be passed to `Lark`, but not to `Load._lark`, but which still make sense when using the cache, e.g. the entire `_LOAD_BLOCKED_OPTIONS`. I will push a commit, but I am not sure how nice this is.

lark/lark.py
MegaIng: Yup, thanks

lark/lark.py
MegaIng: @erezsh Went in the other direction, so we keep this.

lark/lark.py
erezsh: I don't understand why it's a pain. The standalone parser isn't affected by `cache` anyway. (it can be assumed to be always on regardless)

MegaIng: Because the same method (`_load`) is used. Changes to the system how the standalone parser operators also have to be reflected in the cache handling in some way. This is one of them. View the latest commit. But it is less of a pain than I thought.

lark/lark.py
```diff
@@ -230,8 +234,10 @@ def __init__(self, grammar, **options):
 
             if FS.exists(cache_fn):
                 logger.debug('Loading grammar from cache: %s', cache_fn)
+                for name in (set(LarkOptions._defaults) - _LOAD_ALLOWED_OPTIONS):
+                    options.pop(name, None)
```
erezsh, line 238: Wait, what? Why? We should throw an error

erezsh: "This option is incompatible with `cache`"

MegaIng: It isn't incompatible. Only `_load` thinks they are incompatible. The options that are thrown away are stuff like `parser` and `start` which are compatible if used with cache, but not when used with Standalone.

erezsh: Ah, okay. I see. I guess anything that is really incompatible would be thrown away earlier, right?

MegaIng: I hope? Otherwise this is not a bug in my implementation, but a carry over from before. (In theory the cache is only used if the options are the same, and if not the options are validated by generating a full parser)

erezsh: Okay, that makes sense.

MegaIng:
I think this is complete now? Unless I should add the standalone generator taking options also into this PR.

erezsh:
I think so. I'll take a closer look tomorrow, and if everything fits I'll merge it.

linw1995:
Hello everyone, I am sorry to bother you guys. But why the option `maybe_placeholders=True` is disallowed?

erezsh:
@linw1995 It is not disallowed, but it cannot be changed after the parser is already generated.

To set it, you must supply it when you initiate the Lark instance.

PR #711, opened 2020-10-05
Allows the standalone parser to take more options at the command line

MegaIng:
This adds most of the options that can be passed to a normal Lark instantiation to the standalone generator and the serializer.

Based on #710 and #707.

Note: breaks backwards compatibility by deprecating `python -m lark.tools.standalone grammar-file start` in favor of `python -m lark.tools.standalone -s start grammar-file`

Also adds an overview where new options to `Lark` need to (potentially) be added.


erezsh:
Breaking compatibility isn't great!

Unless you want it to wait for v1.0

MegaIng:
Well, it does have backwards compatibility. It isn't breaking. The only problem being that I use the `warn` function that outputs to stderr that the syntax is no longer allowed. This _should_ not mess with the normal usecase since only stdout _should_ be redirected. Therefor it _should_ not be breaking. If stderr also gets put in the file, python will simply not read it.

erezsh:
Ah, ok. Well, you can output to stderr starting with `#` just to be safe.

Also, can you squash some of those commits?

MegaIng:
Yeah, but then I can't use `warn`.

Also, squashing would be easier if #707 is merged first.

MegaIng:
Btw, is there anywhere a list which breaking changes will be made with 1.0?

erezsh:
#707 Merged.

I keep a list of v1.0 changes locally, but I'm not sure I'll do all of them:

- change "standard" lexer into "simple" or "basic"
- new grammar syntax?
- lexer_conf.tokens -> terminals
- unify position attribute names
- propagate positions on every node with a meta attribute?
    - tokens with meta attribute instead of direct position attributes?

MegaIng:
Ok, now I have managed to squash correctly. I had lots of trouble because my master wasn't up to date.

lark/tools/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+    'keep_all_tokens',
+    'regex',
+    'propagate_positions',
+    'maybe_placeholders',
+    'use_bytes'
+]
+
+options = ['start', 'lexer']
+
+base_argparser.add_argument('-v', '--verbose', action='count', default=0)
```
erezsh, line 26: More like `lalr_argparser`, as this combination wouldn't apply in any other situation.

erezsh: It should be documented that `verbose` is between 0 and 3

erezsh:
Nice PR! I added a few comments.

lark/tools/__init__.py
MegaIng: Ok, done

erezsh:
Thanks! 

I also added a little bonus.

MegaIng:
Suggestion: drop support for python2...

erezsh:
python2 is still widely used. But feel free to open an issue.

PR #715, opened 2020-10-07
Interegular integration, allowing checks for intersections between regexes.

### LINKED ISSUE #76: Detect collision in regular expressions when creating a lexer

MegaIng:
Fixes #76.

This is a first attempt.

You can install `interegular` via pip: `pip install interegular`. If you want, you can also take a look at the [interegular source code](https://github.com/MegaIng/interegular).


erezsh:
A few notes:

1. Basically, this is the real functional code (which resides in interregular):
```
        for a, b in combinations(keys, 2):
            if not self.isdisjoint(a, b):
                yield a, b
```

I see you already cache the FSMs, so all that's required to prevent duplicate work is to call `memoized_disjoint` instead.

2. Why does lark have to call `mark` and `is_marked`? Seems unnecessary. If you want to avoid duplicate warnings, just keep a set.

3. Why compare regexps (expensive) and only afterwards check `a.priority == b.priority`? This should be the first test, otherwise comparing them is pointless.

4. `skip_validation` means you never use the comparator. So why even create it in the first place?

erezsh:
P.S. re point 3, you can do something like `classify(regexps, lambda r: r.priority)` to get all the subgroups that should be tested together.

MegaIng:
Well, this is now a messed up git history.

PR #725, opened 2020-10-18
Resumable inteface, backwards compatible and ready to merge

erezsh:
Pretty much the same as `resumable`, but fixed custom lexer to be backwards compatible.

### DIFF
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -1,6 +1,6 @@
 from .utils import get_regexp_width, Serialize
 from .parsers.grammar_analysis import GrammarAnalyzer
-from .lexer import TraditionalLexer, ContextualLexer, Lexer, Token, TerminalDef
+from .lexer import LexerThread, TraditionalLexer, ContextualLexer, Lexer, Token, TerminalDef
 from .parsers import earley, xearley, cyk
 from .parsers.lalr_parser import LALR_Parser
 from .grammar import Rule
@@ -23,12 +23,18 @@ def get_frontend(parser, lexer):
         elif lexer == 'contextual':
             return LALR_ContextualLexer
         elif issubclass(lexer, Lexer):
+            class CustomLexerWrapper(Lexer):
+                def __init__(self, lexer_conf):
+                    self.lexer = lexer(lexer_conf)
+                def lex(self, lexer_state, parser_state):
+                    return self.lexer.lex(lexer_state.text)
+
             class LALR_CustomLexerWrapper(LALR_CustomLexer):
                 def __init__(self, lexer_conf, parser_conf, options=None):
                     super(LALR_CustomLexerWrapper, self).__init__(
                         lexer, lexer_conf, parser_conf, options=options)
                 def init_lexer(self):
-                    self.lexer = lexer(self.lexer_conf)
+                    self.lexer = CustomLexerWrapper(self.lexer_conf)
 
             return LALR_CustomLexerWrapper
         else:
@@ -54,7 +60,7 @@ def get_frontend(parser, lexer):
 
 
 class _ParserFrontend(Serialize):
-    def _parse(self, input, start, *args):
+    def _parse(self, start, input, *args):
         if start is None:
             start = self.start
             if len(start) > 1:
@@ -70,6 +76,18 @@ def _get_lexer_callbacks(transformer, terminals):
         if callback is not None:
             result[terminal.name] = callback
     return result
+
+class PostLexConnector:
+    def __init__(self, lexer, postlexer):
+        self.lexer = lexer
+        self.postlexer = postlexer
+
+    def make_lexer_state(self, text):
+        return self.lexer.make_lexer_state(text)
+
+    def lex(self, lexer_state, parser_state):
+        i = self.lexer.lex(lexer_state, parser_state)
+        return self.postlexer.process(i)
 
 
 class WithLexer(_ParserFrontend):
@@ -106,13 +124,14 @@ class WithLexer(_ParserFrontend):
     def _serialize(self, data, memo):
         data['parser'] = data['parser'].serialize(memo)
 
-    def lex(self, *args):
-        stream = self.lexer.lex(*args)
-        return self.postlex.process(stream) if self.postlex else stream
+    def make_lexer(self, text):
+        lexer = self.lexer
+        if self.postlex:
+            lexer = PostLexConnector(self.lexer, self.postlex)
+        return LexerThread(lexer, text)
 
     def parse(self, text, start=None):
-        token_stream = self.lex(text)
-        return self._parse(token_stream, start)
+        return self._parse(start, self.make_lexer(text))
 
     def init_traditional_lexer(self):
         self.lexer = TraditionalLexer(self.lexer_conf)
@@ -138,14 +157,6 @@ class LALR_ContextualLexer(LALR_WithLexer):
         always_accept = self.postlex.always_accept if self.postlex else ()
         self.lexer = ContextualLexer(self.lexer_conf, states, always_accept=always_accept)
 
-
-    def parse(self, text, start=None):
-        parser_state = [None]
-        def set_parser_state(s):
-            parser_state[0] = s
-
-        token_stream = self.lex(text, lambda: parser_state[0])
-        return self._parse(token_stream, start, set_parser_state)
 ###}
 
 class LALR_CustomLexer(LALR_WithLexer):
@@ -156,15 +167,6 @@ class LALR_CustomLexer(LALR_WithLexer):
         WithLexer.__init__(self, lexer_conf, parser_conf, options)
 
 
-def tokenize_text(text):
-    line = 1
-    col_start_pos = 0
-    for i, ch in enumerate(text):
-        if '\n' in ch:
-            line += ch.count('\n')
-            col_start_pos = i + ch.rindex('\n')
-        yield Token('CHAR', ch, line=line, column=i - col_start_pos)
-
 class Earley(WithLexer):
     def __init__(self, lexer_conf, parser_conf, options=None):
         WithLexer.__init__(self, lexer_conf, parser_conf, options)
@@ -174,6 +176,9 @@ class Earley(WithLexer):
         debug = options.debug if options else False
         tree_class = options.tree_class or Tree if options.ambiguity != 'forest' else None
         self.parser = earley.Parser(parser_conf, self.match, resolve_ambiguity=resolve_ambiguity, debug=debug, tree_class=tree_class)
+
+    def make_lexer(self, text):
+        return WithLexer.make_lexer(self, text).lex(None)
 
     def match(self, term, token):
         return term.name == token.type
@@ -219,7 +224,7 @@ class XEarley(_ParserFrontend):
             self.regexps[t.name] = lexer_conf.re_module.compile(regexp, lexer_conf.g_regex_flags)
 
     def parse(self, text, start):
-        return self._parse(text, start)
+        return self._parse(start, text)
 
 class XEarley_CompleteLex(XEarley):
     def __init__(self, *args, **kw):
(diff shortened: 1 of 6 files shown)


erezsh:
@MegaIng  Review it lightly or thoroughly, your choice. I just want to make sure I didn't miss anything obvious.

lark/parser_frontends.py
```diff
@@ -23,12 +23,18 @@ def get_frontend(parser, lexer):
         elif lexer == 'contextual':
             return LALR_ContextualLexer
         elif issubclass(lexer, Lexer):
+            class CustomLexerWrapper(Lexer):
```
MegaIng, line 26: This would prevent people from actually implementing the new interface (like I am doing over at lark-language-server). I would suggest expecting till 1.0 a `new_sytle` class-attribute that allows the lexer to use the new interface if set to true. After 1.0, old style will be unsupported.

I would also suggest adding a deprecation warning.

lark/parser_frontends.py
```diff
@@ -71,6 +77,18 @@ def _get_lexer_callbacks(transformer, terminals):
...
+    def __init__(self, lexer, postlexer):
+        self.lexer = lexer
+        self.postlexer = postlexer
+
+    def make_lexer_state(self, text):
+        return self.lexer.make_lexer_state(text)
+
+    def lex(self, lexer_state, parser_state):
+        i = self.lexer.lex(lexer_state, parser_state)
+        return self.postlexer.process(i)
```
MegaIng, line 90: Could it be interesting for the postlexer to access the lexer_state? I currently can not imagine a situation that would be solved by a CustomLexer anyway.

lark/parsers/lalr_parser.py
```diff
@@ -35,84 +35,116 @@ def parse(self, *args):
...
+    def position(self):
+        return self.state_stack[-1]
+
+    def __copy__(self):
+        return type(self)(
+            self.parse_table,
+            self.lexer, # XXX copy
+            self.callbacks,
+            self.start,
+            list(self.state_stack),
```
MegaIng, line 64: I find `list` here a bit confusing as a replacement for copy. I assume that you want to avoid copying the states with `debug=True`, but just using `copy` is more clear.

lark/parsers/lalr_parser.py
```diff
@@ -35,84 +35,116 @@ def parse(self, *args):
...
+            token = Token.new_borrow_pos('$END', '', token) if token else Token('$END', '', 0, 1, 1)
+            return state.feed_token(token, True)
+        except UnexpectedInput as e:
+            try:
+                e.puppet = ParserPuppet(self, state, state.lexer)
+            except NameError:
+                pass
+            if isinstance(e, UnexpectedCharacters):
+                s = state.lexer.state
+                p = s.line_ctr.char_pos
```
MegaIng, line 136: This makes an assumption over the layout of the LexerState (which is something I though we said we don't do?)

And I am also not sure what you are trying to accomplish here anyway.

lark/parsers/lalr_puppet.py
```diff
@@ -1,107 +1,55 @@
...
-            assert arg != end_state
-
-        assert action is Shift
-        state_stack.append(arg)
-        value_stack.append(token)
-
-    def copy(self):
+        return self.parser_state.feed_token(token)
+
+    def __copy__(self):
```
MegaIng, line 27: This is backwards compatibility breaking, since we exposed `ParserPuppet` with the `copy` method

lark/parsers/lalr_puppet.py
```diff
@@ -1,107 +1,55 @@
...
-            self._state_stack == other._state_stack and
-            self._value_stack == other._value_stack and
-            self._stream == other._stream and
-            self._start == other._start
-        )
+        return self.parser_state == other.parser_state and self.lexer_state == other.lexer_state
 
     def __hash__(self):
-        return hash((tuple(self._state_stack), self._start))
+        return hash((self.parser_state, self.lexer_state))
```
MegaIng, line 45: Neither of the State classes define `__hash__` make this an (almost) useless operation. I would suggest adding `__hash__` to both classes.

lark/parsers/lalr_parser.py
```diff
@@ -35,84 +35,116 @@ def parse(self, *args):
         return self.parser.parse(*args)
 
 
-class _Parser:
-    def __init__(self, parse_table, callbacks, debug=False):
-        self.parse_table = parse_table
-        self.callbacks = callbacks
-        self.debug = debug
+class ParserState:
+    __slots__ = 'parse_table', 'lexer', 'callbacks', 'start', 'state_stack', 'value_stack', 'start_state', 'end_state', 'states'
```
MegaIng, line 39: Only `lexer_thread`, `value_stack` and `state_stack` actually represent the State. I think the rest should be accesed via a `parser` attribute.

lark/parsers/lalr_puppet.py
MegaIng: Scrap that. Remove the implementation of `__hash__`. As stated the python docs, mutable classes shouldn't implement `__hash__`

MegaIng (changes requested):
I didn't look at the diffrence to the stubs, but considering you change non of them, I think something has to be done there.

But really nice in general!

lark/parser_frontends.py
erezsh: Yes, I plan to change this in v1.0

How would you prefer to expose the new interface until then?

lark/parser_frontends.py
erezsh: It might be interesting, though I can't think of a specific case either. But I plan to expose it in v1.0

lark/parser_frontends.py
MegaIng: Have a class attribute (maybe `new_style`?) that, when present and true makes this not wrap the class.

lark/parsers/lalr_parser.py
erezsh: I think they are functionally equivalent. Why does `debug=True` matter?

MegaIng: They are functionally equivalant, but I just think that `copy` is more clear. `debug=True` matters when you would use `deepcopy` instead, since that would deepcopy the state objects, which is not what we want. (We are using `deepcopy` one line down)

lark/parsers/lalr_parser.py
erezsh: `UnexpectedCharacters` is raised when the lexer cannot match any token. That means it won't advance, and resuming a parse will just cause it to return to the same character and fail with the same error.

This advances the character pos by 1, so it will skip the "bad" character if the parse is resumed.

erezsh: > This makes an assumption over the layout of the LexerState (which is something I though we said we don't do?)

Why not? It's a defined class in `lexer.py`

I don't remember the context. Why shouldn't we make any assumption about it?

lark/parsers/lalr_puppet.py
erezsh: I agree, but it's a very new feature, so I doubt anyone had a chance to use it yet.

But okay, I'll keep the `copy()` method.

lark/parsers/lalr_parser.py
MegaIng: This is something that should done in `on_error` or in `Lark.parse`, but not here, if at all. This would as example also break when you try to find the character that was blocking via `text[line_ctr.char_pos]`. I see that the goal is to make `lambda e: True` work, but I think a better solution would be to export maybe a `skip_error` function that does all this.

Once again, this might break with CustomLexer, and even if we don'T expose the new Lexer protocol now, it will break in 1.0.

lark/parsers/lalr_parser.py
erezsh: I save them in `ParserState` for performance reasons, otherwise the `feed_token()` method would have to fetch them again and again in every iteration.

I can prefix them with `_` if you think that would make it clearer.

lark/parsers/lalr_parser.py
MegaIng: gitter chat Sep 26. (Can you somehow to a perma link?). I am using a custom lexer with a custom lexerstate for Lark-language server to allow `on_error` to skip over unknown tokens. Check my branch over there. You are here assuming that the lexer state has a `line_ctr` attribute which is not true. (and I could provided it, but it is completly useless since the tokens are preparsed.)

lark/parsers/lalr_parser.py
MegaIng: Did you measure it? I think it doesn't matter whether or not you get the attribute twice or just once. (e.g. `self.lexer.parse_table` vs `self.parse_table`)

erezsh: I did measure it actually, but I didn't try to put `__slots__` in `ParseTable`. I guess it's worth a try.

lark/parser_frontends.py
MegaIng: Yeah, let it be. Not worth the hassle if we can't think of a usecase.

lark/parsers/lalr_parser.py
erezsh: But then your lexer would never throw `UnexpectedCharacters`, right?

erezsh:
Since the interface didn't change, I can't think of why the stubs should. I tried looking it over, but seems like everything that every interface we exposed remain unchanged (which is the goal, since we want to remain backward compatible).

lark/parsers/lalr_parser.py
MegaIng: That is true. But if we required that (a Lexer that might throw `UnexpecedCharacters` has to have a `LexerState` with a `line_ctr` attribute we have to clearly document that somewhere) And I can easily imagine situation where position information might be unusable (A lexer that directly reads from a stream, or ask the user for new input if the internal buffer is empty like a shell). Also note that this would create an AttributeException even if we don't use `on_error`.

MegaIng:
Yes, we aren't breaking anything, but I would say we should expose this new interface in the stubs.

erezsh:
But then we won't be able to change it until v1.0 :)

I think it's better to wait until it's stable before advertising new API, just so we won't be stuck from making any improvements (like we are with the CustomLexer)

MegaIng:
Yeah ok sure. I will still be using/testing it with `lark-language-server`, which is why I wanted to not block the new interface for CustomLexer `:-P`

erezsh:
Will updating the `pyi` files help you to work on the language server?

lark/parsers/lalr_parser.py
erezsh: You're right, it wasn't in the right place.

I refactored it into `Lark.parse()`. Now it won't trigger if `on_error` is None.

MegaIng:
It will just reduce the IDE complains. Not important, so we can keep it 'hidden.' It will not really help me.

lark/parsers/lalr_parser.py
MegaIng: I am still not completely certain that we shouldn't let the user explicit optin/optout of this behavior, but we can't wait till this because a problem. Since we aren't exposing the interface anyway, we can wait till 1.0 to make a final decision.

erezsh:
@MegaIng Okay, I think that's all of it.

MegaIng (approved):

MegaIng:
I also think so. We might at some point come back to some of this though. (Btw. are you running any coverage tests?)

erezsh:
Right now I don't run any coverage tests.

erezsh:
@MegaIng Just checked. Coverage is at 88%. biggest "miss" is in `earley_forest.py`

lark/parsers/lalr_parser.py
```diff
@@ -35,84 +35,123 @@ def parse(self, *args):
...
-    def parse(self, seq, start, set_state=None, value_stack=None, state_stack=None):
-        token = None
-        stream = iter(seq)
-        states = self.parse_table.states
-        start_state = self.parse_table.start_states[start]
-        end_state = self.parse_table.end_states[start]
 
-        state_stack = state_stack or [start_state]
-        value_stack = value_stack or []
+class ParserState:
```
MegaIng, line 52: I failed, I should have caught this in the original review: classes should be new style and explicit inherit from object.

erezsh: Why is that? Iirc "new style" in Python 2.7 is only for edge features like __getattribute__ and such.

MegaIng: And for `type(self)` like used in a `__copy__` method.

Issue #717, opened 2020-10-13
Show values of terminals in errors, instead of names, when appropriate


erezsh:

omega16:
So, first thing would be to define "apropiate" .

Something like :

- If terminal comes from a simple string and string length is less than some integer, except from debug mode in that case show full string
- If terminal comes from union of strings and total length less than some int... same a before.
- If terminal comes from anon regex it would be great to show them in debug mode  or at least show them as `anon_regex_n`
- if terminal is an anon mix strings and regex show something like `anon_mix_n` 

Is there other kinds of terminals?

charles-esterbrook:
Even when not in debug mode, `/[A-Z]+\-\d+/` is still more informative than `__ANON_17`. But if you stick with `ANON`, then can we get them numbered by their line and column in the grammar? So `__ANON_17` would be `__ANON_L45_C23` or similar. Actually, that labeling scheme for anonymous terminals is probably a good idea regardless of what we do with error messages.

erezsh:
@charles-esterbrook That's a nice idea!

MegaIng:
Is someone working on this right now? Otherwise I will give it a try.

erezsh:
Go ahead. I didn't get started yet, and I don't think anyone else did.

PR #727, opened 2020-10-20
improve error message with token source.

MegaIng:
rename LexerConf.terminals to LexerConf.tokens
Make Exception message generation lazy
Made a few classes new-style

---

Addresses #717. @charles-esterbrook, @omega16 your feedback would be appreciated.

Not quite finished: Not implemented for Regex, might be improvable for strings.

Should work for all valid lexer-parser combinations.


lark/lexer.py
```diff
@@ -76,14 +76,15 @@ def max_width(self):
 
 
 class TerminalDef(Serialize):
-    __serialize_fields__ = 'name', 'pattern', 'priority'
+    __serialize_fields__ = 'name', 'pattern', 'priority', 'nice_print'
```
erezsh, line 79: Why have a `nice_print` attribute?

I think it's better to have it as a method that returns a different result based if `self.pattern` is `PatternStr` and `PatternRE`

Also, you can just call it `user_repr()`

lark/exceptions.py
```diff
@@ -109,32 +110,54 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
-            message += '\nPrevious tokens: %s\n' % ', '.join(repr(t) for t in token_history)
-
-        super(UnexpectedCharacters, self).__init__(message)
-
+            self._s = seq[lex_pos]
+        self._context = self.get_context(seq)
+        
+        super(UnexpectedCharacters, self).__init__()
+
+    def __str__(self):
```
erezsh, line 150: It's probably better to have a `_str()` method, and have `__str__(self): return self._str()`. That way you can debug it with less pain.

lark/exceptions.py
```diff
@@ -109,32 +110,54 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
                             candidate = label, False
 
         return candidate[0]
+    
+    def _format_terminals(self, names):
+        if self._all_terminals:
+            t = []
+            for name in names:
+                try:
+                    t.append(next(t.nice_print for t in self._all_terminals if t.name == name))
```
erezsh, line 119: Looks like `all_terminals` should be a dict.

MegaIng: Yeah, but since this should only be executed when we actually print the error message, I decided that it wasn't worth the hassle (But maybe I overlook something and we just have dictionaries lying around anyway. The inefficiency of this code is way I decided to make it lazy.

lark/exceptions.py
MegaIng: I don't really see a benefit of `e._str()` over `str(e)`.

lark/lexer.py
MegaIng: My plan is to improve this to actually contain the raw source code as accessible in `load_grammar`. (and line information etc.) This information isn't being carried around at the moment, and since it doesn't need to be used for anything but this I decided to just generate the string early. But maybe `user_repr` is a better name. `nice_print` is the best I could come up with.

erezsh: That could be nice. And then instead of `all_terminals` we can have a more generic term like `grammar_info`

lark/exceptions.py
erezsh: This isn't about performance, it's about readability.

lark/lexer.py
MegaIng: Do you mean that I just rename `all_terminals` to `grammar_info` without changing the content? Or do you want to put more info in the attribute?

erezsh: You just said you want to put more content. So yeah, I meant more content, not just rename.

MegaIng: Well, I didn't plan to add more content to the Exception classes `_all_terminals` attribute. Just adding more content to TerminalDef.user_repr.

MegaIng:
I would say that this is pretty much read.

lark/load_grammar.py
```diff
@@ -426,9 +423,9 @@ def _literal_to_pattern(literal):
 
     if literal.type == 'STRING':
         s = s.replace('\\\\', '\\')
-        return PatternStr(s, flags)
+        return PatternStr(s, flags, raw=literal.value)
```
erezsh, line 426: Is it really necessary to save the original value?

lark/exceptions.py
```diff
@@ -109,32 +110,56 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
+    
+    def _format_terminals(self, names):
+        if self._all_terminals:
+            if isinstance(self._all_terminals, list):
+                self._all_terminals = {t.name: t for t in self._all_terminals}
+            ts = []
+            for name in names:
+                try:
+                    ts.append(self._all_terminals[name].user_repr)
+                except StopIteration:
```
erezsh, line 122: StopIteration?

erezsh: Also, like u said, it should never happen.

lark/exceptions.py
```diff
@@ -109,32 +110,56 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
         self.state = state
+        self._all_terminals = _all_terminals
 
         self.allowed = allowed
         self.considered_tokens = considered_tokens
+        self.token_history = token_history
 
         if isinstance(seq, bytes):
-            _s = seq[lex_pos:lex_pos+1].decode("ascii", "backslashreplace")
+            self._s = seq[lex_pos:lex_pos+1].decode("ascii", "backslashreplace")
```
erezsh, line 145: If you're going to save it, use a better name than `_s`

lark/lexer.py
```diff
@@ -76,14 +77,15 @@ def max_width(self):
 
 
 class TerminalDef(Serialize):
-    __serialize_fields__ = 'name', 'pattern', 'priority'
+    __serialize_fields__ = 'name', 'pattern', 'priority', 'user_repr'
     __serialize_namespace__ = PatternStr, PatternRE
 
-    def __init__(self, name, pattern, priority=1):
+    def __init__(self, name, pattern, priority=1, user_repr=None):
```
erezsh, line 83: `user_repr` should be a method. There is no reason for it to be a parameter.

lark/load_grammar.py
```diff
@@ -317,6 +317,7 @@ def pattern(self, p):
             raise GrammarError(u'Conflicting flags for the same terminal: %s' % p)
 
         term_name = None
+        user_repr = p.raw # This will always be ok, independent of what term_name we end up using
```
erezsh, line 320: Why is this needed here? You can do it inside TerminalDef

lark/exceptions.py
```diff
@@ -109,32 +110,56 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
                             candidate = label, False
 
         return candidate[0]
+    
+    def _format_terminals(self, names):
+        if self._all_terminals:
+            if isinstance(self._all_terminals, list):
```
erezsh, line 116: A bit of a code smell. If all of Lark needs access to the same information, maybe it's better to just pass it prepared, instead of half-baked.

erezsh (changes requested):

lark/load_grammar.py
MegaIng: Do you have a better idea? The goal is to show exactly this.

MegaIng:
I hope I didn't make a mistake while merging.

MegaIng:
Note that currently this doesn't work for earley. I didn't find a way of getting access to this information.

charles-esterbrook:
> Note that currently this doesn't work for earley. I didn't find a way of getting access to this information.

Poor Earley parser... always a bridesmaid, never a bride.

lark/exceptions.py
```diff
@@ -116,44 +118,71 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
 
         return candidate[0]
 
+    def _format_terminals(self, terminals):
+        if self._terminals_by_name:
+            ts = []
+            for ter in terminals:
+                ts.append(self._terminals_by_name[ter].user_repr())
```
erezsh, line 125: Why not `ts = [t.user_repr() for t in self._terminals_by_name.values()]` ?

lark/exceptions.py
```diff
@@ -116,44 +118,71 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
         self.line = -1
         self.column = -1
+        self._terminals_by_name = _terminals_by_name
+
+        super(UnexpectedEOF, self).__init__()
 
-        message = ("Unexpected end-of-input. Expected one of: \n\t* %s\n" % '\n\t* '.join(x.name for x in self.expected))
-        super(UnexpectedEOF, self).__init__(message)
+    def __str__(self):
+        # Be aware: Broken __str__ for Exceptions are terrible to debug. Make sure there is as little room as possible for errors
```
erezsh, line 145: I gave it some thought since last time, and I don't think Lark is the right place for general Python tips.

lark/exceptions.py
```diff
@@ -116,44 +118,71 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
+            for ter in terminals:
+                ts.append(self._terminals_by_name[ter].user_repr())
+        else:
+            ts = terminals
+        return "Expected one of: \n\t* %s\n" % '\n\t* '.join(ts)
+
+
 class UnexpectedEOF(ParseError, UnexpectedInput):
-    def __init__(self, expected, state=None):
+    def __init__(self, expected, state=None, _terminals_by_name=None):
```
erezsh, line 132: I don't think anywhere in Lark there are parameters that start with an underscore. Why have them here?

lark/exceptions.py
```diff
@@ -116,44 +118,71 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
-            message += '\nPrevious tokens: %s\n' % ', '.join(repr(t) for t in token_history)
-
-        super(UnexpectedCharacters, self).__init__(message)
+            self.char = seq[lex_pos]
+        self._context = self.get_context(seq)
+
+        super(UnexpectedCharacters, self).__init__()
+
+    def __str__(self):
+        # Be aware: Broken __str__ for Exceptions are terrible to debug. Make sure there is as little room as possible for errors
```
erezsh, line 176: Same remark about general python tips

lark/exceptions.py
```diff
@@ -116,44 +118,71 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
+        # If you run into this, add an `except Exception as e: print(e); raise e` or similar.
+        message = "Unexpected end-of-input. "
+        message += self._format_terminals(self.expected)
+        return message
 
 
 class UnexpectedCharacters(LexError, UnexpectedInput):
-    def __init__(self, seq, lex_pos, line, column, allowed=None, considered_tokens=None, state=None, token_history=None):
+    def __init__(self, seq, lex_pos, line, column, allowed=None, considered_tokens=None, state=None, token_history=None,
+                 _all_terminals=None):
```
erezsh, line 155: Same question about underscored params

lark/exceptions.py
```diff
@@ -163,31 +192,35 @@ class UnexpectedToken(ParseError, UnexpectedInput):
...
         self.column = getattr(token, 'column', '?')
         self.pos_in_stream = getattr(token, 'pos_in_stream', None)
         self.state = state
 
         self.token = token
-        self.expected = expected     # XXX deprecate? `accepts` is better
+        self.expected = expected  # XXX deprecate? `accepts` is better
         self.considered_rules = considered_rules
         self.puppet = puppet
+        self._all_terminals = all_terminals
```
erezsh, line 207: Looks like this isn't used anywhere?

lark/exceptions.py
```diff
@@ -163,31 +192,35 @@ class UnexpectedToken(ParseError, UnexpectedInput):
...
         self.token_history = token_history
 
-        # TODO Only calculate `accepts()` when we need to display it to the user
```
erezsh, line 179: You removed this comment, but didn't change the computation.

`accepts()` is still being called even if it will never be displayed to the user.

Except now, repeated calls to `accepts()` will re-compute it every time instead of just once, so it's actually worse now. (that can be solved with memoization, but then we're back where we started)

lark/lexer.py
```diff
@@ -91,6 +92,12 @@ def __init__(self, name, pattern, priority=1):
 
     def __repr__(self):
         return '%s(%r, %r)' % (type(self).__name__, self.name, self.pattern)
+    
+    def user_repr(self):
+        if self.name.startswith('__'): # We represent a generated terminal
+            return self.pattern.raw or self.name
```
erezsh, line 98: Isn't this a little confusing? What if the raw pattern looks like a name?

lark/parsers/earley.py
```diff
@@ -302,7 +302,7 @@ def parse(self, lexer, start):
         # this column. Find the item for the start_symbol, which is the root of the SPPF tree.
         solutions = [n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0]
         if not solutions:
-            expected_terminals = [t.expect for t in to_scan]
+            expected_terminals = [t.expect.name for t in to_scan]
```
erezsh, line 305: Why did you add this? I think you're breaking `match_examples()` with this.

erezsh (changes requested):
I added a few comments..

I might have more afterwards.

lark/exceptions.py
MegaIng: Because `self.terminals!=self._terminals_by_name`.

lark/exceptions.py
MegaIng: Sure

lark/exceptions.py
MegaIng: To mirror the attribute name, but that isn't useful.

lark/exceptions.py
MegaIng: Stupid me overlook this when renaming.

lark/lexer.py
MegaIng: raw pattern will always be the full source code, e.g. `"<>" or ``/[A-Z]+/i`

lark/parsers/earley.py
MegaIng: This is the same behaviour as displayed in line 247 for `UnexpectedToken`. inside of `UnexpectedEOF.__init__` we are just using `.name` anyway.

lark/exceptions.py
erezsh: They are the same length, so how are they different?

lark/lexer.py
erezsh: You set it as `raw=literal.value`

What if raw="LOOKS_LIKE_TERMINAL" ?

lark/parsers/earley.py
erezsh: Ah, okay. I confused it with a different parameter.

lark/exceptions.py
MegaIng: `terminals` is the collection of names that is passed to `_format_terminals`. This is the expect set. I will rename this.

`_terminals_by_name` is a dictionary.

lark/lexer.py
MegaIng: What is your suggestion?

erezsh: Maybe `repr(raw)` should be good enough.

But I can't say for sure, users are very creative in being confused.

MegaIng: That would be more confusing. Then one would get `'"<>"'` as the what is printed, e.g. the string is `'\'"<>"\''`

MegaIng: or with your example `'"LOOKS_LIKE_TERMINAL"'`

MegaIng: The quotes are already included.

lark/exceptions.py
MegaIng: Better now?

lark/exceptions.py
erezsh: But isn't `set(terminals) == set(_terminals_by_name)`?

lark/exceptions.py
```diff
@@ -118,47 +118,44 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
 
         return candidate[0]
 
-    def _format_terminals(self, terminals):
+    def _format_excepts(self, excepts):
         if self._terminals_by_name:
             ts = []
-            for ter in terminals:
+            for ter in excepts:
```
erezsh, line 124: what is `excepts`?

MegaIng: The argument to the function

erezsh: What does the word "excepts" mean in this context?

MegaIng: Typo: meant 'expected'/'expects'. English isn't my first language.

erezsh: It's not a typo if it appears 20 times :)

MegaIng: 7, most of which with autocomplete.

lark/exceptions.py
erezsh: Awaiting response.

lark/common.py
```diff
@@ -1,24 +1,36 @@
...
 class LexerConf(Serialize):
-    __serialize_fields__ = 'tokens', 'ignore', 'g_regex_flags', 'use_bytes', 'lexer_type'
+    __serialize_fields__ = 'terminals', 'ignore', 'g_regex_flags', 'use_bytes', 'lexer_type'
     __serialize_namespace__ = TerminalDef,
 
-    def __init__(self, tokens, re_module, ignore=(), postlex=None, callbacks=None, g_regex_flags=0, skip_validation=False, use_bytes=False):
-        self.tokens = tokens    # TODO should be terminals
+    def __init__(self, terminals, re_module, ignore=(), postlex=None, callbacks=None, g_regex_flags=0, skip_validation=False, use_bytes=False):
+        self.terminals = terminals
+        self.terminals_by_names = {t.name: t for t in self.terminals}
```
erezsh, line 15: Why `.._by_names`? In other places it's `.._by_name`

lark/exceptions.py
```diff
@@ -163,31 +187,37 @@ class UnexpectedToken(ParseError, UnexpectedInput):
...
-        # This will improve performance when doing automatic error handling
-        self.accepts = puppet and puppet.accepts()
+        super(UnexpectedToken, self).__init__()
 
-        message = ("Unexpected token %r at line %s, column %s.\n"
-                   "Expected one of: \n\t* %s\n"
-                   % (token, self.line, self.column, '\n\t* '.join(self.accepts or self.expected)))
+    @property
+    def accepts(self):
+        if self._accepts is _not_set_marker:
```
erezsh, line 210: No... just no

lark/exceptions.py
```diff
@@ -116,45 +118,67 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
+    def __str__(self):
+        message = "No terminal defined for '%s' at line %d col %d" % (self.char, self.line, self.column)
+        message += '\n\n' + self._context
+        if self.allowed:
+            message += self._format_expected(self.allowed)
+        if self.token_history:
+            message += '\nPrevious tokens: %s\n' % ', '.join(repr(t) for t in self.token_history)
+        return message
 
+_not_set_marker = object()
```
erezsh, line 181: Name this `_NO_VALUE` and put it at the top of the module.

erezsh:
Fixed it myself.

(conflict is due to rebase)

PR #729, opened 2020-10-21
Improved setup.py

ThatXliner:
Make the README the repo's README and a less hacky way to get the `__version__` variable from lark.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,9 +2,13 @@ try:
     import regex as re
 except ImportError:
     import re
-from setuptools import find_packages, setup
+from setuptools import setup  # , find_packages  # unused
 
-__version__ ,= re.findall('__version__ = "(.*)"', open('lark/__init__.py').read())
+from lark import __version__
+
+#__version__ ,= re.findall('__version__ = "(.*)"', open('lark/__init__.py').read())
+with open("README.md") as f:
+    README = f.read()
 
 setup(
     name = "lark-parser",
@@ -19,7 +23,7 @@ setup(
         "nearley": ["js2py"]
     },
 
-    package_data = {'': ['*.md', '*.lark'], 'lark-stubs': ['*.pyi']},
+    package_data = {'': ['*.md', '*.lark'], 'lark-stubs': ['*.pyi']},  # Maybe a MANIFEST.ini would be used here
 
     test_suite = 'tests.__main__',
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,9 +2,13 @@
     import regex as re
 except ImportError:
     import re
-from setuptools import find_packages, setup
+from setuptools import setup  # , find_packages  # unused
 
-__version__ ,= re.findall('__version__ = "(.*)"', open('lark/__init__.py').read())
+from lark import __version__
```
MegaIng, line 7: I am pretty sure the goal was to *not* import `lark`

setup.py
```diff
@@ -2,9 +2,13 @@
     import regex as re
 except ImportError:
     import re
-from setuptools import find_packages, setup
+from setuptools import setup  # , find_packages  # unused
```
MegaIng, line 5: Why comment it out and not remove it? Why add `# unused`?

setup.py
```diff
@@ -19,7 +23,7 @@
         "nearley": ["js2py"]
     },
 
-    package_data = {'': ['*.md', '*.lark'], 'lark-stubs': ['*.pyi']},
+    package_data = {'': ['*.md', '*.lark'], 'lark-stubs': ['*.pyi']},  # Maybe a MANIFEST.ini would be used here
```
MegaIng, line 26: What do you want to add with this comment?

MegaIng (changes requested):

setup.py
ThatXliner: Why though?

setup.py
```diff
@@ -31,30 +35,8 @@
...
- - Unicode fully supported
- - Python 2 & 3 compatible
- - Automatic line & column tracking
- - Standard library of terminals (strings, numbers, names, etc.)
- - Import grammars from Nearley.js
- - Extensive test suite
- - And much more!
-''',
-
+    long_description=README,
```
MegaIng, line 38: To do this, `README.md` should also be changed, since there are a few things that will break when not used on github.

setup.py
ThatXliner: Because if it turns out that we need it, we can always uncomment it. You can remove it if you want

setup.py
```diff
@@ -2,9 +2,13 @@
     import regex as re
 except ImportError:
     import re
-from setuptools import find_packages, setup
+from setuptools import setup  # , find_packages  # unused
 
-__version__ ,= re.findall('__version__ = "(.*)"', open('lark/__init__.py').read())
```
erezsh, line 7: The reason I'm using `re` is so setup.py could be run without having to install dependencies.

See pt 6. https://packaging.python.org/guides/single-sourcing-package-version/

setup.py
```diff
@@ -31,30 +35,8 @@
     keywords = "Earley LALR parser parsing ast",
     url = "https://github.com/erezsh/lark",
     download_url = "https://github.com/erezsh/lark/tarball/master",
-    long_description='''
```
erezsh, line 34: That's a nice idea, but I think lark's readme is too long for this field, and might not display correctly in non-web situations.

But I'm open to be persuaded.

setup.py
erezsh: Heh I'm late to the party. See my comments.

setup.py
ThatXliner: But `import lark` is a relative import.

ThatXliner: I see.

setup.py
```diff
@@ -19,7 +23,7 @@
         "nearley": ["js2py"]
     },
 
-    package_data = {'': ['*.md', '*.lark'], 'lark-stubs': ['*.pyi']},
+    package_data = {'': ['*.md', '*.lark'], 'lark-stubs': ['*.pyi']},  # Maybe a MANIFEST.ini would be used here
```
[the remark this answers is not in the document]
ThatXliner: A `MANIFEST.ini` is a file that tells setuptools which pieces of metadata should be included. See https://packaging.python.org/guides/using-manifest-in/

setup.py
MegaIng: No? A relativ import starts with a `.`, so `import .lark` would be relative. And why does that matter: You are still importing the entire package, including all submodules that might import dependencies.

ThatXliner: Oh ok.

setup.py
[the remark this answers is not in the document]
MegaIng: And...? Why is that relevant?

PR #730, opened 2020-10-22
cache=True now uses a temp directory instead of working directory

erezsh:


MegaIng (approved):
I don't know if this is a good idea. But it probably is.

erezsh:
I think it should be fine.

Polluting the working directory is kinda annoying for me, so it must be annoying for other users as well.

PR #731, opened 2020-10-25
Add documentation on the `x` regexp flag

decorator-factory:
Multiline regexps with the `x` flag were added in https://github.com/lark-parser/lark/pull/645


erezsh:
Thanks!

ThatXliner:
Why not just slap a regex tutorial at the bottom or point a link to one!

erezsh:
There are hundreds of those, anyone can google. If you have a special recommendation I will consider it.

PR #735, opened 2020-10-27
Add on_error to parse lark.pyi

fabioz:


erezsh:
It's not `[Any]`, it's `[UnexpectedInput]`

MegaIng:
If I remember correctly, this was a concision decision to not include it in the stubs, since we are expecting to make changes in this API in the near future (e.g. before 1.0). You are open to use and test it, but we might break your code in minor releases (unlikely, but could happen)

erezsh:
@MegaIng I was referring specifically to the `ParserState` / `LexerState` interfaces. After all, `ParserPuppet` is already in the pyi interface.

fabioz:
I just updated to use `UnexpectedInput`.

erezsh:
Looks alright to me.

@MegaIng Do you still think we should wait with this?

fabioz:
As a note, I don't see any issue in breaking the API in future releases (and fixing the stub when that happens).

It's a bit more troubling for me using the current version and having that typing error as I'll have to silence it.

My rationale is the following: just because it's typed doesn't mean it can never change later on... in fact having the typing check it when I upgrade and then having the error seems better than having to specifically silence that error now and when I upgrade I'd have a runtime error instead of a typing error.

PR #745, opened 2020-11-05
Fix SPPF cycle handling

### LINKED ISSUE #741: Bug in Earley parser when there is ambiguity in empty rules
**Describe the bug**

The Earley parser creates an incorrect tree

**To Reproduce**

```python
from lark import Lark
grammar = """
a: _x
_x: _x? b
b:
"""

p = Lark(grammar, start='a', ambiguity='explicit')
```

This will throw an `IndexError` because the wrong node constructor gets called.

The non-expanding variation doesn't throw an error, but creates a non-sensical result:

```python
from lark import Lark
grammar = """
a: x
x: x? b
b:
"""

p = Lark(grammar, start='a', ambiguity='explicit')
print(p.parse('').pretty())
```

Output
```python
a
  _ambig
    x
      b
    x
      b
```

Earlier versions of Lark throw `ParseError: Infinite recursion in grammar!`, which is technically more correct, although not very useful either.

chanicpanic:
This fixes #741 and a couple other edge cases related to SPPF cycles. Using `ambiguity='resolve'` when the SPPF has cycles is now guaranteed to return a tree.

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -315,7 +315,7 @@ class Parser:
 
         if self.tree_class is not None:
             # Perform our SPPF -> AST conversion
-            transformer = ForestToParseTree(self.tree_class, self.callbacks, self.forest_sum_visitor and self.forest_sum_visitor(), self.resolve_ambiguity)
+            transformer = ForestToParseTree(self.tree_class, self.callbacks, self.forest_sum_visitor and self.forest_sum_visitor(), self.resolve_ambiguity, self.debug)
             return transformer.transform(solutions[0])
 
         # return the root of the SPPF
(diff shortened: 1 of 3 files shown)


MegaIng:
Can you prevent the output from being cluttered in the test runs? The logging messages are a little annoying.

chanicpanic:
The logging level is set to INFO at the start of `test_parser.py`. If this is not necessary, it could be removed to prevent the logging messages during the test run.

Otherwise, I could add a debug attribute to `ForestToParseTree` and wrap the logging line in an if statement.

chanicpanic:
I prevented the logging messages during the test run. I also set the lexer in the tests, something I had forgotten to do initially.

erezsh:
Thanks!

I only did a small adjustment - I changed `logger.warning` into `logger.debug`.

PR #747, opened 2020-11-10
Recons terminals unicode

julienmalard:
This test checks that a parse tree built with a grammar containing only ascii characters can be reconstructed with a grammar that has unicode rules (or vice versa). The original bug assigned ANON terminals to unicode keywords, which offsets the ANON terminal count in the unicode grammar and causes subsequent identical ANON tokens (e.g., `+=`) to mis-match between the two grammars.

I know that this is an uncommon situation, but it is central to use of Lark as a transpiler (e.g., [லஸ்ஸி](https://லஸ்ஸி.இந்தியா)) and the fix was quite simple, so I thought I'd try to send a pull request. Hope you like it!

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -5,7 +5,7 @@ import sys
 from copy import copy, deepcopy
 from io import open
 
-from .utils import bfs, eval_escaping, Py36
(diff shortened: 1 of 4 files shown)


julienmalard:
Deleting because I just noticed this branch was incorrectly rebased. See #748 instead. Sorry!

PR #748, opened 2020-11-10
Reconstruct with unicode terminals

julienmalard:
This test checks that a parse tree built with a grammar containing only ascii characters can be reconstructed with a grammar that has unicode rules (or vice versa). The original bug assigned ANON terminals to unicode keywords, which offsets the ANON terminal count in the unicode grammar and causes subsequent identical ANON tokens (e.g., +=) to mis-match between the two grammars.

I know that this is an uncommon situation, but it is central to use of Lark as a transpiler (e.g., [லஸ்ஸி](https://லஸ்ஸி.இந்தியா/)) and the fix was quite simple, so I thought I'd try to send a pull request. Hope you like it!

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -6,7 +6,7 @@ from copy import copy, deepcopy
 from io import open
 import pkgutil
 
-from .utils import bfs, eval_escaping, Py36, logger, classify_bool
+from .utils import bfs, eval_escaping, Py36, logger, classify_bool, isalnum, isalpha
 from .lexer import Token, TerminalDef, PatternStr, PatternRE
 
 from .parse_tree_builder import ParseTreeBuilder
@@ -328,9 +328,9 @@ class PrepareAnonTerminals(Transformer_InPlace):
                 try:
                     term_name = _TERMINAL_NAMES[value]
                 except KeyError:
-                    if value.isalnum() and value[0].isalpha() and value.upper() not in self.term_set:
+                    if isalnum(value) and isalpha(value[0]) and value.upper() not in self.term_set:
                         with suppress(UnicodeEncodeError):
-                            value.upper().encode('ascii') # Make sure we don't have unicode in our terminal names
+                            value.upper().encode('utf8')  # Why shouldn't we have unicode in our terminal names?
                             term_name = value.upper()
 
                 if term_name in self.term_set:
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -6,7 +6,7 @@ from copy import copy, deepcopy
 from io import open
 import pkgutil
 
-from .utils import bfs, eval_escaping, Py36, logger, classify_bool, is_id_continue, isalpha
+from .utils import bfs, eval_escaping, Py36, logger, classify_bool, is_id_continue, is_id_start
 from .lexer import Token, TerminalDef, PatternStr, PatternRE
 
 from .parse_tree_builder import ParseTreeBuilder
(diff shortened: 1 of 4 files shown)


lark/load_grammar.py
```diff
@@ -328,9 +328,9 @@ def pattern(self, p):
                 try:
                     term_name = _TERMINAL_NAMES[value]
                 except KeyError:
-                    if value.isalnum() and value[0].isalpha() and value.upper() not in self.term_set:
+                    if isalnum(value) and isalpha(value[0]) and value.upper() not in self.term_set:
                         with suppress(UnicodeEncodeError):
-                            value.upper().encode('ascii') # Make sure we don't have unicode in our terminal names
+                            value.upper().encode('utf8')  # Why shouldn't we have unicode in our terminal names?
```
MegaIng, line 333: What is the point of this line? And especially the comment?

lark/utils.py
```diff
@@ -12,6 +13,17 @@
 # By default, we should not output any log messages
 logger.setLevel(logging.CRITICAL)
 
+def isalnum(x):
```
MegaIng, line 16: ~~Why are we behaving differently than `str.isalnum`?~~

~~I know that you didn't write this first, but I think that we should still add a comment explaining this.~~

Ok. I found it. This method has the wrong name. It should be named `is_id_continue`. (and should still have an explaining comment)

MegaIng (changes requested):

lark/utils.py
julienmalard: Hello,
I had named it `isalnum` because it is the unicode-compatible version of `str.isalnum`. But I changed the name now to avoid confusion.

lark/load_grammar.py
julienmalard: I had just been trying to make minimal changes, but I think you're right in that this line doesn't do anything anymore. Tests all pass without it so I removed these few lines. Thank you!

lark/utils.py
MegaIng: It isn't. That is my point. Please add a reference to PEP 3131, and also change the name of `isalpha` to `is_id_continue`.

Also, I think there is a better way of going about this somehow using the information that python internally uses, but I am not sure what it is.

julienmalard: Hello @MegaIng : In that case the distinction eludes me (if you can share some pointers please do so!). I have added the reference to PEP 3131. The problem with renaming `isalpha` is that `isalphanum` has just been renamed to `is_id_continue`, and both serve different purposes (`isalpha` should match only `id_start` characters, and not numeric characters).

MegaIng: Read PEP 3131. The `str.isalphanum` method is already unicode compatible. But this list of unicode categories includes a few additional things that are valid inside python identifiers, but aren't really alphabetic. As example 'connector punctuations (Pc)' or the underscore.

julienmalard: Thank you! When I said "unicode compatible" I think I meant "international compatible". Since `isalpha` is, in practice here, used to detect characters that make up a word, and `isalphanum` the same but including numerals, the goal was to make a function that behaved the same no matter which human language or script the word belongs to.

That said, I am not sure anymore that `is_id_continue` is a good name, because it leads to suppose that `_` would pass, when it does not (and should not, as the original `'_'.isalpha()` returns `False`).

Anyways, the current functionality is correct, and as this is simply a question of the best name for the function I will wait for your final recommendations.

MegaIng: I think `_` should pass. And sorry, I mistyped, `is_id_start` was what it was supposed to be.

julienmalard: Done! Thank you.

Issue #758, opened 2020-11-14
The "custom lexer" doc page has problems


charles-esterbrook:
**Describe the bug**

The "customer lexer" page has a broken link, an API bug and the docs could use improvements:

* https://lark-parser.readthedocs.io/en/latest/examples/advanced/custom_lexer.html

(1) The URL says "latest" but I don't know if that means latest release or latest master branch in git. I clicked "Edit on GitHub" to investigate but that link is https://github.com/lark-parser/lark/blob/master/docs/examples/advanced/custom_lexer.rst which is 404.

(2) The given example no longer runs due to the `lex()` method having insufficient parameters:

```
Traceback (most recent call last):
  File "custom_lexer.py", line 46, in <module>
    test()
  File "custom_lexer.py", line 38, in test
    tree = parser.parse(data)
  File "...\lark-github\lark\lark.py", line 494, in parse
    return self.parser.parse(text, start=start)
  File "...\lark-github\lark\parser_frontends.py", line 138, in parse
    return self._parse(start, self.make_lexer(text))
  File "...\lark-github\lark\parser_frontends.py", line 73, in _parse
    return self.parser.parse(input, start, *args)
  File "...\lark-github\lark\parsers\lalr_parser.py", line 35, in parse
    return self.parser.parse(*args)
  File "...\lark-github\lark\parsers\lalr_parser.py", line 129, in parse
    return self.parse_from_state(parser_state)
  File "...\lark-github\lark\parsers\lalr_parser.py", line 135, in parse_from_state
    for token in state.lexer.lex(state):
  File "...\lark-github\lark\lexer.py", line 449, in lex
    return self.lexer.lex(self.state, parser_state)
TypeError: lex() takes 2 positional arguments but 3 were given
```

After adding the parameter to receive the parser state, the next exception is:

```
TypeError: 'LexerState' object is not iterable
```

(3) In addition to code corrections, it would make sense for that page to describe that the two parameters are `lark.lexer.LexerState` and `lark.parsers.lalr_parser.ParserState`, what they provide, motivation for the design, etc.

It would also be interesting to note if there are any differences in customer lexers when using LALR vs. Earley.

**To Reproduce**

1. Click the "Edit" button.
2. Run the code on the given page with the latest master branch of lark-parser.
3. Read the page as if you were a noob.

MegaIng:
> The "customer lexer" page has a broken link, an API bug and the docs could use improvements:

[...]

PR #759, opened 2020-11-15
Make `LALR_CustomLexerWrapper` behave correctly + tests

MegaIng:
This fixes the bug in #758, e.g. the example didn't work.

Note that it is probably a good idea to add a new example page for the new interface.


erezsh:
Always happy to accept a patch that removes lines :)

MegaIng:
"Sadly" also adds tests, so overall still an addition.

erezsh:
I don't consider tests to be part of the code, and I'm always happy to accept more (as long as they actually test something new)

PR #762, opened 2020-11-16
:memo: Updated docs

ThatXliner:
Replaced the notice saying only `common.lark` is available into "`common.lark`, `python.lark`, and `unicode.lark`".


erezsh:
Thanks!

Issue #763, opened 2020-11-17
Can you have a Transformer or lexer_callback that replaces the output of one rule with multiple values?


rfox12-edu:
**What is your question?**

Try to be accurate and concise.

**If you're having trouble with your code or grammar**

Provide a small script that encapsulates your issue.

Explain what you're trying to do, and what is obstructing your progress.

rfox12-edu:
I have a rule that basically does:
    `number_over_number: NUMBER FORWARD_SLASH NUMBER`

Then I need to validate that this number is a fraction.  If it is "1/2" I've been able to use a Transformer with success to replace the node in the Tree.

If, however, my `number_over_number` node is "5/3" I would like to return more than a single node... i.e. "restore" the parent structure such that my output lists 
STUFF BEFORE
NUMBER
FORWARD_SLASH
NUMBER
STUFF AFTER

I hope this is clear.  In short: is there a way to achieve some validation on a rule, and if that validation cause the rule match to fail?

ThatXliner:
Like.. preprocessing? You can replace `FORWARD_SLASH` with `"/"` if you don't want it

rfox12-edu:
How could I validate that the first number is less than the second number?  I've been parsing and then post-processing...

ThatXliner:
I think you can first have a generic `fraction` rule. Then, when transforming, determine whether to change it to an `improper` or `proper` rule. Personally, I don't you should do this if you want to promote nested expressions

rfox12-edu:
Ok here is my complete example:
```
import logging
from lark import Lark, Transformer, Token, Tree, logger
from lark.visitors import Transformer_InPlace

identifier_grammar = '''
    start: token+
    token: NUMBER -> number
        | FORWARD_SLASH -> forward_slash
        | _number_over_number -> number_over_number
        | UNASSIGNED -> unassigned
    
    FORWARD_SLASH: "/"
    _number_over_number: NUMBER FORWARD_SLASH NUMBER    // TODO: consider whitespace?
    UNASSIGNED.-2: /\S+/
    
    %import common.NUMBER
    %import common.WS
    %ignore WS
'''

p = Lark(identifier_grammar, parser='lalr', debug=True)
print('case 1 -----------------')
print(p.parse("1 and 2 and /").pretty())
print('case 2 -----------------')
print(p.parse("1/2").pretty())
print('case 3 -----------------')
print(p.parse("5/2").pretty())
```
And the output is:
```
case 1 -----------------
start
[...]

PR #765, opened 2020-11-17
 Added support for returning Inline in transformers (Issue #763)

erezsh:


lark/visitors.py
```diff
@@ -113,11 +120,16 @@ def _transform_children(self, children):
...
             try:
                 if isinstance(c, Tree):
-                    yield self._transform_tree(c)
+                    res = self._transform_tree(c)
                 elif self.__visit_tokens__ and isinstance(c, Token):
-                    yield self._call_userfunc_token(c)
+                    res = self._call_userfunc_token(c)
+                else:
+                    res = c
+                if type(res) is Inline:
```
MegaIng, line 128: bad style: `isinstance` should be used.

tests/test_trees.py
```diff
@@ -233,6 +233,24 @@ def b(cls, args):
...
         self.assertEqual(x, t2)
 
+    def test_inline(self):
+        class MyTransformer(Transformer):
+            def b(self, children):
+                return 'b'
+
+            def a(self, children):
+                return Inline('c', 'd')
+
```
MegaIng, line 243: Adding a test to test for a Token would be nice just in case we somehow break that in the future.

lark/visitors.py
```diff
@@ -15,9 +15,16 @@ class Discard(Exception):
     """
     pass
 
-# Transformers
+class Inline(object):
+    """When returning an Inline instance in a transformer callback,
```
MegaIng, line 19: Question: Will this make sure that it is added to the docs?

Anyway, an example would be nice, as well as pointing out to use it `Inline(a, b, c)` and not `Inline((a,b,c))`.

And it should be added to the stubs.

MegaIng (changes requested):
Left a few comments.

MegaIng:
And this PR is certainly not complete as can be seen from comments on #763. (e.g. still missing for the `transformer` parameter. That should also be in the tests.

PR #766, opened 2020-11-18
Adding missing Testcase to `__main__`

MegaIng:
Has already been implemented for some time, but wasn't executed. It should pas both python2.7 and python3.


erezsh:
Good catch.

PR #770, opened 2020-11-20
Added `lark.lark` to the standardlib of grammars.

MegaIng:
This moves `lark.lark` into the `lark.grammars` folders. This means that it will be shiped with lark and can be used by packages that want to do something with lark grammars, like `lark-railroad` or the language server.

Since it is now in an even more official place than before, I added a check that tries to parse every grammar used inside the `_make_parser_tests`. This forces us to keep the grammar up-to-date. (as example, I already had to fix something in the grammar since we introduces multi line regex.) The cost of this check is minimal, but not zero. It adds (on my machine) ~2.5% to the total runtime.

### DIFF
--- a/examples/lark_grammar.py
+++ b/examples/lark_grammar.py
@@ -7,13 +7,13 @@ A reference implementation of the Lark grammar (using LALR(1))
 import lark
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


erezsh:
Thanks! Nice idea to automatically check the grammars of tests.

Issue #760, opened 2020-11-16
error reporting with match_examples broken with earley when lexer is dynamic


ornariece:
Trying to match error examples doesn't work when using earley with the dynamic lexer, seemingly because the parser state is not passed to the UnexpectedCharacters object. The assertion on the state not being None fails.

From what i gather, in https://github.com/lark-parser/lark/blob/0.11.0/lark/parsers/xearley.py#L116, a state should be passed to the UnexpectedCharacters. Not sure where to get the state from though, hence why no PR from me.

erezsh:
I'm sorry if you were misled by the docs. The `match_examples()` function only worked with LALR until now.

But, you inspired me to try and fix it. I pushed a new branch, `earley_error_reporting`. Can you check if it works for you?

ornariece:
it now works with earley, but it breaks with lalr for some strings (i'm using both earley and lalr, for 2 different grammars). the same assertion is raised, on the state being None.

erezsh:
Can you provide a minimal example where it fails with LALR?

ornariece:
here is a simplified grammar:
```
start: subject "." test operator value
test: CNAME
value: CNAME
CNAME: ("_"|"."|LETTER) ("_"|"."|LETTER|DIGIT)*
num: NUMBER
!subject: "self"
!operator: "=="

%import common.LETTER
%import common.DIGIT
%import common.NUMBER
%import common.WS_INLINE

%ignore WS_INLINE
```

and here are the 2 examples where it fails to match the errors (`AssertionError: Not supported for this exception`):
`self.a =< b`
`self.a == b &^ self.c == d`

ornariece:
hi, any update on this? not trying to pressure you, i'd just want to know if i should move on to a workaround

MegaIng:
`match_examples` only works for `UnexpectedToken` and `UnexpectedEOF`, not for `UnexpectedCharacters`. Make sure you use it on a exception of those types.

ornariece:
well i'm using it on ``UnexpectedInput``. previously, in 0.10.0, it wasn't throwing an ``UnexpectedCharacter`` (or if it did it didn't prevent the error from getting matched). it seems 0.11.0 introduced this behavior. is it intended?

MegaIng:
No. It doesn't work in 0.10.0. It doesn't work with the same exception for the last three years.

ornariece:
[...]

PR #772, opened 2020-11-21
Fix `match_examples` support for #760

erezsh:
- Added Earley support

- Added test-case

- Also made sure `token_history` is in UnexpectedToken

### DIFF
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -147,7 +147,7 @@ class UnexpectedToken(ParseError, UnexpectedInput):
 
     see: :ref:`ParserPuppet`.
     """
-    def __init__(self, token, expected, considered_rules=None, state=None, puppet=None):
+    def __init__(self, token, expected, considered_rules=None, state=None, puppet=None, token_history=None):
         self.line = getattr(token, 'line', '?')
         self.column = getattr(token, 'column', '?')
         self.pos_in_stream = getattr(token, 'pos_in_stream', None)
@@ -157,6 +157,7 @@ class UnexpectedToken(ParseError, UnexpectedInput):
         self.expected = expected     # XXX deprecate? `accepts` is better
         self.considered_rules = considered_rules
         self.puppet = puppet
+        self.token_history = token_history
 
         # TODO Only calculate `accepts()` when we need to display it to the user
         # This will improve performance when doing automatic error handling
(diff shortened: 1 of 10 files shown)


lark/parsers/earley.py
```diff
@@ -299,7 +299,7 @@ def parse(self, stream, start):
         solutions = [n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0]
         if not solutions:
             expected_terminals = [t.expect for t in to_scan]
-            raise UnexpectedEOF(expected_terminals)
+            raise UnexpectedEOF(expected_terminals, state={i.s for i in to_scan})
```
MegaIng, line 302: Wouldn't a `frozendict` be better?

MegaIng: Also, I trust you that this makes sense. I do not know enough about earley to judge this.

tests/test_parser.py
```diff
@@ -2342,6 +2342,26 @@ def test_lexer_detect_newline_tokens(self):
                 self.assertEqual(a.line, 1)
                 self.assertEqual(b.line, 2)
 
+        @unittest.skipIf(LEXER=='standard' and PARSER!='lalr', "Puppet error handling only works with LALR for now")
```
MegaIng, line 2345: Why is this skip here?

lark/exceptions.py
```diff
@@ -108,12 +101,25 @@ def match_examples(self, parse_fn, examples, token_type_match_fallback=False, us
...
                         except AttributeError:
                             pass
-                        if not candidate[0]:
+                        if candidate[0] is None:
                             logger.debug("Same State match at example [%s][%s]" % (i, j))
                             candidate = label, False
 
         return candidate[0]
 
+class UnexpectedEOF(ParseError, UnexpectedInput):
```
MegaIng, line 110: Are we throwing this exception for Lark?

MegaIng (changes requested):

lark/parsers/earley.py
erezsh: I don't think it matters, since no hashing is going to take place.

It's basically just a set of the next terminals to consider. It's somewhat similar to LALR state, but doesn't behave exactly the same. Neither of them are a sound and well-understood solution, they're both just a hack that should usually work.

lark/exceptions.py
erezsh: What do you mean?

erezsh:
@MegaIng Made a few changes. I think it's going in the right direction.

I think the future should have a `ParserStateInterface` the implements things like equality, `accepts`, `considered_tokens` etc. That way we'll have a unified interface.

lark/exceptions.py
MegaIng: Are we using this exception when Lalr reached the end of its input, or only earley?

MegaIng (approved):

lark/exceptions.py
erezsh: Only for Earley, right now LALR throws UnexpectedToken instead. We could change it so that both throw it on EOF, or neither one. Maybe it's even a good idea. But it seems outside the scope of this PR.

PR #776, opened 2020-11-23
Earley custom lexer implementation

MegaIng:
This is the 'easy plug'. It is probably a good idea to refactor 'parser_frontends' at some point at a large scale.

This PR also contains a few changes to the tests that were discovered while fixing this. (most notable: ('earley', 'dynamic_complete') wasn't run). It for that guarantees that all auto generated tests will be run.

### DIFF
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -28,10 +28,7 @@ def get_frontend(parser, lexer):
                     self.lexer = lexer(lexer_conf)
                 def lex(self, lexer_state, parser_state):
                     return self.lexer.lex(lexer_state.text)
-
             class LALR_CustomLexerWrapper(LALR_WithLexer):
-                def __init__(self, lexer_conf, parser_conf, options=None):
-                    super(LALR_CustomLexerWrapper, self).__init__(lexer_conf, parser_conf, options=options)
                 def init_lexer(self):
                     future_interface = getattr(lexer, '__future_interface__', False)
                     if future_interface:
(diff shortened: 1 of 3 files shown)


erezsh:
Nice one!

It was relatively easy, you just made it a little harder by going the extra mile :)

> It is probably a good idea to refactor 'parser_frontends' at some point at a large scale

Sure, maybe we can do that soon.

Do you have any ideas on how to approach it?

MegaIng:
Reduce the different behaviour between the different lexers. We don't need to differntiate between different frontends for the traditional and custom lexer. All we have to do is create a factory function that returns a class (like `wrap_lexer`, but for a frontend). If we also add another attribute similar to `__future_interface__`, or pass a function that generates the states, we can even allow custom contextual lexer. Then we don't need different subclasses.

erezsh:
I agree that custom lexer should have the same interface as traditional lexer. I think that's how it used to be before all the changes.

Now that the parser state is always passed, then yes, the contextual lexer interface isn't unique anymore. Maybe it can be folded in.

I wonder if it can be done without breaking compatibility, or is it better to do it as part of a v1.0 effort.

MegaIng:
We could add a 'temporary' attribute 'states' to lexer state, e.g. that is set inside of the FrontEnd init.

erezsh:
Didn't quite get that

MegaIng:
Sorry, typo. meant the `LexerConf`, e.g. we add a `LexerConf.state` attribute that will be set inside of `WithLexer.__init__`. This can then be used by `Lexer.__init__`.

erezsh:
That sounds a little hacky.. But we'll see. I think the first step should be to decide on the interface. Ideally there should be only 3 entities involved in this: 

- Parsers

- Lexers

- A single managing class `ParserFrontend` (or whatever name) that accepts a `ParserConf`, a `LexerConf`, and  extra options, and provides a single interface to the `Lark` class.

The managing class should have no knowledge of the specific parsers and the lexers, outside of the interface that they provide. All it will do is set them up, so that the parser gets the lexer instance, as it is, or wrapped by a postlexer (there will also be a no-lexer mode for earley+dynamic)

The parser will talk directly to the lexer (or wrapper) and they will exchange between them the `ParserState` and `LexerState`, perhaps wrapped by a single `State` instance.

Does that all sound right so far?

MegaIng:
Sounds good. Currently, this is managed via having classes for each combination. This is sub optimal in my opinion.

PR #777, opened 2020-11-24
Made trees iterable (and indexable)

ThatXliner:
Because I don't want to keep on doing `tree.children`.

See https://gitter.im/lark-parser/Lobby?at=5fbc8970e6f2b51c68b28a22

Without this, I'm doing things like

```python
    def declare(self, tree):
        return "%declare " + " ".join(
            map(
                operator.itemgetter(0),
                map(operator.attrgetter("children"), tree),
            )
        )
```
and 
```python
def single_import(self, tree):
    (tree,) = tree
    tree = tree.children
    return "%import " + str(tree[0].children[0]) + "." + str(tree[2].children[0])
```

### DIFF
--- a/lark-stubs/tree.pyi
+++ b/lark-stubs/tree.pyi
@@ -58,15 +58,15 @@ class Tree:
     def __hash__(self) -> int:
         ...
 
+    def __iter__(self) -> Iterator[Union[str, Tree]]:
+        ...
+
 
 class SlottedTree(Tree):
     pass
 
 
 def pydot__tree_to_png(
-    tree: Tree,
-    filename: str,
-    rankdir: Literal["TB", "LR", "BT", "RL"] = ...,
-    **kwargs
+    tree: Tree, filename: str, rankdir: Literal["TB", "LR", "BT", "RL"] = ..., **kwargs
 ) -> None:
     ...
--- a/lark-stubs/tree.pyi
+++ b/lark-stubs/tree.pyi
@@ -61,6 +61,9 @@ class Tree:
     def __iter__(self) -> Iterator[Union[str, Tree]]:
         ...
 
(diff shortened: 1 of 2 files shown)


ThatXliner:
I know that I haven't updated the type hints.

Quick question: Should I do `return iter(self.children)` instead? Would it be cleaner?

ThatXliner:
Should I also add a `__getitem__` attribute for becoming indexable? (i.e. being able to do "`tree[some_index]`")

erezsh:
That's great. Now you can pass your tree class to Lark. No PR necessary.
```python
>>> class MyTree:
...     def __init__(self, data, children)::::
...             self.data = data
...             self.children = children
... 
>>> 
>>> from lark import Lark
>>> l = Lark('start: "a"', tree_class=MyTree).parse('a')
>>> l
<__main__.MyTree object at 0x0000026CDC8E00A0>
```

ThatXliner:
Oh cool. But I wonder if others would like this feature added, too.

ThatXliner:
Also, I get

```traceback
AttributeError: 'IndexableTree' object has no attribute '_meta'
```

MegaIng:
Yeah, you should subclass `Tree`.

ThatXliner:
> Yeah, you should subclass `Tree`.

I did it. I fixed it by passing in meta as a param
```py
def __init__(self, data, children, meta=None):
        self.data = data
        self.children = children
        self._meta = meta
```

MegaIng:
You don't need to define `__init__`.

erezsh:
@ThatXliner If you are inheriting from `Tree`, you should call `Tree.__init__`. 

But other than that, yes Lark expects the `meta` attribute.

ThatXliner:
`'Tree' object does not support indexing`

PR #781, opened 2020-11-25
Refactored parser_frontends. Now significantly simpler

erezsh:
It's not perfect yet, but I think it's a big step in the right direction.

### DIFF
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -4,9 +4,8 @@ from .parsers.grammar_analysis import GrammarAnalyzer
 from .lexer import LexerThread, TraditionalLexer, ContextualLexer, Lexer, Token, TerminalDef
 from .parsers import earley, xearley, cyk
 from .parsers.lalr_parser import LALR_Parser
-from .grammar import Rule
 from .tree import Tree
-from .common import LexerConf
+from .common import LexerConf, ParserConf
 try:
     import regex
 except ImportError:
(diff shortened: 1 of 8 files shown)


lark/parser_frontends.py
```diff
@@ -27,56 +26,112 @@ def lex(self, lexer_state, parser_state):
...
-        elif lexer == 'contextual':
-            return LALR_ContextualLexer
-        elif issubclass(lexer, Lexer):
-            wrapped = _wrap_lexer(lexer)
-            class LALR_CustomLexerWrapper(LALR_WithLexer):
-                def init_lexer(self):
-                    self.lexer = wrapped(self.lexer_conf)
-            return LALR_CustomLexerWrapper
-        else:
+        if lexer not in ('standard' ,'contextual') and not issubclass(lexer, Lexer):
```
MegaIng, line 123: If `dynamic` gets here in `lexer`, this will throw TypeError. I would suggest splitting this up into multiple if statements, or using `isinstance(lexer, type)`. In theory, `LexerOptions` already checks if it is a subclass.

lark/parser_frontends.py
```diff
@@ -27,56 +26,112 @@ def lex(self, lexer_state, parser_state):
...
-                    self.lexer = wrapped(self.lexer_conf)
-            return Earley_CustomLexerWrapper
-        else:
-            raise ConfigurationError('Unknown lexer: %s' % lexer)
     elif parser == 'cyk':
-        if lexer == 'standard':
-            return CYK
-        else:
+        if lexer != 'standard':
             raise ConfigurationError('CYK parser requires using standard parser.')
```
MegaIng, line 130: Could Cyk be made to support Custom Lexer? (I mean, I don't know why, but it should work, shouldn't it)

MegaIng (changes requested):
This is approximately what I imagined. Now we just need to remove the distinction between Standard and Contextual. (but not in this PR).

(Also, stubs? Do we need to update something there?)

lark/parser_frontends.py
```diff
@@ -100,119 +155,26 @@ def lex(self, lexer_state, parser_state):
...
-    def init_lexer(self):
-        self.init_traditional_lexer()
-
-class LALR_ContextualLexer(LALR_WithLexer):
-    def init_lexer(self):
-        states = {idx:list(t.keys()) for idx, t in self.parser._parse_table.states.items()}
-        always_accept = self.postlex.always_accept if self.postlex else ()
-        self.lexer = ContextualLexer(self.lexer_conf, states, always_accept=always_accept)
 
+make_early = NotImplemented
```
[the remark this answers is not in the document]
erezsh: Yeah, I'll change it to `create_earley`.

lark/parser_frontends.py
erezsh: Yeah, it should work. Maybe it already works and we just need to remove this check.

erezsh: Yep. Easy.

lark/parser_frontends.py
erezsh: Done

lark/parser_frontends.py
```diff
@@ -230,31 +187,49 @@ def _prepare_match(self, lexer_conf):
...
         self.parser = cyk.Parser(parser_conf.rules)
 
         self.callbacks = parser_conf.callbacks
 
-    def parse(self, text, start):
-        tokens = list(self.make_lexer(text).lex(None))
-        parse = self._parse(start, tokens)
-        parse = self._transform(parse)
-        return parse
+    def parse(self, lexer, start):
```
MegaIng, line 229: I picked this line as an example: I would suggest clearly making a distinction between variables that contain a Lexer, LexerThread, or LexerState. This function confused me, since `lexer.lex(None)` looks like we are trying to get the Token that get generated from Nothing.

MegaIng (changes requested):
Scrolling through a PR with this many red lines is quite satisfying. Especially if you know that is also adds new features :-D .

lark/parser_frontends.py
erezsh: I can rename that specific variable to `lexer_thread`.

This doesn't happen anywhere else in the code, only CYK needs special handling.

Other than that, I think everything is named correctly, no?

erezsh: Actually, right now I'm setting `LexerConf.name`, but it could also be a CustomLexer. Any idea for a better name for this attribute?

lark/parser_frontends.py
```diff
@@ -27,57 +26,108 @@ def lex(self, lexer_state, parser_state):
...
             if len(start) > 1:
                 raise ConfigurationError("Lark initialized with more than 1 possible start rule. Must specify which start rule to parse", start)
             start ,= start
         return self.parser.parse(input, start, *args)
 
+    def parse(self, text, start=None):
+        if self.skip_lexer:
+            return self._parse(start, text)
+
+        lexer = LexerThread(self.lexer, text)
```
MegaIng, line 115: Here. Also, other files outside of this commit also had this problem.

lark/parser_frontends.py
MegaIng: Maybe `lexer_type`?

lark/parser_frontends.py
erezsh: Hmm okay.. I mean, you see the init right there. But I guess it can't hurt to be more explicit.

MegaIng: This is in general a 'problem' in this library: many things that could be justified into the same variable name:

`parser`:

- `Lark`
- `ParsingFrontend`
- `LALR_Parser`/`Eearley`/`XEealey`/`CYK` instances
- `LALR_Parser`/`Eearley`/`XEealey`/`CYK` classes
- the strings specifying  a parser

`lexer`

- `Lexer` class
- `Lexer` instance
- `LexerState` instance (but I don't think this occurs anywhere)
- `LexerThread`
- the strings specifying a lexer

`options`/`config`

- `LarkOptions` instance
- `LexerConfig`
- `ParserConfig`

while it is almost always clear from context which of these is meant, it doesn't hurt to be specific.

MegaIng: Btw, I am not saying 'rename all variables  that might be confusing'. Just be aware of especially problematic sections.

as example, it should clear if we are looking at an implementation for `parser.parse`, `parser.parser.parse` or `parser.parser.parser.parse` if there is a difference in what we are supposed to give them (e.g. `text` vs `lexer_thread`).

lark/parser_frontends.py
```diff
@@ -76,44 +76,44 @@ def __init__(self, lexer_conf, parser_conf, options, parser=None):
                 'lalr': create_lalr_parser,
                 'earley': create_earley_parser,
                 'cyk': CYK_FrontEnd,
-            }[parser_conf.name]
+            }[parser_conf.parser_type]
             self.parser = create_parser(lexer_conf, parser_conf, options)
 
         # Set-up lexer
+        lexer_type = lexer_conf.lexer_type
+        lexer_type = lexer_conf.lexer_type
```
MegaIng, line 84: ?

MegaIng (approved):

lark/parser_frontends.py
erezsh: Thanks for catching it

PR #785, opened 2020-11-27
Adding Power support(ppc64le) with ci and testing to the project for architecture independent

asellappen:
I am part of IBM team and working on porting power arch to packages as continuous integration & testing.

This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing,We typically build applications for customers and ISVs, and while we don't use this package directly,

we do count on all of the packages in debian/ubuntu to build other packages. So we more likely have this as a second or third level dependency and couldn't tell you explicitly which features we use or our usage model,For more info tag @gerrith3.

Pls verify and merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 dist: xenial
+arch:
+  - amd64
+  - ppc64le
 language: python
 python:
   - "2.7"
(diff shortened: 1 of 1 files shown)


MegaIng:
This is unnecessary, since this package has zero platform depended parts. We are acting completely within the python specs, so there is no reason to test on different platforms.

erezsh:
What @MegaIng said.

If Lark doesn't work on a specific cpu architecture, it's a bug in Python, not Lark.

erezsh:
Closed for lack of response

PR #786, opened 2020-11-29
Reconstruct with unicode terminals (Based on #748)

erezsh:
@julienmalard Your test is failing. Can you tell me why?

I don't think I changed anything substantial.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -7,7 +7,7 @@ from io import open
 import pkgutil
 from ast import literal_eval
 
-from .utils import bfs, Py36, logger, classify_bool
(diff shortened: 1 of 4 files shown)


MegaIng:
For me it seems like the bug is simply not fixed. Maybe something in master broke it again (the behavior exactly matches the reported bug in the test comment.)

MegaIng:
@erezsh I think you accidentally reverted the change at line 337 in file `load_grammar.py`.

erezsh:
You're right. I removed the check for unicode in terminal names. I guess it wasn't really necessary in the first place.

julienmalard:
@erezsh I think @MegaIng is right, and that the latest commit should do the trick!
Thanks, 
Julien

PR #790, opened 2020-12-03
type stubs: A consistent type for `Token.value`

unbrice:
The Token's value is described as being a in the attribute list: https://github.com/lark-parser/lark/blob/460a221923317299422f14d87901dee1eb5f5fda/lark-stubs/lexer.pyi#L84 and as `Any` in `__init__` parameters:  https://github.com/lark-parser/lark/blob/460a221923317299422f14d87901dee1eb5f5fda/lark-stubs/lexer.pyi#L91

One of the recipes, [Use a transformer to parse integer tokens](https://github.com/lark-parser/lark/blob/master/docs/recipes.md#use-a-transformer-to-parse-integer-tokens), recommends to set it to other types. For that recipe to type-check, `Any` is the correct type.

### DIFF
--- a/lark-stubs/lexer.pyi
+++ b/lark-stubs/lexer.pyi
@@ -81,7 +81,7 @@ class TerminalDef:
 class Token(str):
     type: str
     pos_in_stream: int
-    value: str
+    value: Any
     line: int
     column: int
     end_line: int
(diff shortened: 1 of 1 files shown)


MegaIng (approved):

erezsh:
Thanks! It's usually `str`, but you're right that it doesn't have to be.

PR #794, opened 2020-12-06
Prevent unreadable exceptions.

MegaIng:
we said 't_name should always be in `terminals_by_name`'. Well, isn't the case when the expected set contains wither `$END` or a `%declared` terminal.


erezsh:
That's a very indirect way of saying that your pull request inserted a bug to the code base.

Anyway, I already fixed it (just pushed now)

PR #800, opened 2020-12-20
Add grammar for ISO 8601 (datetime)

aatifsyed:
Hi Lark maintainers!

This is my first PR to an open source project.  
It offers a grammar that covers some of ISO 8601.  
I'd like to add some tests with your guidance, and polish this PR until it's of acceptable quality :)


MegaIng:
This is probably better as a single Terminal.

aatifsyed:
Ah, I see now 
```
lark.exceptions.GrammarError: Rules aren't allowed inside terminals (prefixed_week_number in WEEK_DATE)
```
Why is that?
Also, what is the example that you were alluding to?

aatifsyed:
I'm getting an error that I can't quite wrap my head around:
```
lark.exceptions.GrammarError: Using an undefined rule: NonTerminal('start')
```

tests/test_datetime_iso_8601.py
```diff
@@ -0,0 +1,28 @@
+from pathlib import Path
+
+import pytest
+from lark import Lark
+
+
+@pytest.fixture
```
MegaIng, line 7: We aren't using pytest. Please use `unittest`.

.gitignore
```diff
@@ -1,14 +1,20 @@
...
-.vscode
+
+# Editors and tools
 .idea
-.ropeproject
-.cache
 .mypy_cache
-/dist
-/build
-docs/_build
```
MegaIng, line 13: Why are you changing the order here?

MegaIng (changes requested):

.gitignore
aatifsyed: I just alphabetised and grouped here. Not particularly partial

erezsh:
> Rules aren't allowed inside terminals

Terminals, by definition, are the smallest element in the grammar. They are merely a string of characters. They cannot contain rules, which are the grammar equivalent of functions, in the same way that a Python string cannot contain a list (but a list can contain a string).

aatifsyed:
Hm, I'm still having trouble with this

> I'm getting an error that I can't quite wrap my head around:
> 
> ```
> lark.exceptions.GrammarError: Using an undefined rule: NonTerminal('start')
> ```

What am I missing?

aatifsyed:
Hm, I think I'll need to level up my `lark`. I'm erroring on just the following, emulating `common.lark`, and I'm not sure why
```
%import common.DIGIT

YEAR  : DIGIT DIGIT DIGIT DIGIT

MONTH : "1".."9"
```
```
lark.exceptions.GrammarError: Using an undefined rule: NonTerminal('start')
```

aatifsyed:
Alright, I've found my feet on this again, and actually been able to use it, after getting more familiar with Lark :)
A few questions for the maintainers:
- Where should tests go? Their own file? Or is there somewhere they already fit, as @MegaIng alluded to?
- Should everything really be a terminal? It might be useful to users to e.g pull out the `TIME` token from a `DATETIME`?

ThatXliner:
Heyo 👋 what's the status of this PR? A date could be parsed via regex

erezsh:
Sorry, I didn't notice this went unanswered.

> Where should tests go? 

Probably a new file, `test_grammar_lib.py` or something like that.

> Should everything really be a terminal? It might be useful to users to e.g pull out the TIME token from a DATETIME?

That would still be possible with terminals. The user could do `%import date_iso (TIME)`.

PR #801, opened 2020-12-23
Grammar: Added new %override statement for rewriting imported rules (like in class inheritance)

erezsh:
Works for templates too, of course.

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -289,3 +289,16 @@ Note that `%ignore` directives cannot be imported. Imported rules will abide by 
 
 Declare a terminal without defining it. Useful for plugins.
 
+### %override
+
+Override a rule, affecting all the rules that refer to it.
+
+Useful for implementing an inheritance pattern when importing grammars.
+
+**Example:**
+```perl
+%import my_grammar (start, number, NUMBER)
+
+// Add hex support to my_grammar
+%override number: NUMBER | /0x\w+/
+```
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -1015,8 +1015,10 @@ class GrammarLoader:
             term_defs += new_td
             rule_defs += new_rd
 
+        # replace rules by overridding rules, according to name
         for r in overriding_rules:
             name = r[0]
+            # remove overridden rule from rule_defs
             overridden, rule_defs = classify_bool(rule_defs, lambda r: r[0] == name)    # FIXME inefficient
             if not overridden:
                 raise GrammarError("Cannot override a nonexisting rule: %s" % name)
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -93,6 +93,7 @@ TERMINALS = {
     'COMMENT': r'\s*//[^\n]*',
     '_TO': '->',
     '_IGNORE': r'%ignore',
+    '_OVERRIDE': r'%override',
     '_DECLARE': r'%declare',
     '_IMPORT': r'%import',
     'NUMBER': r'[+-]?\d+',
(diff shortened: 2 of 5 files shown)


lark/load_grammar.py
```diff
@@ -1045,10 +1060,10 @@ def load_grammar(self, grammar_text, grammar_name='<?>', import_paths=[]):
 
         resolve_term_references(term_defs)
 
-        rules = rule_defs
+        ## Handle rules
```
MegaIng, line 1063: A very vague comment

lark/load_grammar.py
```diff
@@ -1009,6 +1015,15 @@ def load_grammar(self, grammar_text, grammar_name='<?>', import_paths=[]):
             term_defs += new_td
             rule_defs += new_rd
 
+        for r in overriding_rules:
+            name = r[0]
+            overridden, rule_defs = classify_bool(rule_defs, lambda r: r[0] == name)    # FIXME inefficient
```
MegaIng, line 1020: I think a small comment explaining what is happening here would help. It took me a moment to realize that this throws away the `overriden` rules.

Also, can't we do the check at a later point when we are iterating over `rule_defs` anyway? But that would probably be less clear so I am not sure that it is a good idea.

MegaIng (commented):
Good grieve, this is on something I didn't think about till now.

Questions: What does this do with nested imports? Can this also work for terminals? Can we create an `%extends` statement that allows one to split the definition of a single rule over multiple files?

MegaIng:
Also, examples + docs are missing.

lark/load_grammar.py
erezsh: It just declares what that section does, similar to a function name

erezsh:
> Questions: What does this do with nested imports? Can this also work for terminals? Can we create an `%extends` statement that allows one to split the definition of a single rule over multiple files?

Nested imports should work as usual? As long as you can import a rule, you can override it.

I think we can also implement it for terminals, it shouldn't be hard.

I was also thinking of `%extends`! Yes, it should be possible, and possibly desirable.

But all of that can happen in separate pull requests, I think.

lark/load_grammar.py
erezsh: Maybe there's a better way to do this, but I think it would require changing a lot of how that function operates. And it's probably best to separate new features and big refactors, if possible.

erezsh: Anyway, added comments.

erezsh:
Added documentation too.

I think examples can wait until we think of a good example showcasing this. Maybe extend the python grammar to support extra syntax.

docs/grammar.md
```diff
@@ -289,3 +289,16 @@ Note that `%ignore` directives cannot be imported. Imported rules will abide by
...
 Declare a terminal without defining it. Useful for plugins.
 
+### %override
+
+Override a rule, affecting all the rules that refer to it.
+
+Useful for implementing an inheritance pattern when importing grammars.
+
+**Example:**
+```perl
```
MegaIng, line 299: Why `perl`? Shouldn't `lark` work?

MegaIng (approved):

lark/load_grammar.py
MegaIng: > And it's probably best to separate new features and big refactors, if possible.

Agreed. I wasn't even sure how to do this correctly. I was just saying that it took me a moment that this was the place where the old definition is removed.

docs/grammar.md
erezsh: Good point! But vscode (or other editors) still can't highlight lark inside markdown

erezsh: But it's worth considering. In that case we'll have to change it all over the documentation.

PR #802, opened 2020-12-24
Add link to the repo with real-life Lark grammars

ligurio:
Spoiler: I'm the owner of that repo

I believe it would be useful for Lark users.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -79,6 +79,7 @@ Resources
 -  :doc:`philosophy`
 -  :doc:`features`
 -  `Examples`_
+-  `Third-party examples`_
 -  `Online IDE`_
 -  Tutorials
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,7 +83,7 @@ Lark is great at handling ambiguity. Here is the result of parsing the phrase "f
 
 ![fruitflies.png](examples/fruitflies.png)
 
-[Read the code here](https://github.com/lark-parser/lark/tree/master/examples/fruitflies.py), and [more examples here](https://github.com/lark-parser/lark/tree/master/examples)
+[Read the code here](https://github.com/lark-parser/lark/tree/master/examples/fruitflies.py), and [more examples here](https://github.com/lark-parser/lark/tree/master/examples) and [here](https://github.com/ligurio/lark-grammars).
```
MegaIng, line 86: I am not sure if this is the best place to add this, since the other links a complete python programs, where as this is more a collection of grammars.

ligurio: What is a better place for it?

MegaIng: I am not sure. @erezsh Opinion on this would also be appreciated.

- Further down in the list of projects.
- An additional `### Header` above/below the projects marking this as a library of grammars.

erezsh: Considering that the grammars have no tests, and some of them aren't exactly idiomatic, I'm reluctant to allow them into the README.

We can place a link to them in the docs, for example [in this page](https://lark-parser.readthedocs.io/en/latest), and just make it clear that they are external examples.

ligurio: Moved a link to a list with resources in doc index with name "Third-party examples".

erezsh:
Thanks!

PR #804, opened 2020-12-25
Added Terminal support for `%override`

MegaIng:
First point in #803

Issue #803, opened 2020-12-24
Improve support for grammar inheritance


erezsh:
- Implement `%override` for terminals
- Implement `%extend` for both rules and terminals
- Fix `%declare` to work for rules (to allow abstract grammars)

@MegaIng if you feel like it, you can implement these. Otherwise I'll do it at some point.

MegaIng:
Would you opposed to add these points:

[ ] refactor of `GrammarLoader.load_grammar`
[ ] adding a `GrammarBuilder` class, allowing for a gradual building of a `Grammar` object, to be used when creating plugin system.

I have a few ideas for the `GrammarBuilder`. (It could also just the `GrammarLoader` class with a bit of refactor work put into it.)

Here a few methods I would expect:

- `add_grammar`, which takes a source code and adds it in it entirety.
  - This should also have an parameter `mangle_except` or similar, which would do what is currently done to the names of import rules and terminals
- `import_grammar`, which takes the name (or path) of a grammar file (also with `mangle_except`)
- `define`/`override`/`extend`/`declare`/`ignore`, allowing one to define manually add entries to the grammar
  - I am not sure in what format `define` etc. should take their arguments
- `check` checks whether or not the currently added content is a legal grammar (same checks as done at the end of `load_grammar` at the moment). 
  - This implies that the other methods allow one to create a grammar in an illegal state without throwing (e.g. undefined rules/terminals)
- `build`, or `compile`, which would create either
  - a `Grammar` object, which would then be passable to `Lark`, meaing we would have to change the interface there
  - directly a `Lark` instance
- `to_source`, creating a source representation that should be at least somewhat humanly readable.

MegaIng:
Also, is there a particular reason why `expansions` is handle as a recursive structure and not inlined while parsing like `expansion` is?

erezsh:
> refactor of GrammarLoader.load_grammar

Yes, that's probably a good idea, especially with the recent additions

> adding a GrammarBuilder class

How do you imagine it helping the creation of a plugin system? Will they be able to interject at the different stages of building the grammar? That would make any future internal refactor backwards incompatible.

I have a few more questions:

> mangle_except

What does that mean?

> define

What does that do? Why is it needed?

Other than the separate `check` and addition of `to_source` (which should work with the current implementation), it seems like a very similar design to the current one.

> Also, is there a particular reason why expansions is handle as a recursive structure

Not sure I understand the question.

erezsh:
[...]

PR #805, opened 2020-12-25
Added `%extend`

MegaIng:
Second point in #803.

Here it becomes very apparent that the current approach creates a lot of duplicate code. A complete refactor of the `GrammarLoader` class is in my opinion a good idea. (Especially breaking up `.load_grammar` to allow them to be reused in other parts more easier).

Based on #804.

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -302,3 +302,22 @@ Useful for implementing an inheritance pattern when importing grammars.
 // Add hex support to my_grammar
 %override number: NUMBER | /0x\w+/
 ```
+
+### %extend
+
+Extend the definition of a rule or terminal, e.g. add a new option on what it can match, like when separated with `|`.
+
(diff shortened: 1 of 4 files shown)


erezsh:
Yes, I agree that a refactor would improve the implementation by a lot, by this point. I will hold off on a review for now.

One thing to keep in mind is that for every name there should be only one rule (or terminal). So it makes sense to store them in a dict rather than a list, which will make override/extend a lot faster and idiomatic.

The reason it was kept as a list so far is to validate that there are no duplicates. But that can be done earlier instead of later on.

erezsh:
And of course, it probably makes sense to unite the handling of rules and terminals. They can even be stored in the same dictionary, and treated differently in the few cases necessary, based on `tree.data`, or some other flag.

erezsh:
Outdated

PR #808, opened 2020-12-27
Grammar builder

MegaIng:
This is in the direction I want to imagine this interface to look like.

The first push only contains code that does the same as the old code, not defining a new API (it also still contains `GrammarLoader` to easily check for changed behavior. But the entire class can be removed before this PR is finished)

(Maybe you can tell why the tests are failing?)

### DIFF
--- a/lark-stubs/load_grammar.pyi
+++ b/lark-stubs/load_grammar.pyi
@@ -1,0 +1,28 @@
+from typing import List, Tuple, Union, Callable, Dict, Optional
+
+from lark import Tree
+from lark.grammar import RuleOptions
+
+
+class Grammar:
+    rule_defs: List[Tuple[str, Tuple[str, ...], Tree, RuleOptions]]
+    term_defs: List[Tuple[str, Tuple[Tree, int]]]
+    ignore: List[str]
+
+
+class GrammarBuilder:
+    global_keep_all_tokens: bool
+    import_paths: List[Union[str, Callable]]
+
+    def __init__(self, global_keep_all_tokens=..., import_paths=...): ...
+
+    def load_grammar(self, grammar_text: str, grammar_name: str = ..., mangle: Callable[[str], str] = None): ...
+
+    def do_import(self, dotted_path: Tuple[str, ...], base_path: Optional[str], aliases: Dict[str, str],
+                  base_mangle: Callable[[str], str] = None):  ...
+
+    def get_mangle(self, prefix: str, aliases: Dict[str, str], base_mangle: Callable[[str], str] = None): ...
+
+    def check(self): ...
+
+    def build(self) -> Grammar: ...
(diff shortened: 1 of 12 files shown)


MegaIng:
@erezsh Ok, While this PR is not finished, I think that the core is finished.

The big things missing at the moment:

- docs: Where should I put this?
- names: I am not that happy with the names of the functions. Do you have better ideas?
- cleanup: Are there things that shouldn't be in this PR/debug stuff I overlooked?
- example: Is the example I added ok like it is?

lark-stubs/load_grammar.pyi
```diff
@@ -0,0 +1,28 @@
+from typing import List, Tuple, Union, Callable, Dict, Optional
+
+from lark import Tree
+from lark.grammar import RuleOptions
+
+
+class Grammar:
+    rule_defs: List[Tuple[str, Tuple[str, ...], Tree, RuleOptions]]
```
erezsh, line 8: If it's going to be an official interface, it might worth it to create classes for `RuleDef` and `TokenDef`.

It might also make `load_grammar` a bit cleaner.

MegaIng: (official interface in the sense that it might be interesting to look at these attributes. Modifying them directly is probably still to be discouraged)

I agree, but we once again run into naming issues since `TerminalDef` is already an existing class. And I would probably just make them named tuples, so that the names are readable.

tests/test_parser.py
```diff
@@ -1961,7 +1961,7 @@ def test_import_custom_sources(self):
 
             p = _Lark(grammar, import_paths=[custom_loader])
             self.assertEqual(p.parse('ab'),
-                             Tree('start', [Tree('startab', [Tree('ab__expr', [Token('ab__A', 'a'), Token('ab__B', 'b')])])]))
+                             Tree('start', [Tree('startab', [Tree('ab__expr', [Token('AB__A', 'a'), Token('AB__B', 'b')])])]))
```
MegaIng, line 1964: Btw, this is also a change I made. Should this stay changed, or should this be reverted? My goal here was to make the mangled names valid terminal names.

erezsh: I think it's better to preserve the case of the module name. It will be easier for debugging.

The letter-case of symbols is only used as part of the grammar language, and isn't a necessity for any other part of Lark.

Also, we might change that in the future, so that terminals are *any* symbol with an upper-case letter, to allow `TerminalsLikeThis`.

lark-stubs/load_grammar.pyi
erezsh: Well, there are a few ways to resolve the naming situation. Like prefix it with `Internal` or rename `TerminalDef`. Just ask yourself, what is the main difference between `lexer.TerminalDef` to the one in load_grammar.

tests/test_parser.py
MegaIng: The problem with this is that it is then `term.isupper()` will no longer hold true. This can however be fixed by just looking at the part after the last `__`.

Issue #807, opened 2020-12-27
Adding a priority to a rule causes parser to run without termination


michelleful:
**Describe the bug**

I am writing a grammar to parse a few Classical Chinese sentences as I work through a book on the subject. The grammar is highly ambiguous (nouns can act as verbs, etc). I added a rule (below, `verbal_sentence_no_subject`) that is necessary but resulted in a lot of the default parses opting to use that rule incorrectly. Therefore, I tried to increase the priority of a competing rule so that `verbal_sentence_no_subject` would not be picked as often. However, upon adding the priority setting, the script, which previously took 0.106s to run, now continues running for over an hour without terminating.

**To Reproduce**

I apologise that this is not short but I think the amount of ambiguity is necessary to reproduce the behaviour. I don't expect anyone to actually work through this grammar to figure out what's going on, I'm just wondering (a) whether it's expected that adding a priority can cause this behaviour, (b) whether I did something wrong with the priority setting, or (c) if there is another way to approach this. Basically, I want the `verbal_sentence_no_subject` rule to only kick in when a parse is impossible without it.

```
from lark import Lark

grammar = """
    sentence: sentence*
            | verbal_sentence
            | verbal_sentence_no_subject
            | nominal_sentence

    verbal_sentence.1: noun verb  # <------- Adding the positive priority to this rule
                                  #          causes the issue. Without it, the script
                                  #          runs very fast (but with what are, from 
                                  #          my perspective, sub-optimal parses)
                                  #          remove this comment to run the code.
    verbal_sentence_no_subject: verb
    nominal_sentence: noun noun nominal_particle

    noun: PROPER_NOUN | NOUN | stative_verb
    verb: adverb? VERB direct_object? indirect_object?
        | negative? stative_verb
    stative_verb: adverb? STATIVE_VERB | noun
    direct_object: noun
    indirect_object: preposition noun

    negative: NEGATIVE
    adverb: ADVERB
    nominal_particle: NOM_PRT
    preposition: PREPOSITION
    
    PROPER_NOUN: "子" | "孔子" | "齊景公" | "公" | "樊遲"
    NOUN: "性" | "習" | "政" | "君" | "臣" | "父" | "子" | "仁" | "人"
    ADVERB: "相" | "對"
    STATIVE_VERB: "近" | "遠"
    VERB: "曰" | "問" | "愛" | "知"
    NEGATIVE: "不"
    NOM_PRT: "也"
    PREPOSITION: "於"
"""

[...]

PR #809, opened 2020-12-28
Enable faster forest visiting

chanicpanic:
Related to #807.

As discussed, I initially tried extracting cycle handling to `ForestVisitor`. Unfortunately, while this did reduce the execution time, the prioritizer forest walk still took several times longer than the forest transformation itself. So, I found the root of the problem: `ForestVisitor` traverses all paths in the SPPF and as a consequence can visit the same node several times. However, `ForestSumVisitor` only needs to visit each node once. Therefore, I added an option to `ForestVisitor` to prevent it from revisiting nodes. Using this option on `ForestSumVisitor` and `ForestToPyDotVisitor` greatly improves their performance on SPPFs with many shared nodes.


erezsh:
Whoops, this one slipped through the cracks.

Thanks for the PR!

PR #810, opened 2021-01-04
Load pyodide from JsDelivr in the IDE app

rth:
Following https://pyodide.readthedocs.io/en/latest/changelog.html#version-0-16-1, it's preferable to use JsDelivr instead of the previous CDN for pyodide.  The pyodide version can likely be updated separately.


erezsh:
Now using a new IDE. New code for loading pyodide is here: https://github.com/lark-parser/ide/blob/master/src/python.js#L7

PR #811, opened 2021-01-09
Added example  python_reconstruct.py + Corrected python3.lark

MegaIng:
This shows how one might handle a postlexer (more specifically the Indenter) while reconstructing. It also updates the stubs.

The changes to `python3.lark` are to make it actually work. I am not sure if these changes are complete, but without them, the recosntructed output has syntatic differences:

- `def test(a, b): ` would become `def test(a,b,/,)`
- `{a:b, c:d, e:f}` might become `{c:b,**c,**d,e:f,}`
- `for a in range():` might become `for a, in range():`

There might be more special cases, but the worst I think is fixed

### DIFF
--- a/examples/advanced/python3.lark
+++ b/examples/advanced/python3.lark
@@ -134,7 +134,7 @@ AWAIT: "await"
 tuplelist_comp: (test|star_expr) (comp_for | ("," (test|star_expr))+ [","] | ",")
 ?subscriptlist: subscript
               | subscript (("," subscript)+ [","] | ",") -> subscript_tuple
-subscript: test | [test] ":" [test] [sliceop]
+subscript: test | ([test] ":" [test] [sliceop]) -> slice
 sliceop: ":" [test]
 exprlist: (expr|star_expr)
         | (expr|star_expr) (("," (expr|star_expr))+ [","]|",") -> exprlist_tuple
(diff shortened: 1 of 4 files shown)


erezsh:
Very nice!

There is a bug you would have noticed if you had tried to run the output.

`actions[1:]` turns into `actions[1]`

MegaIng:
Yeah, I should have attempted to run it. But I assume we will find more and more bugs of this kind while this grammar is in used. As you may have noticed, my recent interested in `GrammarLoader` and `Reconstructur` is fueled by https://github.com/MegaIng/syntax-extensions-base/ , which is a project that uses lark to exend the python grammar. I just finished the first proof of concept non-trivial extension: https://github.com/MegaIng/syntax-extensions-pep634/ . (well, 'finished').

Btw, what is your current stand on #808?

erezsh:
I merged it, but github doesn't recognize it for some reason.

erezsh:
Re #808, I don't stand anywhere yet, because I haven't read it deeply. It's a huge diff of code, in a part of the codebase that isn't well tested. So I just need to clear some time to really look into it.

Re syntax-extensions-base, that's pretty cool! I also wanted to write something like that, but never got around to it. Are you open to adding a syntax for code-blocks? How about shorter lambdas?

I mean something like:

```python
@codeblock
def repeat(n, f):
    for _ in range(n):
        f()

repeat(10):
    print "this"
```

erezsh:
Btw, have you seen my new project? https://github.com/erezsh/Preql

PR #816, opened 2021-01-19
GrammarBuilder - after refactor

erezsh:
I made a few changes to your pull request. Might as well let you make sure I didn't miss anything.

Biggest change was to load_grammar. I deal with the imports first, which simplifies the function.

This is close to ready to merge.

The `grammar_building.py` example needs a docstring (like all the other exampes), and a few more comments.

I'm also not a fan of how "mangling" is implemented. I'd much rather that instead of the mangle function, imports received a naming-path that they could join on.


MegaIng:
Note that this refactor does change the behavior: my design allowed imported grammars to extend rules defined in the base grammar. But I have to admit that I right know can't think of a use-case that doesn't have a cleaner solution.

This weird design was based on my original hope that I could implement `load_grammar` in a way in which there is not difference between executing `load_grammar` for the entire grammar and executing `load_grammar` for each single grammar item/statement. This sadly doesn't work because imports are combined by the old implementation (this is of course the correct & smart behavior), which I couldn't replicate without a lot more complex things (like delaying imports/extend to `build_grammar` or renaming already mangled rules back, which is also confusing).

In regards to mangle:

I am using it like this: https://github.com/MegaIng/syntax-extensions-base/blob/5d38d9c73d4c6fe84d38b916d3256a47119bf85a/src/syntax_extensions/base/python_parsing.py#L267

I don't wanna lose the ability to use `GrammarBuilder` like this (e.g. extend a base grammar with arbitrary extensions that also use a set of symbols from the base grammar). If you have a good idea on how to change the implementation without breaking this. But I agree that better information for imports would be nice.

MegaIng (approved):

MegaIng:
Also, I think there is still a problem in the form that unused imported Terminals/Token are not filtered out. This was previously done, but doesn't happen right now. They still get filtered out, but only after being processed in `Grammar.compile`. (I discovered this because it exposed a version related bug in `common.lark`, the now changed `C_COMMENT`). I couldn't really think of a way of implementing this (well, no simple way at least).

erezsh:
> my design allowed imported grammars to extend rules defined in the base grammar

I think it's good that this is no longer possible

> In regards to mangle:

I'll give it some thought. Anyway, it's something that can always be done later.

> Also, I think there is still a problem 

Can you create a code snippet that reproduces it?

MegaIng:
> Can you create a code snippet that reproduces it?

It isn't a problem per se. But just do `%import common.NUMBER` and compare the content of the `Grammar` object with and without this PR. previously https://github.com/lark-parser/lark/blob/d98925525f70285331987665a96e2859fd0a73f1/lark/load_grammar.py#L742 filtered a lot out. This doesn't happen now. The result of `Grammar.compile` should be the same. The 'problem' here is that it might represent a performance hit.

erezsh:
But it's just for explicitly imported symbols, right? So it's unlikely to have any performance impact (unless someone manually imports thousands of unused symbols)

MegaIng:
Supose two grammars:

a.lark

```
start: rb1 ra1
ra1: "A"
%import rb1
```

b.lark
```
start: rb1 rb3
rb1: "B" rb2
rb2: "C"
rb3: "D"
```

We are loading grammar a

-----

Before this PR:

`load_grammar('a.lark')` would `load_grammar('b.lark')`, which would return a `Grammar` instance. This grammar instance contains all rules in `b.lark` (`start`,`rb1`, `rb2`, `rb3`). Then `load_grammar('a.lark')` would call `import_from_grammar_into_namespace(b_grammar, 'b', ['rb1'])`, which would filter out all rules not directly or indirectly imported, so only (`rb1`, `b__rb2`) would be exported, and the final grammar object for `a.lark` would contain (`start`, `ra1`, `rb1`, `b__rb2`).

This means that `Grammar.compile` never sees these rules and doesn't have to do any processing on top.
-----

After this PR:

`load_grammar('a.lark')` calls `GrammarBuilder.load_grammar`, which doesn't return a `Grammar` instance. It also doesn't call something comparable to `import_from_grammar_into_namespace`, instead importing all rules. This means that the final `Grammar` instance for `a` contains (`start`, `ra1`, `b__start`, `rb1`, `b__rb2`, `b__rb3`), despite two of these rules being unused. They only get filtered out at the end of `Grammar.compile`, which means that potentially complex optimizations (EBNF2BNF or similar) are done potentially a lot of unused rules. I don't know if this is an performance hit for real grammars.

erezsh:
Oh, right, you removed the clever bfs thing I put there.

Why not put it back at the end of `load_grammar`, and add another parameter `used_symbols: Optional[list[str]]` ?

erezsh:
@MegaIng I fixed this branch, to remove unused rules/terminals again when importing.

Have a look. I'll probably merge it later today if there are no objections.

lark/load_grammar.py
```diff
@@ -849,6 +846,32 @@ def _parse_grammar(text, name, start='start'):
     return PrepareGrammar().transform(tree)
 
 
+def _get_mangle(prefix, aliases, base_mangle=None):
```
MegaIng, line 849: I would like this to be part of the public interface.

MegaIng (changes requested):
Also, stubs might still have to be updated. And I am still not convinced that the names are good like they are.

MegaIng:
Major refactor idea: Make `Symbol` another string subclass like `Token`, and add a new attributes `id` (this is the old value, e.g. with the mangled input names), `name ( the actual name in the grammar (e.g. only the last name, without the path)), `nice_name`/`qualname` (dot separated, for nicer debugging). This new `Symbol` would then be used everywhere (`Tree`, `Token`, `GrammarBuilder`). Of course, we can't expect `Tree.data` to always be a `Symbol` (because users might create them), but I think in the backend it might be very helpful if we start using the `Symbol` class earlier than we are currently doing, and also expose it to the user.

erezsh:
Can you explain what this would improve?

lark/load_grammar.py
erezsh: It would be easy to bring back later. Tbh, I'm not sure I want it even for the internal interface :)

erezsh: Another approach: change names from string to tuple, and store the path there. Comparisons and checks should all work as usual, with minor changes. Then in the end of the process, all names will pass through a "mangler" which by default just joins the tuple with `__`, but can be provided to the class by the user.

That should work, right?

Issue #826, opened 2021-02-04
Transfomer_NonRecursive Not Calling Token Userfunc


mikeiovine:
**Description**

I am trying to use `Transfomer_NonRecursive` since I have a very large parse tree. However, I don't think it is calling the token user funcs I defined. 

**To Reproduce**
Example grammar:
```
OP: "+"
?expr : (expr OP)? STRING
STRING: /\"[^"^+]*\"/
```
And a script:
```python
import lark

parser = lark.Lark.open(
    'example.lark', start='expr')

tree = parser.parse('"test"+"hello"+"world"')

class Transformer(lark.visitors.Transformer):

    def expr(self, values):
        concat = str(values[0]) + str(values[1]) + str(values[2])
        return concat

    def STRING(self, value):
        print('Token func called')
        return value[1:-1]

transformer = Transformer()
print(transformer.transform(tree))
```
When I use `Transformer`, I get the expected result; "Token func called" is printed 3 times, and the output of the last line is `test+hello+world`. However, when I use `Transformer_NonRecursive`, the token func is not called, and the output of the last line is `"test"+"hello"+"world"`

**Possible Fix**
I believe the issue [is here](https://github.com/lark-parser/lark/blob/master/lark/visitors.py#L222). That line should be:
`stack.append(self._call_userfunc_token(x))`.
Edit: err, I guess it should probably query `__visit_tokens__`, but the idea is the same.

MegaIng:
I made a PR #827  based on your suggestion. Could you check that it solves your problem?

mikeiovine:
Looks good. Thank you!

erezsh:
Merged. Good job everyone.

PR #827, opened 2021-02-05
Fix for #826 + stubs + tests

MegaIng:
Makes Transfomer_NonRecursive  call the Token callbacks. Also introduces Tests that all four Transformer variants do the same stuff.

The fix @mikeiovine proposed was exactly correct.


erezsh:
Nice, thanks!

Now we just need it to work for embedded transformers :)

Issue #818, opened 2021-01-25
Terminals cannot be transformed in tree-less LALR?


qwertystop:
**Describe the bug**

I've written a transformer with the line `DECIMAL = float` (from `%import common.DECIMAL` in the grammar) and it seemed to work fine post-transforming an Earley-parsed tree – that is, the relevant locations in the transformed tree would be `48.0` rather than `Token('DECIMAL', '48.0')` and similar.

I've written another transformer, this one for a grammar that I was able to make run in LALR. Therefore, I tried a tree-less parse. Now I get `ValueError: Callbacks must return a token (returned 48.0)`. To make the grammar work for a tree-less parse, I must wrap all my terminals in rules just to be able to transform them.

This seems to me unnecessarily tedious, and expands the size of the grammar unhelpfully. Also, transformer callbacks for rules are more long-winded to write – you can't do `decimal = float`, at minimum you have to do `decimal = lambda self, n: return float(n[0])`

**To Reproduce**
```python
from lark import Lark, Transformer
grammar = """
    start: NUM
    NUM: /\d+/
"""

class T(Transformer):
    NUM = int

T().transform(Lark(grammar, parser="lalr").parse("32"))
# returns Tree('start', [32])
Lark(grammar, parser="lalr", transformer=T()).parse("32")
# raises ValueError("Callbacks must return a token (returned 32)")
```

MegaIng:
@erezsh The problem is that the TERMINAL callbacks get added to `lexer_callbacks`, which of course happens before the parser see them. This also means the Transformer can actually change how to parser deals with the terminals, since it can change their type. This also means the this will be a backwards incompatible fix. I am not even sure if there is a clean way of fixing this.

erezsh:
I don't think there's a way to fix this that makes sense.

We can process the terminals after the parse tree, but the user can do that manually as well, and it's better to be transparent about this.

erezsh:
I suppose another approach would be to replace the return values with a proxy object, that also maintains the token type. But that sounds a little too "magical".

MegaIng:
[...]

PR #829, opened 2021-02-06
Fix for #818 + tests

MegaIng:
Allows internal Transformers to return non-Tokens from TERMINAL callbacks, like when used as a separate step.

This is done by moving the handling from after lexing to after matching in `ParserState.feed_token`.

This PR still has a few problems, as shown by the failing tests:

- deserializing is broken since we now have a cycle dependency between creating the callbacks and deserializing the list of terminals. Do you know an easy fix?
- The new tests expose a bug introduced in #379, which makes all `Transformer_InPlace` instances passed as internal transformers behave like they have `@v_args(tree=True)`. I do not have an overview what the best fix here is.

For both I can take a look but maybe you have a simple idea how to fix it.

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -185,7 +185,7 @@ class LarkOptions(Serialize):
 
 # Options that can be passed to the Lark parser, even when it was loaded from cache/standalone.
 # These option are only used outside of `load_grammar`.
-_LOAD_ALLOWED_OPTIONS = {'postlex', 'transformer', 'use_bytes', 'debug', 'g_regex_flags', 'regex', 'propagate_positions', 'tree_class'}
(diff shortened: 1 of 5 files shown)


MegaIng:
@erezsh Ok, this is now working. The bug in #379 of course is not a bug, but the correct behavior for the `_InPlace` Transformers. I just missed it.

erezsh:
It seems that you have a bug, when used with caching.

I added a test here: https://github.com/lark-parser/lark/tree/MegaIng-fix-818

I think the fix shouldn't be too hard.

MegaIng:
@erezsh Ok, I fixed this, but while playing around with the test I discovered that caching tries to pickle stuff like `Transformer` and `lexer_callbacks`, which breaks if they are function local, despite them probably being accessible later when passed as arguments again. I would suggest completely removing them from being pickled and forcing users to pass them again.

PR #844, opened 2021-02-25
Added load_grammar.find_grammar_errors

erezsh:


erezsh:
@MegaIng No comment?

MegaIng:
Sorry I forget this. I will take a look. I couldn't do a deepdive when I first saw this PR.

lark/load_grammar.py
```diff
@@ -853,6 +853,48 @@ def _parse_grammar(text, name, start='start'):
...
+                yield path + (choice,), new_p
+
+    for path, p in bfs_all_unique([((), puppet)], expand):
+        if predicate(p):
+            return path, p
+
+def find_grammar_errors(text, start='start'):
+    errors = []
+    def on_error(e):
+        errors.append((e.line, e.column, _error_repr(e)))
```
MegaIng, line 885: Why not return a copy of the actual error instance, potentially without the puppet? Just returning a string loses quite a bit of information.

lark/load_grammar.py
```diff
@@ -853,6 +853,48 @@ def _parse_grammar(text, name, start='start'):
...
+
+        # recover to a new line
```
MegaIng, line 887: This doesn't reset the lexer to a new line, right? I think that might hurt this quite a bit since it produces a lot of duplicate errors.

MegaIng (changes requested):

MegaIng:
Also, that function should be added to the stubs.

lark/load_grammar.py
erezsh: You're right 👍

lark/load_grammar.py
erezsh: I added a test for bad characters. Unless you mean something else?

MegaIng: I mean to advance the lexer state manually so that we are after the next newline. Then there is a higher chance to avoid errors chains of false positives.

erezsh:
@MegaIng I addressed all your comments.

lark-stubs/load_grammar.pyi
```diff
@@ -24,3 +24,6 @@ class GrammarBuilder:
     def validate(self) -> None: ...
 
     def build(self) -> Grammar: ...
+
+
+def find_grammar_errors(text: str, start: str='start'): ...
```
MegaIng, line 29: Adding a return type is a good idea I think

erezsh: You're not wrong. Added now.

lark-stubs/load_grammar.pyi
```diff
@@ -24,3 +25,6 @@ class GrammarBuilder:
     def validate(self) -> None: ...
 
     def build(self) -> Grammar: ...
+
+
+def find_grammar_errors(text: str, start: str='start') -> List[UnexpectedInput, str]: ...
```
MegaIng, line 30: List[Tuple[]]

erezsh: Ach, my bad. Fixed.

MegaIng: I recognize this instantly because I also do this regularly.

lark/load_grammar.py
MegaIng: This is IMO the only open thing here. If you don't wanna do this its fine.

erezsh: Look, it's working, isn't it? 

https://github.com/lark-parser/lark/pull/844/files#diff-e4667073752b1036b6ea3454b46d6224c9137197d5e970e82da63d82867d04aaR195

MegaIng: Yes it does. But if you look at the locations report for the text you linked (not only the set, but the actual entries), this is what you get: `[(2, 9), (2, 10), (2, 11), (2, 12), (2, 13), (2, 14), (2, 15), (2, 16), (2, 17), (6, 14)]` 9 errors, when there are actually only two. Yes, we could just filter afterwards but I think doing some effort in this function is a good idea.

MegaIng: Ok that was the wrong example. There are actually that many errors in that grammar. For the second example on the other hand we get `[(3, 11), (4, 9), (4, 13), (6, 11), (6, 13), (6, 22)]`. despite there being only 2 (maybe three)

erezsh: Now I understand what you mean.

It would be nice if we could make perfect recovery, using a more clever heuristic. But I currently don't have a good idea on how to do it.

I'll give it some more thought, but it's probably better to do it at a new PR, and for now to just return the first error on each line.

That should be good enough for now (i think?) and can be used in the various IDE efforts, which will also benefit as it improves over time.

MegaIng: Yes just returning the first error on every line has approximately the same result as my idea.

MegaIng (approved):

erezsh:
Merged. Thanks!

PR #846, opened 2021-03-01
Added ast_utils + example

erezsh:


MegaIng:
I would suggest not forcing the user to use the `camel_to_snake` system. Also sometimes you might want one class for different nodes. I would suggest adding a `__init_subclass__` and a class level attribute/class level keyword argument to modify this.

erezsh:
@MegaIng Rules are snake_case, so I can't see a situation when you would want to keep the caps.

You can do something like

```
class MultiPurposeNode(Ast): ...

usecase1 = MultiPurposeNode
usecase2 = MultiPurposeNode
usecase3 = MultiPurposeNode
usecase4 = MultiPurposeNode
```

It should work.

MegaIng (approved):

erezsh:
(nothing changed, I just added the pyi file)

MegaIng (approved):

MegaIng:
Btw, mypy throws errors, but the check still succeededs.

erezsh:
I don't use mypy, I'm only adding the stubs as a courtesy for the users.

If you know how to fix the mypy errors, go ahead.

MegaIng:
Well, we have a check setup here on github for type checking. If you look at the output, you see that mypy complains. But the check doesn't fail.

MegaIng:
We are specifically ignoring these errors: https://github.com/lark-parser/lark/blob/eac537a434dd6b5d2145d147f8fd90aa0131a5db/.github/workflows/mypy.yml#L19

I think I will try to fix all the errors.

PR #859, opened 2021-03-22
A few updates to stubs (fix #856)

### LINKED ISSUE #856: type hints need to be updated for lark.Lark's postlex parameter
**Describe the bug**

Mypy complain about the following code:
```py
from pathlib import Path
from typing import List

import lark
import lark.indenter

here = Path(__file__).parent


class Indenter(lark.indenter.Indenter):
    NL_type = "_NL"
    OPEN_PAREN_types: List[str] = []
    CLOSE_PAREN_types: List[str] = []
    INDENT_type = "_INDENT"
    DEDENT_type = "_DEDENT"
    tab_len = 4


grammar = lark.Lark(
    here.joinpath("grammar.lark").read_text(), parser="lalr", postlex=Indenter()
)
```
And it complains:

```
foo/parse.py:20: error: Argument "postlex" to "Lark" has incompatible type "Indenter"; expected "Optional[PostLex]"
Found 1 error in 1 file (checked 1 source file)
```
I'm copy-pasting [the docs](https://lark-parser.readthedocs.io/en/latest/examples/indented_tree.html) (with some edits). I believe this is a type stub problem.

---

(Poetry) Lockfile: https://gist.github.com/ThatXliner/e311980a936ef568e2d9caeb0fb5d2a7
Python: 3.6.13

**To Reproduce**

See above

Provide a short script that reproduces the erroneous behavior.

If that is impossible, provide clear steps to reproduce the behavior.

MegaIng:
Contains a few updates for the stubs. Also introduces `PostLex` as an actual subclassable Runtime class to reduce confusion.

### DIFF
--- a/lark-stubs/grammar.pyi
+++ b/lark-stubs/grammar.pyi
@@ -7,3 +7,8 @@ class RuleOptions:
     priority: int
     template_source: Optional[str]
     empty_indices: Tuple[bool, ...]
+
+
+class Symbol:
+    name: str
+    is_term: bool
--- a/lark-stubs/indenter.pyi
+++ b/lark-stubs/indenter.pyi
@@ -3,9 +3,10 @@
 from typing import Tuple, List, Iterator, Optional
 from abc import ABC, abstractmethod
 from .lexer import Token
+from .lark import PostLex
 
 
-class Indenter(ABC):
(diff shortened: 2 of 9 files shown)


erezsh:
Thanks!

PR #862, opened 2021-03-24
Add cache to ForestToParseTree

chanicpanic:
This PR adds a cache to `ForestToParseTree` to save the results of some node transformations. Correct me if I'm wrong, but the reuse of subtrees should not be problematic as Lark's parse trees are actually DAGs. The cache can provide significant performance benefits for ambiguous parses with `ambiguity='explicit'`. 

Here is a script that can be run with and without the cache to highlight the difference in runtimes:

```python
from lark import Lark

grammar = r"""
start: (string | string2)+
string: /\w+/
string2: /\w+/
"""

p = Lark(grammar, ambiguity='explicit', lexer='dynamic_complete')
p.parse('abcdefghijk')
```

For me, the above code runs around 16 times faster with the cache (1 second vs. 16 seconds).

### DIFF
--- a/lark/parsers/earley_forest.py
+++ b/lark/parsers/earley_forest.py
@@ -499,14 +499,18 @@ class ForestToParseTree(ForestTransformer):
     resolve_ambiguity: If True, ambiguities will be resolved based on
         priorities. Otherwise, `_ambig` nodes will be in the resulting
         tree.
+    use_cache: If True, the results of packed node transformations will be
+        cached.
     """
 
-    def __init__(self, tree_class=Tree, callbacks=dict(), prioritizer=ForestSumVisitor(), resolve_ambiguity=True):
+    def __init__(self, tree_class=Tree, callbacks=dict(), prioritizer=ForestSumVisitor(), resolve_ambiguity=True, use_cache=True):
         super(ForestToParseTree, self).__init__()
         self.tree_class = tree_class
         self.callbacks = callbacks
         self.prioritizer = prioritizer
         self.resolve_ambiguity = resolve_ambiguity
+        self._use_cache = use_cache
+        self._cache = {}
         self._on_cycle_retreat = False
         self._cycle_node = None
         self._successful_visits = set()
@@ -515,6 +519,7 @@ class ForestToParseTree(ForestTransformer):
         if self.prioritizer:
             self.prioritizer.visit(root)
         super(ForestToParseTree, self).visit(root)
+        self._cache = {}
 
     def on_cycle(self, node, path):
         logger.debug("Cycle encountered in the SPPF at node: %s. "
@@ -578,6 +583,8 @@ class ForestToParseTree(ForestTransformer):
         self._check_cycle(node)
         if self.resolve_ambiguity and id(node.parent) in self._successful_visits:
             raise Discard()
+        if self._use_cache and id(node) in self._cache:
+            return self._cache[id(node)]
         children = []
         assert len(data) <= 2
         data = PackedData(node, data)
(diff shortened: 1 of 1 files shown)


MegaIng:
While they are DAGs, and therefor there is no infinite loop, the fact that they are shared and mutable, does mean that modifying one branch of the ambiguity might have effects everywhere else. We have a few options:

- Leave as is, and just say 'you aren't supposed to modify the resulting Tree'. This is a potentially breaking change
- Give the end user the option to disable caching. This ofcourse loses the performance benefit
- Add a 'true_copy' method to the `Tree` class, that makes sure that there are no shared Trees (neither `.copy()` nor `deepcopy()` work)
- Make sure to copy the trees when using them from cache. For that we still need `true_copy` (I think. Maybe `deppcopy` is enough)

chanicpanic:
Right, I see the potential challenges. After thinking through the various options, my suggestions are:

- Disable caching by default, but give the user an option to enable it with warnings in the docs.
- Only use caching when `ambiguity='explicit'` and make a copy of the parse tree before returning it to the user. I think a deep copy may be achieved with `Transformer().transform(tree)`.

erezsh:
Tbh, I'm not sure that it's such a big deal. No where in Lark does it say that trees are all unique objects.

At the same time, I can see how it might end up confusing users who change the tree in-place.

@chanicpanic, Is there any use for this outside of `ambiguity='explicit'`? If not, maybe we can add a new value, `ambiguity='explicit_shared'` ?

Fwiw, I think most users would prefer performance over the hazard of shared trees.

MegaIng:
Is this caching relevant when `ambiguity!='explicit'`? If it brings a performance benefit, I would be against disabling it.

> Only use caching when ambiguity='explicit' and make a copy of the parse tree before returning it to the user. I think a deep copy may be achieved with Transformer().transform(tree).

After thinking about it a bit more, I would say this: Default behavior should be to *not* copy to save memory, but emit a warning when `debug=True` telling the user that they can easily make a copy with `<insert new function name>` (or with `Transformer().transform`, but that feels a bit weird for me), when they need to modify it. I would assume that most of the time they don't need to modify it, and they will just appreciate the speed up and memory down.

chanicpanic:
@erezsh Off the top of my head, I cannot think of a situation in which caching with `ambiguity!='explicit'` would provide any notable performance benefit. Explicit ambiguity benefits because it can reuse common subtrees across different derivations, but `ambiguity='resolve'` does not as it only produces a single derivation.
I like the idea of `ambiguity='explicit_shared'`.

@MegaIng I agree, it may not be worthwhile to copy the tree by default. Also, I figure that most users who want to modify the tree would be using a custom subclass of `Transformer` which would not modify the tree in-place.

MegaIng:
> Also, I figure that most users who want to modify the tree would be using a custom subclass of Transformer which would not modify the tree in-place

Unless they use `Transformer_InPlace`, which still has a decent chances of not breaking, since it already correctly visits duplicate Trees only once.

erezsh:
@chanicpanic @MegaIng I'm planning to merge it in this week, but keep the cache off by default, for now.

Any objections?

MegaIng:
@erezsh Wasn't the current consensus of this discussion that while the cache might in theory break stuff, we couldn't really find a situation where it is a big problem that can't be resolved easily? I would actual suggest keeping cache on by default and therefore making a very minor breaking change. The speedup is worth it.

erezsh:
Alright. Let's do it then. I'll just make a mental note that the next version should be 0.12 (if not 1!)

PR #866, opened 2021-04-02
Fix copying and comparing of some classes

### LINKED ISSUE #865: ParserPuppet.copy does not copy LexerState
**Describe the bug**

`ParserPuppet.copy` does not copy the lexer state. Instead it makes a shallow copy of `LexerThread`.

**To Reproduce**

```python3
# make a copy
puppet_copy = puppet.copy()

# advance LineCounter of copy
puppet_copy.lexer_state.state.line_ctr.feed('foo')

# assertion fails
assert puppet.lexer_state.state.line_ctr.char_pos != puppet_copy.lexer_state.state.line_ctr.char_pos
```

This should be simple to fix. The reason for the bug is that the member `lexer_state` of `ParserPuppet` is in fact not a `LexerState` but a `LexerThread` (see [here](https://github.com/lark-parser/lark/blob/1d61de4f94b4af441b2f32c790db4290348d7a93/lark/parsers/lalr_parser.py#L149)), which does not have a custom copy function. So `copy(self.lexer_state)` would only be a shallow copy. I think we can just add a `__copy__` method to `LexerThread` ([here](https://github.com/lark-parser/lark/blob/1d61de4f94b4af441b2f32c790db4290348d7a93/lark/lexer.py#L459)) to fix this, namely
```python3
    def __copy__(self):
        return type(self)(
            self.lexer,
            copy(self.state)
        )
```

evilnose:
This fixes #865 and extends the ParserPuppet test. Additional copy and comparison methods are added for a few relevant classes, for easier testing.

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -186,6 +186,13 @@ class LineCounter:
         self.column = 1
         self.line_start_pos = 0
 
+    def __eq__(self, other):
+        if not isinstance(other, LineCounter):
+            return False
+        return (self.newline_char == other.newline_char and self.char_pos == other.char_pos
+            and self.line == other.line and self.column == other.column
+            and self.line_start_pos == other.line_start_pos)
+
     def feed(self, token, test_newline=True):
         """Consume a token and calculate the new line & column.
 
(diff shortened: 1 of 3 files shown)


MegaIng (approved):

erezsh:
Thanks!

PR #869, opened 2021-04-07
Added get_puppet, helpers and tests

### LINKED ISSUE #868: Obtain ParserPuppet directly?
**Suggestion**
I'm writing a language server with Lark, and I'm using a recursion-based approach for error handling and inference. For this I make use of `ParserPuppet` extensively, as a helper to store copies of states and try other branches. However, right now `ParserPuppet` can only be created through exceptions during parsing. It would be much more convenient if there is a method that generates a `ParserPuppet` directly based on the initial states of the parser, so that I can start off with a puppet at the root.

This should be as simple as creating a method similar to `_Parser.parse` in `LALR_Parser` but instead returning a `ParserPuppet` instead of parsing.

Creating a `ParserPuppet` manually, but this feels quite hacky as `ParserPuppet`, `ParserState` and `ParseConf` are not explicitly exported.

**Additional context**
None

MegaIng:
Closes #868.


erezsh:
Overall looks okay. I'll take a deeper look and merge it later on.

@MegaIng  @evilnose 

Slightly related: What do you think of renaming the "puppet" to "interactive parser" ?

Then instead of starting it in "puppet mode", you'll be starting the parser in "interactive mode".

An alternative name is "single step parser". Another suggestion that's been raised is "parser cursor".

Let me know your thoughts.

evilnose:
> Overall looks okay. I'll take a deeper look and merge it later on.
> 
> @MegaIng @evilnose
> 
> Slightly related: What do you think of renaming the "puppet" to "interactive parser" ?
> 
> Then instead of starting it in "puppet mode", you'll be starting the parser in "interactive mode".
> 
> An alternative name is "single step parser". Another suggestion that's been raised is "parser cursor".
> 
> Let me know your thoughts.

I think "interactive parser" is a good and descriptive name, especially since after this merge the user can start off with the puppet and write their own LALR parser loop.

MegaIng:
I agree that "interactive parser" is the most descriptive name. I don't like it because it is another class named `Parser`, and `InteractiveParser` is quite long. `Puppet` is a nice short name. But changing it is probably a good idea.

evilnose:
Specifically for renaming `Puppet`, I think `Cursor` is a nice alternative too.

Issue #871, opened 2021-04-08
Invalidate cache when grammar changes


alexandermorozov:
Currently you have to manually delete cache files if you update the grammar. It's quite error prone, and sometimes can go unnoticed.

It'd be great to ensure that all source .lark files that were used haven't changed, and invalidate the cache if they did.

MegaIng:
The problem here is that we don't know on which files the grammar depends before parsing it, which would kinda defeat the purpose of the cache. While this is certainty possible, the easiest fix is for the user to just not specify `cache=True` while developing the grammar.

alexandermorozov:
During initial parse it's possible to store list of files which generated the grammar and their checksums. If the grammar doesn't support incudes with globs and specifies file names exactly, then on cache load we can just test that checksums match.

It'll ensure that all files are unchanged. And since they are unchanged, no new files were referenced.

Actually patterns (*, ?, ...) can be also handled -- just include the list in cache and do expansion when loading the cache to check that no new files are referenced.

MegaIng:
Yes, it is possible, at the cost of quite a bit of complexity in different places:

Currently there are two distinct steps to loading from cache:

- Check if we got different inputs hash(given grammar + options_string)
- Deserialize the `Lark` instance

We would have to add another non-trivial step in here that:
- Deserialize another object (e.g. a third value stored in the file)
- Does a non-trivial amount of file access, potentially reducing the performance of the cache significantly.

We would also have to record the accessed files:
- (No lark does not support glob imports)
- This would certainly be possible, but would be another decent spaghetti factor in `load_grammar.py`.

Considering that this would reduce the overall performance of `cache=True` even when nothing changed, and you can get equivalent results by passing `cache=False`, I don't think this is worth the trouble.

erezsh:
@MegaIng It can actually become quite cumbersome when the grammar is fragmented (i.e. includes many `%import`s)

Also, most of the performance hit comes from the LALR analysis, and not the initial parsing of the file.

So I think this feature is useful, not that hard to do, and I also don't think the performance difference will be noticeable.

alexandermorozov:
There definitely would be an impact from having to read and checksum all source files. Usually it should be quite fast, though -- I suspect impact with respect to CPU time would be less than a few percent of what it takes to unpickle the grammar. IO cost may wary though -- if those files are on sshfs or such it can easily multiply load time by 2-10x depending on number of files accessed. That's an edge case though.

[...]

PR #872, opened 2021-04-09
Better cache

### LINKED ISSUE #870: Make cache file more resilent
Currently grammar cache files that Lark creates can end up containing malformed data. It can happen as a result of:
1. Python version mismatches,
2. Truncated file in case of power loss, or python interpreter getting killed or interrupted in the middle of the write.
3. Simultaneous write from multiple threads / processes,

It can be solved by implementing the following:

1. Adding a suffix that indicates version of interpreter that created the cache file. I believe you'd need first 2 numbers from python version + interpreter type, e.g. cpython-3.8, pypy-3.5.

2. Making sure that contents of file are commited to disk and the cache file is atomically replaced. Here are approximate steps: a) create a temp file in the same dir as cache file (to make sure that the files are on the same filesystem). Python has a lib to work with temp files, b) after data is written to the file, do `f.flush()` to flush in-python buffer, and then `os.fsync(f)`. It'll make sure that contents are properly commited to disk, and not just pushed into OS-level RAM fs cache. Then do `os.rename` to replace the old cache file (if any?) with the new one. This op is usually atomic -- the file is either renamed or it's not.

3. Actually step 2) would already prevent data races from multithreaded writes since they'd write into different files, and rename is atomic. But it's possible to go futher and use file-based locks (e.g. filelock module) to make sure that only one process does all the compute, and all other processes idle until the cache is ready. Larger grammars can take tens of seconds to prepare, so it can add up, especially if lark is used in multithreaded/multiprocess server where grammar is loaded on init (20 processes, 30 seconds each -> 10 minutes of CPU time). I'm not sure that it's Lark responsibility though, and you may not want to add dependency on other packages.

MegaIng:
Should close #870 and #871

@alexandermorozov I hope this correctly addresses your concerns.

(especially the file verification was actually a lot simpler than I originally though. Although it isn't perfect (it doesn't detect changes to`common.lark` for example.))

### DIFF
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -38,6 +38,7 @@ class LarkOptions:
     use_bytes: bool
     import_paths: List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]
     source_path: Optional[str]
+    safe_cache: Literal[False, True, "atomic"]
 
 
 class PackageResource(object):
@@ -81,6 +82,7 @@ class Lark:
         use_bytes: bool = False,
         import_paths: List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]] = ...,
         source_path: Optional[str]=None,
+        safe_cache: Literal[False, True, "atomic"]=True,
     ):
         ...
 
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -38,7 +38,6 @@ class LarkOptions:
     use_bytes: bool
     import_paths: List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]
     source_path: Optional[str]
-    safe_cache: Literal[False, True, "atomic"]
 
 
 class PackageResource(object):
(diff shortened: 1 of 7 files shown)


erezsh:
I like the idea of saving the list of dependencies in the cache!

Why won't it detect changes in common.lark ?

erezsh:
Why would anyone ever set `safe_cache=False` ?

If you're not using any imports, `verify_used_files()` is a no-op. And if you are using them, you'd want to them checked.

If someone is really trying to minimize file reads, they'd be better off using the standalone parser, or manually save/load.

MegaIng:
> Why won't it detect changes in common.lark ?

Because we are using `pkgutil.get_data` to support zip files. Supporting that as well is quite a bit more work (I am not even sure how we would do that).

> I like the idea of saving the list of dependencies in the cache!

Thank @alexandermorozov in #871.

MegaIng:
> (I am not even sure how we would do that).

Ok, I just got an idea. Will try to implement

erezsh:
@MegaIng 

`TypeError: 'override' is an invalid keyword argument for open()`

Did you mean overwrite?

PR #877, opened 2021-04-15
Rename puppet to interactive_parser

erezsh:
Based on #869


lark/parsers/lalr_puppet.py
```diff
@@ -1,96 +1,3 @@
...
-    def __hash__(self):
-        return hash((self.parser_state, self.lexer_state))
-
-    def feed_token(self, token):
-        c = copy(self)
-        c.result = ParserPuppet.feed_token(c, token)
-        return c
\ No newline at end of file
+# Deprecated
+
```
MegaIng, line 2: I would say that we can log a warning when this module is imported.

MegaIng (approved):
I still think that this name is a bit wordy. But I still can't really think of something better (IMO Cursor has little benefit over Puppet).

The only placse where puppet is still in the project are in expected places (e.g. compatibility), and in `examples/standalone/json_parser.py`. Might is be possible to not push the generated file? It will often be out-of-date anyway.

lark/load_grammar.py
```diff
@@ -866,7 +866,7 @@ def _error_repr(error):
     else:
         return str(error)
 
-def _search_puppet(puppet, predicate):
+def _search_interactive_parser(interactive_parser, predicate):
```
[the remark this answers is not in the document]
erezsh: Yes, it could be.

lark/parsers/lalr_puppet.py
erezsh: Sure, probably a good idea.

erezsh:
> Might is be possible to not push the generated file?

I agree it might be better not to. I think I had a reason for including it, but I can't remember what it was.

Issue #237, opened 2018-09-18
Add EOF symbol to match end of input


erezsh:
Can be signified with `$`

Must be optional (grammars don't have to contain it to be correct)

Need to make sure it works for both LALR and Earley, and with indentation.

SarthakDandriyal:
Hey can I work on this issue?

BTaskaya:
Any update?

erezsh:
I have a version that should be working for LALR, in branch `end_symbol`

If anyone wants to test it and let me know how it goes, that would be great!

@isidentical @jruales @coreyt

jisaacstone:
I did a rebase of this onto the master branch. There was a conflict in `lalr_parser.py` that I'm not sure I resolved correctly. 

https://github.com/jisaacstone/lark/blob/end_symbol/tests/test_parser.py#L1650-L1660

I but all the tests seem to pass. To be sure I added another test in and it failed

https://github.com/jisaacstone/lark/blob/end_symbol/tests/test_parser.py#L1650-L1660

```
======================================================================
ERROR: test_end_symbol2 (tests.test_parser._make_parser_test.<locals>._TestParser)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/larpy/lark/tests/test_parser.py", line 1659, in test_end_symbol2
    self.assertEqual(parser.parse('axa'), Tree('start', [Tree('a', []),Tree('a', [])]))
  File "/Users/larpy/lark/lark/lark.py", line 293, in parse
    return self.parser.parse(text, start=start)
  File "/Users/larpy/lark/lark/parser_frontends.py", line 88, in parse
    return self._parse(token_stream, start)
  File "/Users/larpy/lark/lark/parser_frontends.py", line 54, in _parse
    return self.parser.parse(input, start, *args)
  File "/Users/larpy/lark/lark/parsers/lalr_parser.py", line 36, in parse
    return self.parser.parse(*args)
  File "/Users/larpy/lark/lark/parsers/lalr_parser.py", line 110, in parse
    reduce(arg)
  File "/Users/larpy/lark/lark/parsers/lalr_parser.py", line 77, in reduce
    value = self.callbacks[rule](s)
  File "/Users/larpy/lark/lark/parse_tree_builder.py", line 28, in __call__
    res = self.node_builder(children)
  File "/Users/larpy/lark/lark/parse_tree_builder.py", line 121, in __call__
    filtered.append(children[i])
IndexError: list index out of range

[...]

PR #880, opened 2021-04-18
Reimplementation of end symbol (Issue #237)

erezsh:


MegaIng (approved):
Do you remember what the problem was last time you tried to implement this? And is it fixed?

lark/parsers/lalr_interactive_parser.py
```diff
@@ -21,18 +22,18 @@ def feed_token(self, token):
...
         
-        Note that this modifies the instance in place and does not feed an '$END' Token"""
+        Note that this modifies the instance in place and does not feed an END Token"""
         for token in self.lexer_state.lex(self.parser_state):
             self.parser_state.feed_token(token)
     
     def feed_eof(self, last_token=None):
-        """Feed a '$END' Token. Borrows from 'last_token' if given."""
-        eof = Token.new_borrow_pos('$END', '', last_token) if last_token is not None else Token('$END', '', 0, 1, 1)
+        """Feed a END Token. Borrows from 'last_token' if given."""
```
MegaIng, line 35: Sorry, overlooked this in the review: Shouldn't we feed multiple ENDs here as well?

erezsh: Hmm I think you're right.

erezsh:
I don't remember the problem, only that it messed some grammars up. But I changed the LALR logic since, made it a little cleaner, so maybe that was solved on the way?

I'll make sure to do thorough tests before merging it in.

ThatXliner:
I'm trying to test this right now. What's the terminal to use? `$END`?

MegaIng:
No, It is just `$`

ThatXliner:
Maybe I'm going too far but is this supposed to work with custom Indenters? Because I'm facing a problem where 
```
lark.exceptions.UnexpectedToken: Unexpected token Token('__$END$__', '') at line 1, column 1.
```

MegaIng:
It is supposed to work. Can you give a short example script?

erezsh:
`on_error` was broken before, but now I fixed it. So if you were using it, you should try again with the latest commit.

MegaIng (approved):

ThatXliner:
Now it's just really slow. Or maybe it's just my grammar....

erezsh:
@ThatXliner Thanks for finding that. I think I fixed it, so try now.

ThatXliner:
nope

erezsh:
@ThatXliner Ok then, hit me with your enchilada. But make sure it's easy to see it fail.

ThatXliner:
Probably it was hanging for valid input when I sent my last comment but I most likely didn't double-check. Now, when I make it valid, it hangs.

Here's the enchilada: https://gist.github.com/ThatXliner/96829ce180b5659198a822cf74b6c309

Tell me if you can reproduce or not

erezsh:
I found a minimal breaking example:
```python
from lark import Lark

p = Lark("""
    ?start: "foo" x "bar"
    x: $ x
""", parser="lalr")

p.parse("foo")
```

The infinite loop happens because `$` is accepted forever, but has no valid route to unwind the stack.

I need to think about how to solve this. My current intuition is that this grammar shouldn't even be allowed. But I'm not sure how easy it is to detect reliably.

But maybe I just need to rethink the loop mechanism, and solve it by only providing the end token once.

ThatXliner:
I'm wondering how you're implementing the end token for right now. I think it's better to parse normally, get an EOF error, see if the grammar expects it and if so, continue parsing. Unless that's what you're doing right now.

erezsh:
So, it seems that this can be "solved" by testing that the parser only shifts once on END, and otherwise throwing a run time error for infinite loop.

But I don't like it. Lark shouldn't allow to compile a grammar that can never run to completion. But, I'm not sure how to detect it at compile time.

lark/parsers/lalr_parser.py
```diff
@@ -132,7 +133,7 @@ def feed_token(self, token, is_end=False):
 
             if action is Shift:
                 # shift once and return
-                assert not is_end
+                # assert not is_end
```
ThatXliner, line 136: Wonder what would happen if you undid this

erezsh:
It's just an assert. It doesn't matter.

lark/parsers/lalr_parser.py
erezsh: It's just an assert. It doesn't matter.

seimit:
Hi
Do you have plan this PR to be merged into the master soon?
Regards!

erezsh:
@seimit There are issues with this PR, so currently no. 

The problem is that in some edge use cases, the parser gets stuck in a loop forever. Which is something I never want Lark to be able to do.

MegaIng:
@erezsh Shouldn't it be possible to add some early point reject all BNF rules that have a Terminal following an `$` symbol? E.g. if we find a rule of the form

```
a: $ A
```
we completely throw it away, since it can never actually match.

For rules of the form

```
a: $ b
```

We need to check if `b` can be empty, and if it can not be empty we throw the rule away. I think we check during Grammar Analyzing anyway if a rule can be empty.

erezsh:
It's a good point that we can detect it during grammar analysis. And then we can just throw an error if anything follows the $END symbol. I don't think there's a good enough reason to allow them, empty rules or not.

MegaIng:
Yes there is. `($|_NEWLINE)*` should be allowed. I think that would break if we don't allow it.

MegaIng:
OTOH, maybe it should be allowed. In effect it is the same as `_NEWLINE* $?`. I am not sure. But we need to make sure that for example the python grammar works when everywhere were we accept a statement ending `_NEWLINE` we also accept a `$`.

But also, can we somehow detect this:

```
a: end A
end: $
```
?

erezsh:
No, it would create a separate rule, `_ANON_STAR: _ANON_STAR _NEWLINE | _ANON_STAR $

erezsh:
If we do it at the grammar analysis level, we can make sure `FOLLOWS[$]` is empty. That should take care of all edge cases (I think)

erezsh:
Hmm actually, no, that's the crux of the problem, which I guess is why I left it open.

On one hand, the grammar has to allow the option of something after an end token, otherwise, there's no advantage in implementing it. On the other hand, it's not obvious how to block the bad uses while allowing the good ones.

MegaIng:
The point is more that there is that it is possible to accept nothing after `$`. Not sure how that can be expressed.

daniel-leicht:
This is a bad bad code, that will hurt - many people that are good.

Erotemic:
Throwing in my $0.02, just to say that this feature would be useful to me.

PR #883, opened 2021-04-24
Create __main__.py so the standalone tool can actually be used

ThatXliner:


MegaIng:
What? You are supposed to call `python -m lark.tools.standalone`

ThatXliner:
Thanks for the quick response!
> What? You are supposed to call `python -m lark.tools.standalone`

Wait really? I didn't see that in the docs:
Screen Shot 2021-04-24 at 1 51 58 PM
Screen Shot 2021-04-24 at 1 52 11 PM

MegaIng:
So your actual complaint is that the standalone is not in the docs? Considering the only alternative info you have about standalone should come from [the example](https://github.com/lark-parser/lark/tree/master/examples/standalone) which clearly shows how to use it, this seems like a weird PR then. Also, how about adding docs if you can't find them and create a solution?

ThatXliner:
I didn't know you were supposed to use `python -m lark.tools.standalone`. Sorry about that!

erezsh:
@ThatXliner Thanks for bringing it to my attention. There should be better documentation for it.

PR #889, opened 2021-05-05
Support 32bit unicode escape sequence

tsangwpx:
This PR adds 32bit unicode escape sequence and its test.

* 16bit unicode escape sequence `\uxxxx`: small `u` and 4 hexadecimal digits.
* 32bit unicode escape sequence `\Uxxxxxxxx`: capital `U` and 8 hexadecimal digits.

See 
https://docs.python.org/3.9/reference/lexical_analysis.html#string-and-bytes-literals


MegaIng (approved):

erezsh:
Thanks!

PR #891, opened 2021-05-06
Python3 Grammar, refactor for a better structure

erezsh:


erezsh:
@MegaIng This is mid-work, but maybe you'd like to review it, to stop me if I'm going in the wrong direction.

PR #907, opened 2021-06-05
Refactor PropagatePositions, change Token.pos_in_stream to start_pos

erezsh:
Increased minor version, due to possibly breaking changes, + invalidates serialized objects.

### DIFF
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -33,7 +33,7 @@ class LarkOptions:
     regex: bool
     debug: bool
     keep_all_tokens: bool
-    propagate_positions: bool
+    propagate_positions: Union[bool, str]
     maybe_placeholders: bool
     lexer_callbacks: Dict[str, Callable[[Token], Token]]
     cache: Union[bool, str]
--- a/lark-stubs/lark.pyi
+++ b/lark-stubs/lark.pyi
@@ -77,7 +77,7 @@ class Lark:
         regex: bool = False,
         debug: bool = False,
         keep_all_tokens: bool = False,
-        propagate_positions: bool = False,
+        propagate_positions: Union[bool, str] = False,
         maybe_placeholders: bool = False,
         lexer_callbacks: Optional[Dict[str, Callable[[Token], Token]]] = None,
         cache: Union[bool, str] = False,
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -6,6 +6,7 @@ from .utils import Str, classify, get_regexp_width, Py36, Serialize, suppress
 from .exceptions import UnexpectedCharacters, LexError, UnexpectedToken
 
 ###{standalone
+from warnings import warn
 from copy import copy
 
 
@@ -128,9 +129,9 @@ class Token(Str):
             end_column will be 5.
         end_pos: the index where the token ends (basically ``pos_in_stream + len(token)``)
     """
-    __slots__ = ('type', 'pos_in_stream', 'value', 'line', 'column', 'end_line', 'end_column', 'end_pos')
+    __slots__ = ('type', 'start_pos', 'value', 'line', 'column', 'end_line', 'end_column', 'end_pos')
 
-    def __new__(cls, type_, value, pos_in_stream=None, line=None, column=None, end_line=None, end_column=None, end_pos=None):
+    def __new__(cls, type_, value, start_pos=None, line=None, column=None, end_line=None, end_column=None, end_pos=None, pos_in_stream=None):
         try:
             self = super(Token, cls).__new__(cls, value)
         except UnicodeDecodeError:
@@ -138,7 +139,7 @@ class Token(Str):
             self = super(Token, cls).__new__(cls, value)
 
         self.type = type_
-        self.pos_in_stream = pos_in_stream
+        self.start_pos = start_pos if start_pos is not None else pos_in_stream
         self.value = value
         self.line = line
         self.column = column
@@ -146,6 +147,11 @@ class Token(Str):
         self.end_column = end_column
         self.end_pos = end_pos
         return self
+
+    @property
+    def pos_in_stream(self):
+        warn("Attribute Token.pos_in_stream was renamed to Token.start_pos", DeprecationWarning)
+        return self.start_pos
 
     def update(self, type_=None, value=None):
         return Token.new_borrow_pos(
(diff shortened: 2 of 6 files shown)


lark/parse_tree_builder.py
```diff
@@ -21,6 +21,16 @@ def __call__(self, children):
...
 
 
+
+def _pp_get_meta(children):
+    for c in children:
+        if isinstance(c, Tree):
+            if not c.meta.empty:
+                return c.meta
+        elif isinstance(c, Token):
+            if not c.isspace():     # Don't regard whitespace-only tokens
```
MegaIng, line 31: It might be interesting to toggle this behavior: Sometimes one wants to not ignore Whit space (extreme example: The language White space) But I agree that ignore WS should be the default behavior. Note that this is backwards incompatible.

lark/parse_tree_builder.py
```diff
@@ -31,37 +41,20 @@ def __call__(self, children):
...
-                    break
+
+            src_meta = _pp_get_meta(children)
+            if src_meta:
+                res_meta.line = src_meta.line
+                res_meta.column = src_meta.column
+                res_meta.start_pos = src_meta.start_pos
+                res_meta.empty = False
+
+            src_meta = _pp_get_meta(reversed(children))
```
MegaIng, line 52: Can't we just reverse the previous `sre_meta`?

MegaIng (approved):

lark/parse_tree_builder.py
erezsh: I do wonder if anyone would actually need to account for the whitespace, when trying to find where nodes start and end.

I can't really imagine the situation, tbh. But perhaps a good compromise would be only to disregard whitespace tokens that being with an underscore?

So `_NEWLINE` won't be counted, but `NEWLINE` will be. I'm not sure how intuitive that is to the casual user. But at the same time, how often will anyone care?

Of course we can add yet another switch to Lark for this, but as you know, I'd like to avoid that as much as possible.

Or a directive like `%ignore_positions WS _NEWLINE`. Which also seems a bit excessive for this case.

lark/parse_tree_builder.py
erezsh: No, `_pp_get_meta` returns an object, not a list. It's more efficient this way

MegaIng: Yeah, my mistake.

lark/parse_tree_builder.py
MegaIng: One usecase where this change is potentially even breaking is when you want to classify what Tree node a specific position in the source file belongs to. If you as example have a dedicated `ws_or_comment: (WS|comment)+` rule then this change might make for confusing boundaries.

We don't need to add a new flag. We can just do what we do in so many other options as well: overload the keyword to something like `False`, `True`, `"keep_ws"`, `"ignore_ws"`.

Does Propagate position act on `%ignore` terminals and `_UNDERSCORE` Terminals? I thought they have already been thrown away by that point. That is at least the behavior I would have expected.

I would certainly not add a new flag/new syntax.

erezsh: > We can just do what we do in so many other options as well: overload the keyword

Hah, good point. Didn't even consider it.

Yeah, that might be the best way. That way "ignore_ws" can be a special behavior, and we won't have to break compatibility.

One question still remains - should it only discard _WHITESPACE, or also WHITESPACE? (underscore or not)

> Does Propagate position act on %ignore terminals and _UNDERSCORE

No, it ignores `%ignore` terminals, but does include _UNDERSCORE terminals, as they might include _COLON or _DEFINE etc.

MegaIng: If it normally behaves the same on with no regard to `_`, then it should continue to do so. If someone comes to us and wants propagate_position to ignore whitespaces at this specific point and not at others, we can still add a new option then.

erezsh:
@MegaIng Added a few changes. I'd appreciate if you take another glance.

All I wanted was to change two lines :'(

lark/parse_tree_builder.py
```diff
@@ -31,40 +32,52 @@ def __call__(self, children):
...
+    def _pp_get_meta(self, children):
+        for c in children:
+            if isinstance(c, Tree):
+                if not c.meta.empty:
+                    return c.meta
+            elif isinstance(c, Token):
+                return c
+
+class PropagatePositions_IgnoreWs(PropagatePositions):
+    def _pp_get_meta(self, children):
```
MegaIng, line 61: We can prevent duplicating the entire function if the base function reads a single class attribute (e.g. `if self.dont_ignore_ws or (c and not c.isspace())`. The child class would then set the attribute to True. But I am not sure if it is better.

lark-stubs/lark.pyi
```diff
@@ -77,7 +77,7 @@ class Lark:
         regex: bool = False,
         debug: bool = False,
         keep_all_tokens: bool = False,
-        propagate_positions: bool = False,
+        propagate_positions: Union[bool, str] = False,
```
MegaIng, line 80: LarkOptions attribute + doc string need to be updated as well.

MegaIng:
> All I wanted was to change two lines :'(

I know that feeling. But I am not really sure what we could not do right now without breaking compatibility.

lark/lexer.py
```diff
@@ -147,6 +148,11 @@ def __new__(cls, type_, value, pos_in_stream=None, line=None, column=None, end_l
         self.end_pos = end_pos
         return self
 
+    @property
+    def pos_in_stream(self):
```
MegaIng, line 152: Stubs need to be updated.

lark/parse_tree_builder.py
erezsh: It might be a bit shorter, but this feels more readable to me (I actually tried your version first, but didn't like it). Also the change would make the default mode slightly slower, though that hardly matters. Anyway, it's all the same to me, so if you want to change it later you're welcome to.

lark/lexer.py
erezsh: Thanks!

erezsh:
Is think we're there. Right?

Thanks for the patience btw. I wonder if there's a way to get some of these things into the tests. If not, I'll make a checklist for next time.

erezsh:
> I know that feeling. But I am not really sure what we could not do right now without breaking compatibility.

For one, we could steer hard into a Python3 only version 1.0, and change all the stuff we don't like. let it simmer for a few weeks, and get a lot of the kinks out.

No one will have to upgrade, of course. Lark is already very stable, and for the unlikely critical update, we can always make more 0.x releases.

Just a thought.

MegaIng:
> Is think we're there. Right?

(Except stubs for `LarkOptions`, but that is not that important) Yeah

erezsh:
Oh, got it.

PR #913, opened 2021-06-16
Adding parse context support for LALR(1) parser

kpdemetriou:


erezsh:
Can you explain the point of this change?

Why not hold the state in your transformer?

erezsh:
See the example. Lark accepts an **instance**, so you can initialize it with whatever you like: https://github.com/lark-parser/lark/blob/master/examples/json_parser.py#L66

kpdemetriou:
Sure, now imagine multiple parallel calls to `parse` where each has some context. How do you use a single transformer instance which holds said context without conflict?

erezsh:
Yes, I see. Have you considered using this?

https://docs.python.org/3/library/contextvars.html

(or `threading.local()`)

kpdemetriou:
I see how that might be a better alternative given the scale of changes that would need to happen in this project, thanks; closing the issue.

erezsh:
Thanks for understanding. And thanks for trying to contribute!

Fwiw, I've found myself using the context pattern more and more in my code. I even wrote a small utility around it: https://gist.github.com/erezsh/e31d930c68d94d0c95ae9bf2104d50a9

PR #925, opened 2021-06-27
Draft of Version 1.0

erezsh:
Changes so far

- Removed deprecated code
- Removed most of the code for Python 2 compatibility.


erezsh:
Tasks remaining:

- Merge .pyi files into the code, and remove them.
- Update the interface to be more self-consistent (ideas?)
- Improve the code with Python3 features, where possible (optional/low priority)
- Anything else?

.github/workflows/tests.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9.0-rc - 3.9, pypy2, pypy3]
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9.0-rc - 3.9, pypy3]
```
MegaIng, line 9: Didn't we say we start support with 3.6? To get more of the new features (`typing`, f-strings)

lark/exceptions.py
```diff
@@ -193,7 +191,7 @@ def __init__(self, token, expected, considered_rules=None, state=None, interacti
         # TODO considered_rules and expected can be figured out using state
         self.line = getattr(token, 'line', '?')
         self.column = getattr(token, 'column', '?')
-        self.pos_in_stream = getattr(token, 'pos_in_stream', None)
+        self.pos_in_stream = getattr(token, 'start_pos', None)
```
MegaIng, line 194: We can just rename the attribute.

lark/lark.py
```diff
@@ -1,14 +1,12 @@
 from __future__ import absolute_import
 
-
-from lark.exceptions import ConfigurationError, assert_config
-
+from abc import ABC, abstractmethod
 import sys, os, pickle, hashlib
 from io import open
```
MegaIng, line 5: Not needed in python3 AFAIK.

lark/tools/standalone.py
```diff
@@ -185,9 +180,6 @@ def main():
         parser.print_help(sys.stderr)
         sys.exit(1)
     ns = parser.parse_args()
-    if ns.old_start is not None:
-        warn('The syntax `python -m lark.tools.standalone <grammar-file> <start>` is deprecated. Use the -s option')
```
MegaIng, line 189: This should als be reflected in the parser itself.

lark/utils.py
```diff
@@ -222,34 +201,12 @@ def dedup_list(l):
...
-try:
-    compare = cmp
-except NameError:
-    def compare(a, b):
-        if a == b:
-            return 0
-        elif a > b:
-            return 1
-        return -1
+def compare(a, b):
```
MegaIng, line 204: Is there a good reason why we are using this function anywhere?

MegaIng (changes requested):
A few initial comments. I will take another look or too and maybe even just implement everything mentioned here in my own commits.

.github/workflows/tests.yml
erezsh: I wasn't sure, so I left it for now.

I think 3.6+ is probably the right idea, but I wonder if there's anyone still stuck on 3.5.

lark/exceptions.py
erezsh: You mean, in the exceptions?

I'm less concerned about that (and anyway, `start_pos` wouldn't be a good name in this case). But I'm open to suggestions.

lark/tools/standalone.py
erezsh: How do you mean?

lark/utils.py
erezsh: Couldn't find any. Might be a relic that we forgot to remove.

lark/tools/standalone.py
MegaIng: A few lines above this, there is a call to `parser.add_arguments('old_start')`

lark/lark.py
erezsh: Why not?

erezsh: Nevermind, right, it's just an alias.

MegaIng: https://docs.python.org/3/library/io.html#io.open

.github/workflows/tests.yml
ThatXliner: https://endoflife.date/python. I'd recommend starting with 3.6 (it was my personal choice for most of my projects, no regrets)

MegaIng: (Also, according to the [stats](https://pypistats.org/packages/lark-parser), python3.5 is used far less than 2.7)

ThatXliner:
About the type-merging, you can possibly use https://github.com/ilevkivskyi/com2ann to merge type comments to type annotations

MegaIng:
@ThatXliner We are not using Type comments. We have separate stub files.

ThatXliner:
> @ThatXliner We are not using Type comments. We have separate stub files.

Oh right. My bad.

I found https://github.com/ambv/retype but it wasn't what I remembered. Hope it helps, though

lark/exceptions.py
MegaIng: Fair enough. Keep `pos_in_stream`

MegaIng (commented):
(I think setup.py also needs to be updated for `py36`)

michael-k:
> Anything else?

Remove

```ini
[bdist_wheel]
universal = 1
```

from setup.cfg.  It's intended for wheels that support both Python 2 and 3.

chanicpanic:
Here is a list of compatibility/deprecated code I have come across in case you want to remove some of it.

- [visitors.py:81](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/visitors.py#L81
): `__visit_tokens__` backwards compatibility
- [lark.py:306](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/lark.py#L306
): Python 2.7 compatibility
- [lark.py:371](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/lark.py#L371
): Old `resolve__antiscore_sum` check
- [lark.py:403](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/lark.py#L403
): Possible deprecation of lexer callbacks
- [exceptions.py:92](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/exceptions.py#L92
): Default value of `use_accepts` is for backwards compatibility
- [exceptions.py:233](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/exceptions.py#L233
): Possible deprecation of `UnexpectedToken.expected`
- [load_grammar.py:819](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/load_grammar.py#L819
): Python 2.7 code?
- [load_grammar.py:1070](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/load_grammar.py#L1070
): Python 2.7 compatibility
- [utils.py:127](https://github.com/lark-parser/lark/blob/be3be1815ead4064e76ce70af70d87088c77c047/lark/utils.py#L127
): Issue with `wraps` in 2.7

erezsh:
Fixes some of the items @chanicpanic mentioned in this PR - https://github.com/lark-parser/lark/pull/993

Regarding 

- `__visit_tokens__` - To be honest I'm not sure why it bothered us. I think it can stay the way it is.

- deprecation of lexer callbacks - I used to think that maybe the `transformer` parameter should be enough, since we added support for tokens. But it's still useful for going over ignored tokens, so unless we let `transformer` run on ignored tokens too, I think we should leave it.

- `UnexpectedToken.expected` - Maybe it's still worth keeping? I have no strong opinion atm

erezsh:
Btw, @chanicpanic & @MegaIng,

I think we should just merge this, and continue work over the `master` branch. Any objections?

MegaIng:
That would mean that we are not going to release any other version before 1.0, not even a `0.12.1` release. But that would be fine with me.

erezsh:
We can always continue working on a 0.12 branch. But I'm not sure that we need to.

MegaIng:
Fair enough. Merge is good IMO.

erezsh:
> Or do we also want to rename to main in the process?

Seems pointless to me

MegaIng:
>Seems pointless to me

Fair enough. I dont have a strong opinion either, but this would be a prime opportunity.

chanicpanic:
No objections to merging from me.

erezsh:
Alright. I'll do it tomorrow then.

erezsh:
Lark `master` is now officially Python 3 only!

PR #926, opened 2021-06-28
Merge .pyi files

chanicpanic:
I started merging the stub files into the code. Here are a couple things I noticed so far:

- In `lexer.py` the signatures of some methods of `TraditionalLexer` and `ContextualLexer` do not match the signatures in `lexer.pyi`.
- `lark.pyi` and `tree.pyi` use `typing.Literal` and `typing.Protocol` which are new in Python 3.8. The [mypy docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module) suggest using the `typing_extensions` package, but it may not be installed. How do we want to approach this?

One other question: Would you like me to add type annotations to more functions where possible, or do you only want them on public interfaces?


ThatXliner:
> * `lark.pyi` and `tree.pyi` use `typing.Literal` and `typing.Protocol` which are new in Python 3.8. The [mypy docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module) suggest using the `typing_extensions` package, but it may not be installed. How do we want to approach this?
> 

My idea:

I feel like we can do something involving strings and `typing.TYPE_CHECKING`. Or we can keep those in `.pyi` files if approach 1 doesn't work.

lark/grammar.py
```diff
@@ -50,11 +52,11 @@ class RuleOptions(Serialize):
     __serialize_fields__ = 'keep_all_tokens', 'expand1', 'priority', 'template_source', 'empty_indices'
 
     def __init__(self, keep_all_tokens=False, expand1=False, priority=None, template_source=None, empty_indices=()):
```
erezsh, line 54: Shouldn't the types be attached to the function parameters?

Does it work the same in the official tooling?

lark/grammar.py
```diff
@@ -50,11 +52,11 @@ class RuleOptions(Serialize):
...
-        self.keep_all_tokens = keep_all_tokens
-        self.expand1 = expand1
-        self.priority = priority
-        self.template_source = template_source
-        self.empty_indices = empty_indices
+        self.keep_all_tokens: bool = keep_all_tokens
```
MegaIng, line 55: I would personally prefer keeping the types out of `__init__` and instead have them on the class level. But we can argue here.

lark/grammar.py
MegaIng: They should be at both places.

erezsh:
I'd rather not add dependencies, even if really basic ones. It might affect the standalone parser.

I think we can keep those exceptions in .pyi files for now, until we find a better solution.

erezsh:
As for adding type annotations, it might make sense in situations where the type isn't obvious. But I prefer if we went through that on a separate PR.

lark/grammar.py
chanicpanic: I believe MegaIng is right. I will add them in both places.

lark/grammar.py
erezsh: I'm against any duplication of information, as much as we can help it.

chanicpanic:
> As for adding type annotations to internal code, it might make sense in situations where the type isn't obvious. But I prefer if we went through that on a separate PR.

Sounds good. I'll only apply the existing types from the .pyi's in this PR.

lark/grammar.py
chanicpanic: I think it makes sense to declare all the instance variables on the class level and keep them in one place. I have noticed a few classes that initialize instance variables outside `__init__` which would make type annotations spread out.

chanicpanic:
An idea I had for `Literal` combining `typing_extensions` and `typing.TYPE_CHECKING`:

```
import sys
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    if sys.version_info >= (3, 8):
        from typing import Literal
    else:
        from typing_extensions import Literal

x: 'Literal["open", "close"]' = "open"
```

However, the same probably won't work for `Protocol` because it is a class that is extended.

erezsh:
What if we only support type checking for 3.8 and up? The code will still run with 3.6

chanicpanic:
Type checking need not be limited to 3.8+. `mypy` lists `typing_extensions` as a dependency, so the new types should be available on whichever version of python the type checking is being run. The challenge is preventing the interpreter from complaining about undefined names for those who do not use type checking. This case is the reason why the type of `x` in the example above is provided as a string literal. It prevents python from trying to evaluate it. 

Upon further inspection, I don't think `PostLex` should be a `Protocol`. Since #859, `PostLex` is an abstract class and thus provides nominal subtyping. So, I think we can use the approach given above for `Literal` and fully merge the stub files.

lark/common.py
```diff
@@ -1,14 +1,29 @@
 from .utils import Serialize
-from .lexer import TerminalDef
+from .lexer import TerminalDef, Token
 
 ###{standalone
+from types import ModuleType
+from typing import Any, Callable, Collection, Dict, Optional, TYPE_CHECKING
 
+if TYPE_CHECKING:
+    from .lark import PostLex
```
MegaIng, line 9: I think we can just move the definition of `PostLex` into  `common.py`. It fits better here anyway.

lark/lark.py
```diff
@@ -23,41 +19,51 @@
 except ImportError:
     regex = None
 
+
+###{standalone
```
MegaIng, line 23: @erezsh This can happen in another PR, but we might want to consider removing type annotations like we are removing docstrings from the standalone parser.

lark/lark.py
```diff
@@ -23,41 +19,51 @@
...
     propagate_positions: Union[bool, str]
     maybe_placeholders: bool
-    lexer_callbacks: Dict[str, Callable[[Token], Token]]
     cache: Union[bool, str]
+    regex: bool
     g_regex_flags: int
+    keep_all_tokens: bool
+    tree_class: Any
+    parser: 'Literal["earley", "lalr", "cyk", "auto"]'
+    lexer: 'Union[Literal["auto", "standard", "contextual", "dynamic", "dynamic_complete"], Type[Lexer]]'
```
MegaIng, line 60: We can instead define types inside a version guarded if-statement that when `Literal` is not available is just `Union[str, Type[Lexer]]` (and similar for all other types here). Not sure if it is better thou.

lark/lexer.py
```diff
@@ -9,12 +10,23 @@
 ###{standalone
 from copy import copy
 
+from types import ModuleType
+from typing import (
```
MegaIng, line 14: We don't need to have duplicates imports inside of `###{standalone`

lark/lexer.py
```diff
@@ -9,12 +10,23 @@
...
+    TypeVar, Type, Tuple, List, Dict, Iterator, Collection, Callable, Optional, FrozenSet, Any,
+    Pattern as REPattern, TYPE_CHECKING
+)
 
-class Pattern(Serialize):
-    raw = None
-    type = None
+if TYPE_CHECKING:
+    from .common import LexerConf
```
MegaIng, line 20: Probably a TODO for later, but I think we can instead of guarding the circle with `TYPE_CHECKING` move more stuff into `common.py`. That is why it exists. ( In this case, the definition of `TerminalDef`/the Patterns/`Token`)

lark/lexer.py
```diff
@@ -289,14 +320,15 @@ def _regexp_has_newline(r):
...
 
-class Lexer(object):
+class Lexer(ABC):
     """Lexer interface
 
     Method Signatures:
         lex(self, text) -> Iterator[Token]
     """
-    lex = NotImplemented
+    lex: Callable[..., Iterator[Token]] = NotImplemented
```
MegaIng, line 331: If we are inheriting from `ABC`, this should just be a normal `abstractmethod`.

lark/lexer.py
```diff
@@ -452,7 +494,7 @@ def __init__(self, conf, states, always_accept=()):
     def make_lexer_state(self, text):
         return self.root_lexer.make_lexer_state(text)
 
-    def lex(self, lexer_state, parser_state):
+    def lex(self, lexer_state: Any, parser_state: Any) -> Iterator[Token]:
```
MegaIng, line 497: Another TODO for potentially later: corrects these annotations

lark/load_grammar.py
```diff
@@ -1072,7 +1082,7 @@ def _unpack_definition(self, tree, mangle):
         return name, exp, params, opts
 
 
-    def load_grammar(self, grammar_text: str, grammar_name: str="<?>", mangle: Callable[[str], str]=None) -> None:
+    def load_grammar(self, grammar_text: str, grammar_name: str="<?>", mangle: Optional[Callable[[str], str]]=None) -> None:
```
MegaIng, line 1085: Maybe I am wrong here, but isn't `Optional` automatically added when `=None` is the default?

MegaIng (commented):
A few comments. Not all of them are really directly solvable problems, but some are more a TODOs for a potentially future PR. I wouldn't be opposed if you decided to 'fix' none of these.

lark/load_grammar.py
chanicpanic: Interesting, I didn't know about that feature. However, it looks like its use is no longer recommended. [https://stackoverflow.com/questions/62732402/can-i-omit-optional-if-i-set-default-to-none](https://stackoverflow.com/questions/62732402/can-i-omit-optional-if-i-set-default-to-none)

lark/lark.py
chanicpanic: I prefer the current way. It provides the precision of `Literal` even when type checking on 3.6 and 3.7.

lark/common.py
erezsh: Why not `lexer.py`?

Btw this is an opportunity to change the postlex interface completely, if we wish to.

lark/lark.py
erezsh: I agree that we should, if we can.

lark/load_grammar.py
MegaIng: Aha. I didn't see that it was deprecated. Good to know.

lark/lark.py
MegaIng: What does the current implementation do when `typing_extensions` is not installed?

chanicpanic: The current implementation executes normally because the import of `typing_extensions` is wrapped in an `if TYPE_CHECKING`. I believe it is reasonable to assume that `typing_extensions` will be installed when `TYPE_CHECKING` is true because `typing_extensions` is a dependency of `mypy`.

MegaIng (approved):

erezsh:
@chanicpanic Thanks a lot! I'm sure it was hard work. Sorry it took me so long to review.

One question, is `__init__(...) -> None` really necessary? It seems meaningless, and it can't really be anything else.

chanicpanic:
@erezsh No problem. I'm glad I can help.

Annotating the return type of `__init__` is optional as long as at least one of its arguments is annotated. The relevant documentation is at [https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods)

erezsh:
I see, thanks for explaining. It's a little surprising, if I already run mypy, why wouldn't I want to scan all the functions by default?

chanicpanic:
I believe mypy behaves this way to make it easier for existing projects to adopt type annotations little by little. Annotations can be applied to the most important or straightforward functions first without receiving a bunch of nasty errors for those that have not been annotated. Also, some functions/variables may be so dynamic that it is hardly worth the trouble to try to declare precise types for them. Thus, you can opt to omit annotations and mypy won't complain.

PR #929, opened 2021-06-29
Use enums instead of strings

ThatXliner:
See #928

TODO:
 - [ ] Update tests
 - [ ] Update documentation


erezsh:
I think we'd like to keep using strings.

But thanks for submitting a PR!

PR #930, opened 2021-06-29
fix tree_matcher when keep_all_tokens=True by setting sym.filter_out …

MegaIng:
…correctly.

Previously `filter_out` might have the wrong value. Inside of tree_matcher (and especially reconstructor) the rule options are not accessible. The easiest fix is to just make sure `sym.filter_out` is correct.

Failing example:

```
start: _A
_A: "a"
```

Wouldn't get reconstructed when `keep_all_tokens=True`

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -630,7 +630,10 @@ class Grammar:
                 else:
                     exp_options = options
 
-                assert all(isinstance(x, Symbol) for x in expansion), expansion
+                for sym in expansion:
+                    assert isinstance(sym, Symbol)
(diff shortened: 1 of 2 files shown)


erezsh:
It's been a while since I touched that code, but why would `filter_out` not be correct there?

MegaIng:
Because at the point where the symbols are created: https://github.com/lark-parser/lark/blob/ec2ba8826ea5f396abab063f47ceaf914333e04c/lark/load_grammar.py#L544 the information is not accessible. And it is never set afterwards. If you know a better place to correct the values, tell me.

erezsh:
Thanks!

Sorry it took me a while to merge it.

PR #932, opened 2021-07-01
Refactor + propagate_positions now accepts a filter predicate.

erezsh:


lark/lexer.py
```diff
@@ -248,40 +251,54 @@ def _create_unless(terminals, g_regex_flags, re_, use_bytes):
             if strtok.priority > retok.priority:
                 continue
             s = strtok.pattern.value
-            m = re_.match(retok.pattern.to_regexp(), s, g_regex_flags)
-            if m and m.group(0) == s:
+            if s == _get_match(re_, retok.pattern.to_regexp(), s, g_regex_flags):
```
MegaIng, line 254: Why? I don't think that function improves readability.

MegaIng (approved):

lark/lexer.py
erezsh: I think it's a slight improvement. But either way, I needed it for portability. I'm writing a standalone generator to Javascript, and the regex api is quite different in js.

erezsh: Hmm maybe I need to explain.. I'm trying to do the generation automaticly from Python, so that it keeps up to that with patches.

MegaIng: Fair enough. Anywhere I can help with that? Seems interesting.

erezsh: How's your Javascript?

This is what I have so far: https://gist.github.com/erezsh/f0745a48f8a7f8a6b6375e798ec7dbfa

I still have a lot of debugging ahead of me, but if you can pinpoint what needs changing, it will save me time. Ideally one for each category, since a fix to the converter will take care of the rest.

You can also help with the converter if you like, but for now there are no pending tasks, which is why debugging the js is necessary.

PR #935, opened 2021-07-05
Bugfix in propagate_positions

erezsh:
Corrected to account for 'container nodes'.

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -40,8 +40,8 @@ class PropagatePositions:
 
             first_meta = self._pp_get_meta(children)
             if first_meta is not None:
-                # meta was already set, probably because the rule has been inlined (e.g. `?rule`)
                 if not hasattr(res_meta, 'line'):
+                    # meta was already set, probably because the rule has been inlined (e.g. `?rule`)
                     res_meta.line = getattr(first_meta, 'container_line', first_meta.line)
                     res_meta.column = getattr(first_meta, 'container_column', first_meta.column)
                     res_meta.start_pos = getattr(first_meta, 'container_start_pos', first_meta.start_pos)
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -30,23 +30,36 @@ class PropagatePositions:
     def __call__(self, children):
         res = self.node_builder(children)
 
-        # local reference to Tree.meta reduces number of presence checks
         if isinstance(res, Tree):
+            # Calculate positions while the tree is streaming, according to the rule:
+            # - nodes start at the start of their first child's container,
+            #   and end at the end of their last child's container.
+            # Containers are nodes that take up space in text, but have been inlined in the tree.
+
             res_meta = res.meta
 
             first_meta = self._pp_get_meta(children)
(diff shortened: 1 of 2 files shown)


lark/parse_tree_builder.py
```diff
@@ -30,23 +30,36 @@ def __init__(self, node_builder, node_filter=None):
...
             res_meta = res.meta
 
             first_meta = self._pp_get_meta(children)
             if first_meta is not None:
-                res_meta.line = first_meta.line
-                res_meta.column = first_meta.column
-                res_meta.start_pos = first_meta.start_pos
-                res_meta.empty = False
+                if not hasattr(res_meta, 'line'):
+                    # meta was already set, probably because the rule has been inlined (e.g. `?rule`)
```
MegaIng, line 44: Can you create a test showcasing this?

MegaIng (changes requested):

lark/parse_tree_builder.py
erezsh: Added.

MegaIng: @erezsh Could you approve my request to use Jetbrains (e.g. my IDE) to access this repo?

tests/test_parser.py
```diff
@@ -94,6 +94,26 @@ def test_propagate_positions(self):
...
+
+        start = g.parse("(t)")
+        a ,= start.children
+        t ,= a.children
+        assert t.children[0] == "t"
+
+        assert t.meta.column == 2
+        assert t.meta.end_column == 3
+
+        assert start.column == a.column == 1
```
MegaIng, line 112: These are deprecated. Also, I would suggest using `self.assertEqual` for better output on failure.

MegaIng (changes requested):

tests/test_parser.py
erezsh: They're not meant to fail ;)

lark/parse_tree_builder.py
erezsh: Approved.

MegaIng: Thank you.

tests/test_parser.py
MegaIng: Then why even test :-P

erezsh: So that when we refactor, or add new features, we know that we didn't break anything. I'm not against using assertEquals, I'm just don't see why it matters. Debugging would usually take a lot longer than replacing the offending assert, or setting a breakpoint on it.

MegaIng: I mean yeah it isn't that important. More important (and why I am not approving with this review) is the usage of `Tree.column`, which is [deprecated](https://github.com/lark-parser/lark/blob/eb2b0b91663ff29a3c2a15e988178b53129d2436/lark/tree.py#L150).

erezsh: Oh okay, I got it now. Funny I didn't get any warnings.

erezsh: Anyway, good catch. Fixed now.

MegaIng (approved):

PR #942, opened 2021-07-13
Updated a few links

### LINKED ISSUE #941: Link leads to 404 page in JSON tutorial.
**Examples link doesn't work**

The link at the bottom of the [JSON tutorial](https://github.com/lark-parser/lark/blob/master/docs/json_tutorial.md) leads to the 404 page.

**To Reproduce**

Just go to the tutorial and click the [link](https://github.com/lark-parser/lark/blob/master/docs/examples) at the very bottom of the page. I thought it might be a good idea to report this so people don't get confused.

MegaIng:
I believe that the changed link from `examples` to `/examples` isn't a problem on readthedocs, but we should check. If it works, this PR fixes #941 .

### DIFF
--- a/docs/json_tutorial.md
+++ b/docs/json_tutorial.md
@@ -427,9 +427,9 @@ I measured memory consumption using a little script called [memusg](https://gist
 | Lark - Earley *(with lexer)* | 42s | 4s | 1167M | 608M |
 | Lark - LALR(1) | 8s | 1.53s | 453M | 266M |
 | Lark - LALR(1) tree-less | 4.76s | 1.23s | 70M | 134M |
-| PyParsing ([Parser](http://pyparsing.wikispaces.com/file/view/jsonParser.py)) | 32s | 3.53s | 443M | 225M |
-| funcparserlib ([Parser](https://github.com/vlasovskikh/funcparserlib/blob/master/funcparserlib/tests/json.py)) | 8.5s | 1.3s | 483M | 293M |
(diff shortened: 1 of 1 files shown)


erezsh:
You can run `make` on the docs folder to build them and see for yourself.

MegaIng:
@erezsh I am not a 100% sure this will work. The question is how readthedocs deals with rooted links. I would in theory expect him to redirect to `latest`, but I don't know if it will do that.

MegaIng:
Ok, I just did find something I can test this with. It probably won't work.

erezsh:
I don't see how restructuring the docs will help?

The only solution I can think of is interface with sphinx-examples and change the way it produces links.

erezsh:
But if it works when you run `make` and view it statically, it will *probably* work on readthedocs.

MegaIng:
Unless readthedocs does page rewriting (which I don't think they do), the final link will be https://lark-parser.readthedocs.io/examples/, which is a 404. It work's locally with a http server served from `_build/html`, but that is not the actual root on RTD.

MegaIng:
But this looks good: https://docs.readthedocs.io/en/latest/user-defined-redirects.html

MegaIng:
Can you try to merge this Pr and rebuild readthedocs and test if it works? If not, I think the best solution is to use an Exact redirect like this:

```
Type: Exact Redirect
From URL: /examples
To URL:   /en/latest/examples
```

While not ideal (on old versions it will still link to the latest), I think it is the best we are going to get.

erezsh:
404

Try your idea.

MegaIng:
@erezsh You need to follow the steps [here](https://docs.readthedocs.io/en/latest/user-defined-redirects.html).

erezsh:
It works. Nice one!

PR #944, opened 2021-07-15
v1.0a (merged recent master)

erezsh:


.travis.yml
```diff
@@ -1,14 +1,12 @@
 dist: xenial
 language: python
 python:
-  - "2.7"
   - "3.4"
```
MegaIng, line 4: Should also be removed

lark/exceptions.py
```diff
@@ -14,7 +12,7 @@ class ConfigurationError(LarkError, ValueError):
     pass
 
 
-def assert_config(value, options, msg='Got %r, expected one of %s'):
+def assert_config(value, options: list, msg='Got %r, expected one of %s'):
```
MegaIng, line 15: This is not a correct typehint. We uses set and `dict.keys()` at multiple point.

MegaIng (approved):

MegaIng (approved):

PR #945, opened 2021-07-22
Docs: Improved documentation of exceptions

erezsh:


lark/lark.py
```diff
@@ -569,6 +571,10 @@ def parse(self, text, start=None, on_error=None):
             If a transformer is supplied to ``__init__``, returns whatever is the
             result of the transformation. Otherwise, returns a Tree instance.
 
+        :raises UnexpectedInput: On a parse error, one of these sub-exceptions will rise:
+                ``UnexpectedCharacters``, ``UnexpectedToken``, or ``UnexpectedEOF``.
+                For convenience, these sub-exceptions also inherit from ``ParserError`` and ``LexerError``.
```
MegaIng, line 576: Maybe add a note that currently some parser raise EOF and some rease Token instead.

MegaIng (approved):

lark/lark.py
erezsh: I think for v1.0 we'll probably want to make them raise the same one, no?

PR #949, opened 2021-07-25
Better repeats

MegaIng:
This Splits up tilde repeats `~mn..mx` up into multiple rules so that it can be used even for large numbers (It ran relatively quickly for 131071 repeats).

Currently not Python2.7 compatible: Should I bother?

I should probably add more comments, please ask at all places where something is unclear.

`small_factors` could probably be improved, but I didn't think to much about it. (Focusing on a pure prime factorization might be better for example), but I am not sure if this will ever make a noticeable difference.

### DIFF
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -373,7 +373,7 @@ def small_factors(n):
 
     Currently, we also keep a + b <= 10, but that might change
     """
-    assert n > 0
+    assert n >= 0
     if n < 10:
         return [(n, 0)]
     # TODO: Think of better algorithms (Prime factors should minimize the number of steps)
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -379,18 +379,12 @@ def small_factors(n):
     Currently, we also keep a + b <= SMALL_FACTOR_THRESHOLD, but that might change
     """
     assert n >= 0
-    if n < SMALL_FACTOR_THRESHOLD:
+    if n <= SMALL_FACTOR_THRESHOLD:
         return [(n, 0)]
     # While this does not provide an optimal solution, it produces a pretty good one.
     # See above comment and PR #949
     for a in range(SMALL_FACTOR_THRESHOLD, 1, -1):
-        b = n % a
-        if a + b > SMALL_FACTOR_THRESHOLD:
-            continue
-        r = n // a
-        assert r * a + b == n  # Sanity check
-        if r <= SMALL_FACTOR_THRESHOLD:
-            return [(r, 0), (a, b)]
-        else:
+        r, b = divmod(n, a)
+        if a + b <= SMALL_FACTOR_THRESHOLD:
             return small_factors(r) + [(a, b)]
     assert False, "Failed to factorize %s" % n
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -361,8 +361,9 @@ def _serialize(value, memo):
     return value
 
 
-# 10 is arbitrarily chosen
-SMALL_FACTOR_THRESHOLD = 10
+# Value 5 keeps the number of states in the lalr parser somewhat minimal
+# It isn't optimal, but close to it. See PR #949
+SMALL_FACTOR_THRESHOLD = 5
 
 
 def small_factors(n):
@@ -380,7 +381,8 @@ def small_factors(n):
     assert n >= 0
     if n < SMALL_FACTOR_THRESHOLD:
         return [(n, 0)]
-    # TODO: Think of better algorithms (Prime factors should minimize the number of steps)
+    # While this does not provide an optimal solution, it produces a pretty good one.
+    # See above comment and PR #949
     for a in range(SMALL_FACTOR_THRESHOLD, 1, -1):
         b = n % a
         if a + b > SMALL_FACTOR_THRESHOLD:
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -187,7 +187,7 @@ def get_regexp_width(expr):
                 return 1, sre_constants.MAXREPEAT
             else:
                 return 0, sre_constants.MAXREPEAT
-            
+
 ###}
 
 
@@ -288,7 +288,7 @@ except ImportError:
 
 class FS:
     exists = os.path.exists
-    
+
     @staticmethod
     def open(name, mode="r", **kwargs):
         if atomicwrites and "w" in mode:
@@ -361,9 +361,13 @@ def _serialize(value, memo):
     return value
 
 
+# 10 is arbitrarily chosen
+SMALL_FACTOR_THRESHOLD = 10
+
+
 def small_factors(n):
     """
-    Splits n up into smaller factors and summands <= 10.
+    Splits n up into smaller factors and summands <= SMALL_FACTOR_THRESHOLD.
     Returns a list of [(a, b), ...]
     so that the following code returns n:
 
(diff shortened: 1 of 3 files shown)


erezsh:
Thanks, it's a nice idea!

> Currently not Python2.7 compatible: Should I bother?

Up to you. But I think it would be very easy to fix.

> I should probably add more comments

The code is a little confusing. I think it would be helpful if you could split it up into smaller functions.

Conceptually, the way I understand it is that you're treating `mn..mx` as `mn + (0..(mx-mn))`. Perhaps it would benefit the code if it was expressed more clearly. Also, I think an explanation of `_add_repeat_opt_rule()` would help.

Another thing, please convert all the "magic numbers" to constants with meaningful names. So, `20` can become `REPEAT_BREAK_THRESHOLD = 20`, or something of that sort, and so on.

MegaIng:
@erezsh Done.

erezsh:
Can you explain why `REPEAT_BREAK_THRESHOLD` is a different value than `SMALL_FACTOR_THRESHOLD`?

My understanding is that both refer to the same magic number, the threshold where generating every combination of rules is more efficient than breaking them up. Am I missing something?

MegaIng:
`REPEAT_BREAK_THRESHOLD` is the threshold at which we start generating rules at all. That behavior might be breaking because of shift/reduce conflicts, so I decide to put it to a higher value that `SMALL_FACTOR_THRESHOLD`. (E.g. the old algorithm works fine from a performance perspective for values till 100). You can disagree and say it should be the same, I am not sure if this is perfect.

erezsh:
> That behavior might be breaking because of shift/reduce conflicts

That is a reason not to lower it from 20. But is there a reason to not raise the "small factor"?

> the old algorithm works fine from a performance perspective for values till 100

Then perhaps it's better to break it at a higher value, say 50?

MegaIng:
I am not sure. We should probably do a few performance tests. (But gotta go now, back in 2 hours)

erezsh:
Okay, no rush. I need to go over the patch more thoroughly anyway. I can't say I fully understand it yet, and it feels like there's potential for off-by-1 errors, or other such hidden corners.

MegaIng:
I am pretty sure there are not off-by-1 errors, but better safe than sorry.

From my performance tests:

- When we split `N` (e.g. the number of repeats) up into the 'factors' `[(a0,b0), (a1, b1), ..., (aN, bN)]`, then the total number of states of the resulting section of the parser will be directly linear dependent on the sum of all as and bs (This is what makes the most sense, but I just wanted to make sure).
- The creation time of the parser is approximately cubic dependent on the number of states, with \~100 states taking a time of \~0.1s.
- The parsing time is for all intense and purposes constant. A very slight trend is visible: less nested grammars are faster, which translates to those with fewer states being slower.

I think that it would be fine to set `REPEAT_BREAK_THRESHOLD` to 50 or maybe even 100, but I wouldn't go over 100.

For `small_factors` I will see if I find a better algorithm considering all possible selections, ignoring any kind of `SMALL_FACTOR_THRESHOLD`, since for large numbers any limit might be to low to find the most efficient solution.

MegaIng:
Actually, it seems like it would be perfectly fine to keep `SMALL_FACTOR_THRESHOLD = 5` and the current implementation. That will always give you something that is at most a few states behind the optimal solution. This makes sense: for all `m,n>2`, `m*n>m+n`, so it is better to split them up instead of multiplying them. (The truly optimal solution sometimes uses a 7, but I haven't found anything higher)

The potential benefit of large `SMALL_FACTOR_THRESHOLD` is that less rules will be generated, but there will be more rules. I am not sure how big the cost of more deeply nested temporary trees is.

(Note that all tests till now assume a fixed repeat, not with range. I will test that now)

MegaIng:
> I am pretty sure there are not off-by-1 errors, but better safe than sorry.

Well, I am biting my own words: I found an off-by-one error.

erezsh:
Well, it *is* one of the two hardest problem in programming.

MegaIng:
Ok, I feel good about this PR now. I added a few timing stats in the code next to the Threshold in case we want to change it. (Or someone else wants to override the value. An assignment to `lark.load_grammar.REPEAT_BREAK_THRESHOLD` works perfectly).

lark/utils.py
```diff
@@ -359,3 +359,38 @@ def _serialize(value, memo):
...
+        return [(n, 0)]
+    # While this does not provide an optimal solution, it produces a pretty good one.
+    # See above comment and PR #949
+    for a in range(SMALL_FACTOR_THRESHOLD, 1, -1):
+        b = n % a
+        if a + b > SMALL_FACTOR_THRESHOLD:
+            continue
+        r = n // a
+        assert r * a + b == n  # Sanity check
+        if r <= SMALL_FACTOR_THRESHOLD:
```
chanicpanic, line 392: Is there a reason why this comparison is `<= SMALL_FACTOR_THRESHOLD` whereas above the comparison is `n < SMALL_FACTOR_THRESHOLD`? I was wondering because I think the function could be simplified by using `n <=` and then removing this if..else statement. I believe that change would only affect the output when `n=SMALL_FACTOR_THRESHOLD`.

MegaIng: The upper one should be `<=`. Thank you. I can refactor a bit more here.

MegaIng: Actually, it didn't make a difference for any value of `n`. But anyway, the function is cleaner now.

MegaIng:
While playing around a bit more, I made a discovery: The slowdown is not actually caused by a large number of states, but by large rules:

| i | factors | factor\_sum | states | creation\_time | parsing\_time |
| :--- | :--- | :--- | :--- | :--- | :--- |
| 0 | \[\(211, 0\)\] | 211 | 213 | 67.35625553131104 | 0.06299972534179688 |
| 1 | \[\(106, 0\), \(1, 105\)\] | 212 | 215 | 9.567001819610596 | 0.06499958038330078 |
| 2 | \[\(71, 0\), \(1, 70\), \(1, 70\)\] | 213 | 217 | 4.13800048828125 | 0.06400084495544434 |
| 3 | \[\(53, 0\), \(1, 53\), \(1, 53\), \(1, 52\)\] | 214 | 219 | 2.36299991607666 | 0.06599950790405273 |
| 4 | \[\(43, 0\), \(1, 42\), \(1, 42\), \(1, 42\), \(1, 42\)\] | 215 | 221 | 1.600999355316162 | 0.06599998474121094 |
| 5 | \[\(36, 0\), \(1, 35\), \(1, 35\), \(1, 35\), \(1, 35\), \(1, 35\)\] | 216 | 223 | 1.1770005226135254 | 0.06500029563903809 |
| 6 | \[\(31, 0\), \(1, 30\), \(1, 30\), \(1, 30\), \(1, 30\), \(1, 30\), \(1, 30\)\] | 217 | 225 | 0.9240310192108154 | 0.06600260734558105 |
| 7 | \[\(27, 0\), \(1, 27\), \(1, 27\), \(1, 26\), \(1, 26\), \(1, 26\), \(1, 26\), \(1, 26\)\] | 218 | 227 | 0.7740023136138916 | 0.06599926948547363 |
| 8 | \[\(24, 0\), \(1, 24\), \(1, 24\), \(1, 24\), \(1, 23\), \(1, 23\), \(1, 23\), \(1, 23\), \(1, 23\)\] | 219 | 229 | 0.6199996471405029 | 0.0670008659362793 |

The total number of states increases, but the total time to generate the grammar goes down since the size of the individual rules goes down. So we actually want to keep SMALL_FACTOR_THRESHOLD  quite small.

erezsh:
I have not done any tests myself, so perhaps I'm wrong, but I would imagine that the performance would be best when the number of rules is somewhat proportional to the average length of the rules.

MegaIng:
No, an average rule size around 10 seems to be pretty optimal, independent of the total rule count. That would indicate that a `SMALL_FACTOR_THRESHOLD ` of around 10/11/12 would be best.

MegaIng:
Further testing revealed that for the actual usecase of `SMALL_FACTOR_THRESHOLD ` a value of `5` is essentially optimal. `10` is 'significantly' (read: 0.1 instead of 0.05) worse.

erezsh:
For reference:
```
REPEAT_BREAK_THRESHOLD:

# For a grammar  of the form start: "A"~0..N, these are the timing stats:
#  N  t
# 10 0.000
# 20 0.004
# 30 0.016
# 40 0.049
# 50 0.109
# 60 0.215
# 70 0.383
# 80 0.631
```

PR #952, opened 2021-07-27
Refactored PR #949 and edited the comments/docstrings

erezsh:
Also now only testing on LALR, to prevent needless compute and save the planet.

For reference (because removed from code):
```
REPEAT_BREAK_THRESHOLD:

# For a grammar  of the form start: "A"~0..N, these are the timing stats:
#  N  t
# 10 0.000
# 20 0.004
# 30 0.016
# 40 0.049
# 50 0.109
# 60 0.215
# 70 0.383
# 80 0.631
```

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2226,6 +2226,35 @@ def _make_parser_test(LEXER, PARSER):
             self.assertRaises((ParseError, UnexpectedInput), l.parse, u'ABB')
             self.assertRaises((ParseError, UnexpectedInput), l.parse, u'AAAABB')
 
+        @unittest.skipIf(PARSER == 'cyk', "For large number of repeats, empty rules might be generated")
+        def test_ranged_repeat_large(self):
+            # Large is currently arbitrarily chosen to be large than 20
+            g = u"""!start: "A"~60
+                """
+            l = _Lark(g)
+            self.assertGreater(len(l.rules), 1, "Expected that more than one rule will be generated")
+            self.assertEqual(l.parse(u'A' * 60), Tree('start', ["A"] * 60))
+            self.assertRaises(ParseError, l.parse, u'A' * 59)
+            self.assertRaises((ParseError, UnexpectedInput), l.parse, u'A' * 61)
+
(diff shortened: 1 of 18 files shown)


tests/test_parser.py
```diff
@@ -2226,6 +2226,35 @@ def test_ranged_repeat_terms(self):
             self.assertRaises((ParseError, UnexpectedInput), l.parse, u'ABB')
             self.assertRaises((ParseError, UnexpectedInput), l.parse, u'AAAABB')
 
+        @unittest.skipIf(PARSER != 'lalr', "We only need to test rule generation, we know BNF is solid on all parsers")
```
MegaIng, line 2229: Then just move it into `test_grammar`

tests/test_parser.py
```diff
@@ -2226,6 +2226,35 @@ def test_ranged_repeat_terms(self):
...
+        def test_ranged_repeat_large(self):
+            # Large is currently arbitrarily chosen to be large than 20
```
MegaIng, line 2231: This is wrong. We also might want to dynamically change the 60 below.

MegaIng (changes requested):
Were the tests really a performance problem for you? For me they ran a lot faster than as example the regex test.

MegaIng:
BTW, if you want to run any of the performance tests for yourself, I uploaded my scripts into [a repo](https://github.com/MegaIng/lark-performance-tests)

tests/test_parser.py
erezsh: Good idea!

MegaIng (approved):

erezsh:
> Were the tests really a performance problem for you? For me they ran a lot faster than as example the regex test.

Well, it added almost 10 seconds, and now it takes 0.3 secs. It wasn't critical to remove, but I also think we should generally avoid long computations when they don't test anything.

erezsh:
Btw, what does you think about the coverage report? Useful or annoying?

MegaIng:
In general good, but I am having a hard time interpreting them sometimes.

PR #956, opened 2021-08-09
Remove ineffective description-file key from setup.cfg

sbraz:
Hi,
setuptools warns that:
> UserWarning: Usage of dash-separated 'description-file' will not be supported in future versions. Please use the underscore name 'description_file' instead

This key doesn't seem to do anything unless you use pbr so I simply dropped it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,6 +5,4 @@ zip_safe=
 universal = 1
 
 [metadata]
-description-file = README.md
 license_file = LICENSE
-


erezsh:
It does seem to be unnecessary.

@MegaIng any objections?

MegaIng:
No objections. It might be worth considering using `long_description: file ...` which is apparently somewhat related/equivalent, but we are already setting `long_description` to a custom value, so I don't think it is needed.

PR #959, opened 2021-08-17
Merge latest master into v1.0

erezsh:


lark/exceptions.py
```diff
@@ -136,10 +137,13 @@ def _format_expected(self, expected):
 
 
 class UnexpectedEOF(ParseError, UnexpectedInput):
```
MegaIng, line 139: Wasn't lalr not raising this exception? Seems also like a good change for 1.0.

lark/lexer.py
```diff
@@ -158,20 +158,20 @@ class Token(str):
 
     def __new__(cls, type_, value, start_pos=None, line=None, column=None, end_line=None, end_column=None, end_pos=None):
         try:
-            self = super(Token, cls).__new__(cls, value)
+            inst = super(Token, cls).__new__(cls, value)
```
MegaIng, line 161: I disagree with this naming change: After getting an instant via `super.__name__`, we are initializing it like a `__init__` would. No reason  not to treat it like that.

MegaIng (approved):

lark/lexer.py
erezsh: Okay. I don't have a strong preference either way.

lark/exceptions.py
erezsh: I agree. But that's for another PR.

lark/lexer.py
erezsh: Actually, let me try to push back by a little -

> we are initializing it like a __init__ would. 

Not quite, in `__init__` there is no need to return self.

It is also quite unusual to assign to self. In other language, where self/this are builtin, that would not be allowed.

MegaIng: Yeah fair enough. I don't feel to strong either.

PR #960, opened 2021-08-17
Remove config for Travis-CI.org

michael-k:
Doesn't work any more; you could switch to travis-ci.com though.

image


erezsh (approved):

MegaIng (approved):
Is there any reason to use travis if we are using github actions? We don't need any platform specific tests right now, and travis-ci seems in general be anti Open Source right now.

erezsh:
@michael-k Thanks!

@MegaIng I can't think of any.

michael-k:
> Is there any reason to use travis if we are using github actions?

Some maintainers prefer to stick with what they know.  That's the only reason why I've mentioned travis-ci.com

> We don't need any platform specific tests right now

GitHub Actions lets you run workflows on Mac OS and Windows.  Travis CI doesn't support Windows (iirc)

> travis-ci seems in general be anti Open Source right now.

Yeah, it all went downhill after they were bought in January 2019 https://www.ideracorp.com/pressreleases/acquires-travis-ci

PR #961, opened 2021-08-17
Run tests against Python 3.10

michael-k:
Stable release is expected on 2021-10-04, see https://www.python.org/dev/peps/pep-0619/#release-schedule


erezsh:
Thanks!

PR #963, opened 2021-08-17
V1.0 merge master2

erezsh:
It seems I forgot to include the latest commits.


erezsh:
Nothing unusual here.

PR #964, opened 2021-08-17
More updates to v1.0

erezsh:
Hopefully it's set up right this time.


erezsh:
The Python3 grammar has been updated for compatibility with Lark.js

lark/exceptions.py
```diff
@@ -18,7 +17,7 @@ class ConfigurationError(LarkError, ValueError):
     pass
 
 
-def assert_config(value, options, msg='Got %r, expected one of %s'):
+def assert_config(value, options: Sequence, msg='Got %r, expected one of %s'):
```
MegaIng, line 20: I think `Collection` is more correct here. a `set` is ok.

lark/indenter.py
```diff
@@ -75,31 +74,31 @@ def always_accept(self):
     @property
     @abstractmethod
     def NL_type(self) -> str:
-        ...
+        return NotImplemented
```
MegaIng, line 77: `raise NotImplementedError`

lark/parser_frontends.py
```diff
@@ -32,20 +32,13 @@ def __init__(self, parser_type, lexer_type):
...
-        return ParsingFrontend(lexer_conf, parser_conf, options)
-
     def deserialize(self, data, memo, lexer_conf, callbacks, options):
         parser_conf = ParserConf.deserialize(data['parser_conf'], memo)
         parser = LALR_Parser.deserialize(data['parser'], memo, callbacks, options.debug)
         parser_conf.callbacks = callbacks
         return ParsingFrontend(lexer_conf, parser_conf, options, parser=parser)
 
-
+    # ... Continued later in the module
```
MegaIng, line 41: Why?

lark/tree.py
```diff
@@ -18,6 +16,9 @@
     else:
         from typing_extensions import Literal
 
+###{standalone
+from collections import OrderedDict
```
MegaIng, line 20: Wasn't the goal to move all imports into `standalone.py`?

MegaIng (changes requested):

lark/parser_frontends.py
erezsh: So it won't be part of the standalone code.

lark/tree.py
erezsh: Not exactly, I was mostly concerned about repetitive imports.

OrderedDict is only used once (and it a little meaningless in Python3, I just didn't think removing it fits into this PR)

lark/exceptions.py
erezsh: You're right.

lark/parser_frontends.py
MegaIng: I would suggest not splitting up the class. Instead move the complete class further down.

lark/tree.py
MegaIng: Fair enough. (Also `OrderedDict` is still needed for `3.6` on non-CPython.)

lark/parser_frontends.py
erezsh: It's also a bit strange to split the class in the middle. It breaks the indentation, and obfuscates the end of the standalone directive.

The only other place in Lark that happens is `tree.Tree`, and I'm not sure that it should stay that way.

But my mind isn't made up. If you think splitting the class using inheritance is too weird, I'll accept it.

PR #965, opened 2021-08-17
Added example using dynamic_complete (Thanks @x_TheGreyHat_x)

MegaIng:
Based on the result of a [gitter discussion](https://gitter.im/lark-parser/Lobby?at=611aafca949a3d73863d0ca7), this showcases how to use `parser='earley', lexer='dynamic_complete', ambiguity='explicit'` to parse just about anything that has a valid structure.

(I had a hard time naming this Example. If you find something better, I am open to suggestions.)

### DIFF
--- a/examples/advanced/dynamic_complete.py
+++ b/examples/advanced/dynamic_complete.py
@@ -1,0 +1,103 @@
+"""
+Using lexer dynamic_complete
+============================
+
+Demonstrates how to use ``lexer='dynamic_complete'`` and ``ambiguity='explicit'``
+
+Sometimes you have data that is highly ambiguous or 'broken' in some sense.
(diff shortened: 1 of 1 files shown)


erezsh:
The code looks good.

But maybe give another look over the docstring?

Also `score()` and `RemoveAmbiguities` could use a short docstring explaining what they are and how they're being used.

MegaIng:
I will add a few more docstrings.

For the module doc: I noticed on false word and a slightly complicated sentence:

```
This examples shows how to parse a json input where the quotes have been
replaced by underscores: ``{_foo_:{}, _bar_: [], _baz_: __}``
Notice that underscores might still appear inside strings, so a potentially
valid reading of the above is:
``` 
Anything else?

Issue #966, opened 2021-08-19
Tree.children annotated with str instead of Token


plannigan:
**Describe the bug**

The [type annotation](https://github.com/lark-parser/lark/blob/82ad512c1aaef202e302e0966230f1080a8c0d26/lark-stubs/tree.pyi#L20) for a `Tree`'s children is `Union[str, Tree]`. However, the actual type is `Union[Token, Tree]`. While it is true that `Token` is a `str`, it causes `mypy` errors when trying to use [Token properties](https://lark-parser.readthedocs.io/en/latest/classes.html#lark.Token).

One possible explanation is that the library wants to obscure away the fact that the real type is `Token` and only consider it a `str`.  However, If I attempt to transform a parsed tree with the intent to write out a slightly altered file, `Reconstructor` fails when it gets a `str` instead of a `Token`.

**To Reproduce**

Example of `mypy` displaying an error for something that is valid.
```python
from lark import Tree

tree = Tree("my_node", children=[])
for child in tree.children:
    if isinstance(child, Tree):
        print(f"is tree: {child.data}")
    else:
        print(f"is token: {child.type}")
```
```shell
$ mypy typing_example.py 
typing_example.py:8: error: "str" has no attribute "type"
Found 1 error in 1 file (checked 1 source file)
```
Using `lark` 0.11.3 & `mypy` 0.910.

Example of `Reconstructor` failure when replacing a token with a `str`. (Only `mypy` error has to do with a different annotation error for `reconstruct()` that has already been fixed.)

```python
from lark import Lark
from lark.reconstruct import Reconstructor

parser = Lark("""
start: WORD WORD

%import common.WORD
%import common.WS_INLINE
%ignore WS_INLINE
""")

tree = parser.parse("Hello World")
tree.children[0] = "Good-bye"

print(Reconstructor(parser).reconstruct(tree))
```

```shell
$ python typing_example.py 
Traceback (most recent call last):
  File "/app/typing_example.py", line 15, in <module>
    print(Reconstructor(parser).reconstruct(tree, postproc=None))
  File "/pyenv/versions/my_venv/lib/python3.9/site-packages/lark/reconstruct.py", line 96, in reconstruct
    for item in x:
  File "/pyenv/versions/my_venv/lib/python3.9/site-packages/lark/reconstruct.py", line 79, in _reconstruct
    unreduced_tree = self.match_tree(tree, tree.data)
[...]

PR #967, opened 2021-08-20
Generic tree typing

plannigan:
* Updates the type information for `Tree` so that it is generic based on the type of the child leaf.
* Adjusts the various classes that operate on trees to also understand the type of trees they can operate on.
* Add `ParseTree` alias so that it is easy to refer to tree produced by the lark parser.

The changes don't introduce any new `mypy` errors, but there is one case where more type information demonstrated an existing inconsistency.

Addresses #966

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -7,6 +7,7 @@ from typing import (
 )
 if TYPE_CHECKING:
     from .parsers.lalr_interactive_parser import InteractiveParser
+    from .tree import ParseTree
     from .visitors import Transformer
     if sys.version_info >= (3, 8):
         from typing import Literal
@@ -591,7 +592,7 @@ class Lark(Serialize):
         """
         return self.parser.parse_interactive(text, start=start)
 
-    def parse(self, text: str, start: Optional[str]=None, on_error: 'Optional[Callable[[UnexpectedInput], bool]]'=None) -> Tree:
+    def parse(self, text: str, start: Optional[str]=None, on_error: 'Optional[Callable[[UnexpectedInput], bool]]'=None) -> 'ParseTree':
         """Parse the given text, according to the options provided.
 
         Parameters:
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -183,17 +183,17 @@ class SlottedTree(Tree):
     __slots__ = 'data', 'children', 'rule', '_meta'
 
 
-def pydot__tree_to_png(tree: Tree[Any], filename: str, rankdir: 'Literal["TB", "LR", "BT", "RL"]'="LR", **kwargs) -> None:
+def pydot__tree_to_png(tree: Tree, filename: str, rankdir: 'Literal["TB", "LR", "BT", "RL"]'="LR", **kwargs) -> None:
     graph = pydot__tree_to_graph(tree, rankdir, **kwargs)
     graph.write_png(filename)
 
 
-def pydot__tree_to_dot(tree: Tree[Any], filename, rankdir="LR", **kwargs):
+def pydot__tree_to_dot(tree: Tree, filename, rankdir="LR", **kwargs):
     graph = pydot__tree_to_graph(tree, rankdir, **kwargs)
     graph.write(filename)
 
 
-def pydot__tree_to_graph(tree: Tree[Any], rankdir="LR", **kwargs):
+def pydot__tree_to_graph(tree: Tree, rankdir="LR", **kwargs):
     """Creates a colorful image that represents the tree (data+children, without meta)
 
     Possible values for `rankdir` are "TB", "LR", "BT", "RL", corresponding to
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -36,6 +36,7 @@ class Meta:
 
 
 _Leaf_T = TypeVar("_Leaf_T")
+Branch = Union[_Leaf_T, 'Tree[_Leaf_T]']
 
 
 class Tree(Generic[_Leaf_T]):
@@ -52,9 +53,9 @@ class Tree(Generic[_Leaf_T]):
     """
 
     data: str
-    children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]'
+    children: 'List[Branch[_Leaf_T]]'
 
-    def __init__(self, data: str, children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]', meta: Optional[Meta]=None) -> None:
+    def __init__(self, data: str, children: 'List[Branch[_Leaf_T]]', meta: Optional[Meta]=None) -> None:
         self.data = data
         self.children = children
         self._meta = meta
@@ -135,7 +136,7 @@ class Tree(Generic[_Leaf_T]):
             kid = self.children[i]
             self.children[i:i+1] = kid.children
 
-    def scan_values(self, pred: 'Callable[[Union[_Leaf_T, Tree[_Leaf_T]]], bool]') -> Iterator[_Leaf_T]:
+    def scan_values(self, pred: 'Callable[[Branch[_Leaf_T]], bool]') -> Iterator[_Leaf_T]:
         """Return all values in the tree that evaluate pred(value) as true.
 
         This can be used to find all the tokens in the tree.
@@ -171,7 +172,7 @@ class Tree(Generic[_Leaf_T]):
     def copy(self) -> 'Tree[_Leaf_T]':
         return type(self)(self.data, self.children)
 
-    def set(self, data: str, children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]') -> None:
+    def set(self, data: str, children: 'List[Branch[_Leaf_T]]') -> None:
         self.data = data
         self.children = children
 
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -7,10 +7,10 @@ except ImportError:
 import sys
 from copy import deepcopy
 
-from typing import List, Callable, Iterator, Union, Optional, TYPE_CHECKING
+from typing import List, Callable, Iterator, Union, Optional, Generic, TypeVar, Any, TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from .lexer import TerminalDef
+    from .lexer import TerminalDef, Token
     if sys.version_info >= (3, 8):
         from typing import Literal
     else:
@@ -35,7 +35,10 @@ class Meta:
         self.empty = True
 
 
-class Tree(object):
+_Leaf_T = TypeVar("_Leaf_T")
+
+
+class Tree(Generic[_Leaf_T]):
     """The main tree class.
 
     Creates a new tree, and stores "data" and "children" in attributes of the same name.
@@ -49,9 +52,9 @@ class Tree(object):
     """
 
     data: str
-    children: 'List[Union[str, Tree]]'
+    children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]'
 
-    def __init__(self, data: str, children: 'List[Union[str, Tree]]', meta: Optional[Meta]=None) -> None:
+    def __init__(self, data: str, children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]', meta: Optional[Meta]=None) -> None:
         self.data = data
         self.children = children
         self._meta = meta
@@ -100,7 +103,7 @@ class Tree(object):
     def __hash__(self) -> int:
         return hash((self.data, tuple(self.children)))
 
-    def iter_subtrees(self) -> 'Iterator[Tree]':
+    def iter_subtrees(self) -> 'Iterator[Tree[_Leaf_T]]':
         """Depth-first iteration.
 
         Iterates over all the subtrees, never returning to the same node twice (Lark's parse-tree is actually a DAG).
@@ -109,17 +112,18 @@ class Tree(object):
         subtrees = OrderedDict()
         for subtree in queue:
             subtrees[id(subtree)] = subtree
-            queue += [c for c in reversed(subtree.children)
+            # Reason for type ignore https://github.com/python/mypy/issues/10999
+            queue += [c for c in reversed(subtree.children)  # type: ignore
                       if isinstance(c, Tree) and id(c) not in subtrees]
 
         del queue
         return reversed(list(subtrees.values()))
 
-    def find_pred(self, pred: 'Callable[[Tree], bool]') -> 'Iterator[Tree]':
+    def find_pred(self, pred: 'Callable[[Tree[_Leaf_T]], bool]') -> 'Iterator[Tree[_Leaf_T]]':
         """Returns all nodes of the tree that evaluate pred(node) as true."""
         return filter(pred, self.iter_subtrees())
 
-    def find_data(self, data: str) -> 'Iterator[Tree]':
+    def find_data(self, data: str) -> 'Iterator[Tree[_Leaf_T]]':
         """Returns all nodes of the tree whose data equals the given data."""
         return self.find_pred(lambda t: t.data == data)
 
@@ -131,7 +135,7 @@ class Tree(object):
             kid = self.children[i]
             self.children[i:i+1] = kid.children
 
-    def scan_values(self, pred: 'Callable[[Union[str, Tree]], bool]') -> Iterator[str]:
+    def scan_values(self, pred: 'Callable[[Union[_Leaf_T, Tree[_Leaf_T]]], bool]') -> Iterator[_Leaf_T]:
         """Return all values in the tree that evaluate pred(value) as true.
 
         This can be used to find all the tokens in the tree.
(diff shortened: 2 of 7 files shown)


lark/tree.py
```diff
@@ -109,17 +112,18 @@ def iter_subtrees(self) -> 'Iterator[Tree]':
         subtrees = OrderedDict()
         for subtree in queue:
             subtrees[id(subtree)] = subtree
-            queue += [c for c in reversed(subtree.children)
+            # Reason for type ignore https://github.com/python/mypy/issues/10999
```
plannigan, line 115: With the current type system it isn't possible to convey that  "`_Leaf_T` can be anything except a `Tree`". So this will likely always be a thing `mypy` won't be able to figure out on its own. The referenced issue has more context and might provide some insights on how to better handle it.

MegaIng (approved):
Very nice :-)

lark/visitors.py
```diff
@@ -220,8 +221,8 @@ def transform(self, tree):
             else:
                 stack.append(x)
 
-        t ,= stack  # We should have only one tree remaining
-        return t
+        result, = stack  # We should have only one tree remaining
+        return result
```
plannigan, line 225: This is the case that is currently inconsistent. Because `self.__visit_tokens__` can cause the second if block to be skipped, it is possible to put `_Leaf_T`instances on the stack which might not be `_Return_T`. Also the specific handling of `Token` makes it unclear if this should only operate on `ParseTree` of if it can work on any arbitrary `Tree`.

MegaIng: The problem here is even harder (and probably impossible) to express. `_Return_T` should in a perfect world be only the 'Top level' return type of `transform`, not of any of the intermediate steps. While those intermediate steps also often conform to that, it might not always be the case.

plannigan: That's a good point. The intermediary steps could return something that is eventually converted into a `_Return_T`. Also based on the dynamic nature of how the user functions are defined & retrieved, it likely won't be possible for `mypy` to guarantee that the final result actually is `_Result_T`.

For the first pass I wanted to see what I could do without actually changing any code. But it might be helpful have the public API methods distinct from the inner workings.

MegaIng: If you can think of a way sure. But for me it seems like this is just an unfortunate truth we have to life with. (A lot of Larks internals can't be correct expressed anyway without dependent typing).

lark/tree.py
```diff
@@ -164,30 +168,32 @@ def iter_subtrees_topdown(self):
...
 
+ParseTree = Tree['Token']
+
 
 class SlottedTree(Tree):
     __slots__ = 'data', 'children', 'rule', '_meta'
 
 
-def pydot__tree_to_png(tree: Tree, filename: str, rankdir: 'Literal["TB", "LR", "BT", "RL"]'="LR", **kwargs) -> None:
+def pydot__tree_to_png(tree: Tree[Any], filename: str, rankdir: 'Literal["TB", "LR", "BT", "RL"]'="LR", **kwargs) -> None:
```
erezsh, line 186: Isn't `Tree` the same as `Tree[Any]`? In the same way that `List` is the same as `List[Any]`?

lark/tree.py
```diff
@@ -35,7 +35,10 @@ def __init__(self):
         self.empty = True
 
 
-class Tree(object):
+_Leaf_T = TypeVar("_Leaf_T")
+
+
+class Tree(Generic[_Leaf_T]):
```
erezsh, line 41: Are there any implication, in terms of speed, or memory consumption (or otherwise), of inheriting from Generic?

lark/visitors.py
```diff
@@ -194,18 +195,18 @@ class Transformer_NonRecursive(Transformer):
     Useful for huge trees.
     """
 
-    def transform(self, tree):
+    def transform(self, tree: Tree[_Leaf_T]) -> _Return_T:
         # Tree to postfix
-        rev_postfix = []
-        q = [tree]
+        rev_postfix: List[Union[_Leaf_T, Tree[_Leaf_T]]] = []
```
erezsh, line 200: Is this necessary? Why not just annotate `result`?

lark/lark.py
```diff
@@ -591,7 +592,7 @@ def parse_interactive(self, text: Optional[str]=None, start: Optional[str]=None)
         """
         return self.parser.parse_interactive(text, start=start)
 
-    def parse(self, text: str, start: Optional[str]=None, on_error: 'Optional[Callable[[UnexpectedInput], bool]]'=None) -> Tree:
+    def parse(self, text: str, start: Optional[str]=None, on_error: 'Optional[Callable[[UnexpectedInput], bool]]'=None) -> 'ParseTree':
```
erezsh, line 595: If you change this to `'Tree[Token]'`, there is no need for a `ParseTree` variable.

erezsh (changes requested):
Thanks for the creative PR!

I have a few questions, since I'm not well versed in Python's type annotations. 

I'm also, in general, a little concerned about the added complication. I can't figure out if it's correct just from a cursory glance. But perhaps that would improve as I get used to it.

lark/tree.py
MegaIng: Very minimal, if at all. Especially Python3.8 and up (I think) really minimize the impact.

lark/visitors.py
plannigan: This specific annotation isn't needed by `mypy`. I think I added it at the start of updating this method to help me keep track of what is happening with the various lists.

Based on the conversation with @MegaIng on the result line, I think I'll need to take a second pass at this whole file now that I have a better understanding.

erezsh:
Sorry it took me so long to review this. 

A few notes

- The phrase `Union[_Leaf_T, Tree[_Leaf_T]]` repeats itself quite a lot, making it hard to read. Why not set `Branch = Union[_Leaf_T, 'Tree[_Leaf_T]']` and use that throughout the code? (preliminary tests show that it should work).

lark/visitors.py
```diff
@@ -52,7 +55,7 @@ def __class_getitem__(cls, _):
         return cls
 
 
-class Transformer(_Decoratable, ABC, Generic[_T]):
+class Transformer(_Decoratable, ABC, Generic[_Return_T, _Leaf_T]):
```
erezsh, line 58: Wouldn't it make more sense if `_Return_T` was the last param? i.e. following an (in, out) pattern.

lark/visitors.py
```diff
@@ -194,18 +203,18 @@ class Transformer_NonRecursive(Transformer):
...
+        q: List[Union[_Leaf_T, Tree[_Leaf_T]]] = [tree]
         while q:
             t = q.pop()
             rev_postfix.append(t)
             if isinstance(t, Tree):
                 q += t.children
 
         # Postfix to tree
-        stack = []
+        stack: List = []
```
erezsh, line 217: Why this addition?

erezsh (changes requested):
Additional notes

lark/visitors.py
plannigan: This was brought up earlier, so I've made that change now.

lark/visitors.py
```diff
@@ -131,11 +134,14 @@ def _transform_tree(self, tree):
...
 
-    def transform(self, tree: Tree) -> _T:
+    def transform(self, tree: Tree[_Leaf_T]) -> _Return_T:
         "Transform the given tree, and return the final result"
         return self._transform_tree(tree)
 
-    def __mul__(self, other: 'Transformer[_T]') -> 'TransformerChain[_T]':
+    def __mul__(
+            self: 'Transformer[Tree[_Leaf_U], _Leaf_T]',
+            other: 'Union[Transformer[_Return_V, _Leaf_U], TransformerChain[_Return_V, _Leaf_U]]'
```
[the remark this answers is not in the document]
plannigan: This is true. `Transformer` and `TransformerChain` have the same public interface. [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) could be used to define that without requiring a base class. (It would just require [typing-extensions](https://pypi.org/project/typing-extensions/) for less than Python 3.8, but that's not a big issue.) I can work on that next.

lark/visitors.py
plannigan: Yeah, so this is a fun one:
* There isn't enough type information based on the `stack.append()` calls for `mypy` to understand what should be allowed to be in the list. So it asks for the initialization of the variable to be annotated.
* However, the list can contain anything based on the implementations of the user functions ([See comment](https://github.com/plannigan/lark/blob/732143a708eeef9e76da992684213deb5ef507df/lark/visitors.py#L233-L235)). So that means that the type is `List[Any]`.
* But the generic `Any` is redundant and can be removed.

So we are left we are left with an annotation that seems to be redundant to humans, but is there to explicitly tell `mypy` how the list is intended to be used.

lark/tree.py
```diff
@@ -52,9 +53,9 @@ class Tree(Generic[_Leaf_T]):
     """
 
     data: str
-    children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]'
+    children: 'List[Branch[_Leaf_T]]'
 
-    def __init__(self, data: str, children: 'List[Union[_Leaf_T, Tree[_Leaf_T]]]', meta: Optional[Meta]=None) -> None:
+    def __init__(self, data: str, children: 'List[Branch[_Leaf_T]]', meta: Optional[Meta]=None) -> None:
```
erezsh, line 58: Sorry if it's obvious, but why do we need to write `Branch[_Leaf_T]` and not just `Branch`, since it's already defined using `_Leaf_T`?

plannigan: Not a problem, it is a subtle point. At the top level `_Leaf_T` stand for some type, but it can still be anything. Once a variable containing a tree is defined as `Tree[Foo]`, then that tree's `_Leaf_T` is bound to `Foo`. Since `Branch` is defined at the top level, it can still be any type. Specifically using `Branch[_Leaf_T]` within the context of a `Tree`, means that the type has to match the bound `_Leaf_T` (in our example `Foo`).

lark/visitors.py
```diff
@@ -9,9 +10,14 @@
...
-_Return_T = TypeVar('_Return_T')
-_Return_V = TypeVar('_Return_V')
+import sys
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
+
+_Return_T = TypeVar('_Return_T', covariant=True)
+_Return_V = TypeVar('_Return_V', covariant=True)
```
erezsh, line 20: Quick question: Why is it Leaf_U but Return_V? Do the U and V stand for something?

plannigan: The signature for the multiplication operator has a lot of moving pieces. It starts with a `ITransformer[_Leaf_T, Tree[_Leaf_U]]` instance (a transformer that converts a `Tree[_Leaf_T`] to `Tree[_Leaf_U]`). Then a `ITransformer[_Leaf_U, _Return_V]` instance (a transformer that converts a `Tree[_Leaf_U`] to `_Return_V`).  Which results in a `ITransformer[_Leaf_T, _Return_V]` (a transformer that converts a `Tree[_Leaf_T`] to `_Return_V`).

There isn't a technical reason why it needs to be named with `_Return_V` instead of `_Return_U`. However, I was thinking that using `_Return_V` would make it clearer that the final type isn't necessarily related to the intermediary `_Leaf_U`. I'm open to switching to `_Leaf_U` or putting some of my above description into a comment in the source code if you think the current state should be changed.

chanicpanic (commented):
In #926 we chose not to add `typing_extensions` as a dependency. A workaround would be to put the import of `Protocol` and the definition of `ITransformer` within an `if TYPE_CHECKING`.

plannigan:
Ok, I didn't have the context of that previous PR when I implemented this change. The one catch that makes this different is that `ITransformer` would only exist within `lark` and is exposed as part of the public API (argument to `__mul__()`). So if a user wanted to annotate a variable as `ITransformer[Foo, Bar]`, they would have to do:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
  from lark.visitors import ITransformer

class Foo:
    pass

class Bar:
    pass

my_tansformer: 'ITransformer[Foo, Bar]'
```

That being said, `ITransformer` is a protocol, so doing that is not a requirement.

This leaves us with two options:
* Use the current `Protocol` implementation with a `TYPE_CHECKING` guard. And document that, while variables can be annotated with `ITransformer` using the example above, the recommendation is to not do that (relying on the functionality of `Protocol`).
* Switch back to the previous `Union` implementation and avoid this potentially confusing corner case at the cost of slightly less readable type annotations in a few places.

I'm open to either option, but have a slight preference to using `Union` as it doesn't introduce the odd corner case.

erezsh:
Is there a way to get the same effect without `Protocol`? Maybe with an ABC somehow?

Anyway, I agree, it doesn't seem worth it to add a dependency just to support ITransformer for 3.6 and 3.7

plannigan:
`ABC` could also work, but would require a inheritance of the base class, something `Transformer` and `TransformerChain` don't currently have. I'll revert back to the `Union` implementation.

lark/visitors.py
```diff
@@ -136,7 +138,7 @@ def transform(self, tree: Tree[_Leaf_T]) -> _Return_T:
         "Transform the given tree, and return the final result"
         return self._transform_tree(tree)
 
-    def __mul__(self, other: 'Transformer[_Return_T, _Leaf_T]') -> 'TransformerChain[_Return_T, _Leaf_T]':
+    def __mul__(self: 'Transformer[Tree[_Leaf_U], _Leaf_T]', other: 'Transformer[_Return_V, _Leaf_U]') -> 'TransformerChain[_Return_V, _Leaf_T]':
```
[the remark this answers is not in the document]
plannigan: Reminder from an earlier conversation:
> I was about to add a new comment about that. I agree that it reads better at `Transformer[_Leaf_T, _Return_T] `. I initially added the leaf type at the end to minimize that changes. However, going from 1 to 2 generic parameters is a breaking change anyways, so moving leaf to be the first type passed in isn't a huge issue.
> 
> Note: These typing changes are only breaking changes for for code already taking advantage of type annotations and the generic types.

Since the PR didn't make it into the 1.0 release, that is something that should be kept in mind. If the choice is made to go the 2.0 route, I remember seeing a couple of places where minor API changes could resolve a few of the mypy errors that are currently reported on the library.

plannigan:
Reminder from an earlier conversation:
> I was about to add a new comment about that. I agree that it reads better at `Transformer[_Leaf_T, _Return_T] `. I initially added the leaf type at the end to minimize that changes. However, going from 1 to 2 generic parameters is a breaking change anyways, so moving leaf to be the first type passed in isn't a huge issue.
> 
> Note: These typing changes are only breaking changes for for code already taking advantage of type annotations and the generic types.

Since the PR didn't make it into the 1.0 release, that is something that should be kept in mind. If the choice is made to go the 2.0 route, I remember seeing a couple of places where minor API changes could resolve a few of the mypy errors that are currently reported on the library.

erezsh:
> If the choice is made to go the 2.0 route

You mean Lark 2.0? That's not going to happen for a while.

> minor API changes could resolve a few of the mypy errors 

Like what?

plannigan:
I understand the desire to not have a 2.0 so shortly after 1.0 was released. But it is unfortunate that this PR did not get merged in time when breaking changes were less of a concern.

I think I should be able to resolve most, maybe all, of the existing smaller mypy issue with smaller API-compatible PRs., But two minor API changes I'd recommend:

1 - Have `TransformerChain` take only two `Transformer`s in `__init__()` instead of N. This would allow for catching:
* An attempt to chain a `Transformer` on after a `Transformer` that produces a non-`Tree` value
* The first `Transformer` produces `Tree[Foo]`, while the second `Transformer` expects a `Tree[Bar]`.
* Would disallow things like `TransformerChain()` or `TransformerChain(my_transformer)` as these don't provide much value.
This wouldn't break any code in the code base. However, it could cause break users that explicitly create `TransformerChain`s instead of using multiplication operator AND passed in a less or more than 2 arguments.)

2 - Replace `Tree.expand_kids_by_index()` with a similar method that accepts a predicate (like `find_pred()` & `scan_values()`) instead of an arbitrary number of indexes. The current API expects the given indexes to correspond to `Tree` instances in the list of children. However, this allows for:
* indexes that aren't `Tree` instances (`AttributeError`)
* invalid indexes (`IndexError`)

The locations in the code base that use this method use it correctly by first doing a list comprehension with an if clause that ensures the index points to a `Tree` instance and satisfies the required criteria. Using a predicate instead would:

* Ensure that non-`Tree` instances are skipped
* Prevent indexing outside of the list.
* Remove an extra iteration currently being done to produce the indexes to expand.
* Simplify the calling code by only requiring a predicate of `Callable[[Tree],bool]`.

A larger API change I'd recommend to be included in a theoretical future 2.0 would be to remove the use of custom double underscore names (ex. `__default__()`) with in the library as [PEP 8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) says:
> `__double_leading_and_trailing_underscore__:` "magic" objects or attributes that live in user-controlled namespaces. E.g. `__init__`,` __import__` or `__file__`. Never invent such names; only use them as documented.

erezsh:
> would be to remove the use of custom double underscore names (ex. __default__()) 

What would you name it instead? (the goal is to avoid name collision with any user-defined rule)

plannigan:
Since "rules whose name begins with an underscore will be inlined into their containing rule", a parse tree won't have any with a single leading underscore. So `_default()` could work. `_default_tree()` would more closely match `_default_token()` and be less generic. It is a bit more verbose, but `_tree_func_not_found()` is more explicit.

I actually ran into an issue with a rule name collision with a Python keyword. I'll open a new issue for that to keep this discussion on topic, but the solution I had in mind would also help resolve the issue with rule names colliding with lark "internal" names.

erezsh:
The main issue with something like `_default()` is that single underscores signify "private members" in Python, so I think many users will find it confusing, that it is accessed from the outside. 

(and `__` means private + protection, but it's rarely used in practice).

In contrast, __magic_method__ syntax is the common way to facilitate interaction with the runtime, like __eq__, __name__, and __iter__. Also other libraries, like rich, use this for similar purposes: https://rich.readthedocs.io/en/stable/protocol.html

erezsh (approved):
Seems good enough for now.
Any improvements to it can be added in separate PRs.

erezsh:
@MegaIng @chanicpanic Anything to add before I merge it?

chanicpanic (approved):

MegaIng (approved):

erezsh:
Thanks @plannigan !  Sorry for being such a stickler :)

PR #970, opened 2021-08-25
Fix recursion error for many options in Terminal

MegaIng:
Missing: tests, and I am 90% sure we can just avoid doing this by removing an earlier step instead of flattening later on.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -600,7 +600,21 @@ def _make_joined_pattern(regexp, flags_set):
     return PatternRE(regexp, flags)
 
 
-class TerminalTreeToPattern(Transformer):
+class FlattenExpansions(Transformer_InPlace):
+    @v_args(tree=True)
+    def expansions(self, tree):
+        i = 0
+        while i < len(tree.children):
+            c = tree.children[i]
+            if isinstance(c, Tree) and c.data == 'expansions':
+                tree.children[i:i+1] = c.children
+                i += len(c.children)
+            else:
+                i += 1
+        return tree
+
+
+class TerminalTreeToPattern(Transformer_NonRecursive):
     def pattern(self, ps):
         p ,= ps
         return p
@@ -670,7 +684,7 @@ class Grammar:
     def compile(self, start, terminals_to_keep):
         # We change the trees in-place (to support huge grammars)
         # So deepcopy allows calling compile more than once.
-        term_defs = deepcopy(list(self.term_defs))
+        term_defs = [(n,(nr_deepcopy_tree(t), p)) for n,(t,p) in self.term_defs]
         rule_defs = [(n,p,nr_deepcopy_tree(t),o) for n,p,t,o in self.rule_defs]
 
         # ===================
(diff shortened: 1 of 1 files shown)


erezsh:
The code can be more elegant (imho) and shorter by Using `Tree.expand_kids_by_index`.

MegaIng:
I was expecting that there is a utility function for that somewhere, but I wanted to just quickly create a fix before going to bed.

Why `kids`? Everywhere else it's named `children`.

Btw, mind adding an extra utility `expand_(kids/children)_by_index`? All cases where `expand_kids_by_index` is used is for this exact purpose.

erezsh:
> Why kids

`expand_children_by_index` just seemed a little too long. Not sure if that's a good reason :)

> mind adding an extra utility

Sorry I don't understand, add what?

MegaIng:
> just seemed a little too long

It feels weird, but it's not that bad. Still descriptive enough.

> mind adding an extra utility

An additional method similar to ´expand_kids_by_index`, that does both the checking the `.data` and the expanding.

https://github.com/lark-parser/lark/blob/8f2bef29bca3aaee6844a861b9f1f2ac4f72d2b2/lark/load_grammar.py#L364

and

https://github.com/lark-parser/lark/blob/8f2bef29bca3aaee6844a861b9f1f2ac4f72d2b2/lark/parse_tree_builder.py#L208

are the only places where `expand_kids_by_index` is used, both of which check for `.data` attribute.

erezsh:
Ah, I see. Something like `expand_children_by_data`?

MegaIng:
Yes

erezsh:
Sure, sounds like a reasonable addition.

Although, tbh I'm not sure any of them belong in `Tree`, since I've never used them outside the context of Lark internals.

Maybe they would be better off as an outside function.

MegaIng:
I was never aware that that functions exists (mostly because I never checked), but I would say that it is useful, and I did implement once or twice.

MegaIng:
Btw, do you know why we can't change the grammar to 

```

    'expansions': ['_expansions'],
    '_expansions': ['alias',
                    '_expansions _OR alias',
                    '_expansions _NL _OR alias'],

```

That would remove the need for any for `FlattenExpansions`, but it doesn't work.

```

lark.exceptions.UnexpectedToken: Unexpected token Token('RULE', 'defaults') at line 4, column 9.
Expected one of: 
	* _OR
```

erezsh:
If I had to guess, I would say it's a conflict between 
```python
    'RULE template_params _COLON expansion _NL',
```
and
```python
                   'expansions _NL _OR alias'],
```

I think that when you add a level of indirection, when it encounters `_NL` it reduces `_expansions` instead of `expansions`, whereas before it was the same action, so it could postpone judgement until the next token.

erezsh:
In other words, you're turning a shift/shift (which is fine) into a shift/reduce, and lark by default chooses to shift instead of reduce. (but even if you flip it, it would just cause `|` on new lines to stop working)

erezsh:
It's possible to restructure the grammar to fix this, but perhaps the simplest solution is to introduce a new terminal: `_NL_OR: _NL _WS* _OR`

PR #973, opened 2021-08-27
Make transformers composable

Drvanon:
In order to be able to allow Transformers to be maintainable over larger projects with lots of imports, it might be useful to be able to merge multiple transformers into one. I present here a minimal implementation for a method on transformers that would be able to do that. 

One issue that I see immediately is that methods on transformers already have a dedicated functionality (manipulate trees), so it would be better to keep the Transformer class as sparse as possible. Maybe instead of adding a method, it should be an independent function, though that would go against the OOP pattern.

### DIFF
--- a/examples/advanced/advanced_transformers.py
+++ b/examples/advanced/advanced_transformers.py
@@ -1,3 +1,14 @@
+"""
+Transformer merging
+==================
+
+This example is intended to show how transformers can be merged in order to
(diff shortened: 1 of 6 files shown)


erezsh:
@Drvanon See PR #976, let me know what you think.

erezsh:
Merged as #976

Issue #977, opened 2021-08-29
Using the LALR file cache on Python 2.7 fails with "unbound method exists() must be called with FS instance as first argument (got str instance instead)"


antoniotejada:
I'm using Lark for epycc https://github.com/antoniotejada/epycc on Python 2.7 and I found that when using the LALR with cache=True:
```Python
parser = lark.Lark.open(grammar_filepath, keep_all_tokens="True", 
    lexer="standard", parser="lalr", cache=True)
```
fails with 
```Python
Exception has occurred: TypeError
unbound method exists() must be called with FS instance as first argument (got str instance instead)
        File "C:\Python27\Lib\site-packages\lark\lark.py", line 276, in __init__
        if FS.exists(cache_fn):
        File "C:\Python27\Lib\site-packages\lark\lark.py", line 501, in open
        return cls(f, **options)
                ...
```
The problem is in utils.FS that doesn't declare "exists" as @staticmethod the same way it's done for "open"

An easy workaround is:
```Python
setattr(lark.utils.FS, 'exists', staticmethod(os.path.exists))
parser = lark.Lark.open(grammar_filepath, keep_all_tokens="True", 
    lexer="standard", parser="lalr", cache=True)
```

erezsh:
Thanks for the catch. We'll merge the fix soon.

Just curious, in what context are you using Lark with Python 2.7?

I ask because we are planning to drop support for Python 2 soon.

antoniotejada:
Thanks for the quick turnaround, Erez.

Coincidentally I read https://github.com/lark-parser/lark/pull/925 and https://github.com/lark-parser/lark/pull/925/commits/eb453241464a2461f9642f6810ac966a4c7e77b7 little after submitting my bug.

The short story is that I don't think I have a solid customer story for you to keep 2.7 support, this is just for my pet projects and I may not even be a returning customer and switch away from Lark at some point :-)

The long story is that I have a bunch of 2.7 code and I cannot be bothered switching to Python 3 unless something I care about gets fixed like GIL/multithread, interpreter reinitialization, speed/JIT. I don't even upgrade packages if they just work.

[...]

PR #980, opened 2021-08-29
Sort Options inside a TerminalTree

MegaIng:
Fixes the problem mentioned at the end of #970 and also inside of #977 (And I think it was mentioned in gitter in the last few days)

Note: This only fixes it for the common case, e.g. a direct listing of multiple options on after the other (e.g. `"a" | "ab"`). It probably doesn't fix it reliable when nesting is involved.

I think the sort key `(-x.max_width, -x.min_width, -len(x.value))` is correct/good enough, but I am not sure.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -614,6 +614,10 @@ class TerminalTreeToPattern(Transformer_NonRecursive):
         if len(exps) == 1:
             return exps[0]
 
+        # Do a bit of sorting to make sure that the longest option is returned
+        # (Python's re module otherwise prefers just 'l' when given (l|ll) and both could match)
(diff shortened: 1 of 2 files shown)


erezsh:
Wouldn't we want to let the users control the order?

There might be some rational for choosing one order over the other, possibly.

MegaIng:
We are currently listing it as a gotcha: https://lark-parser.readthedocs.io/en/latest/grammar.html#regular-expressions-ambiguity

Also, I don't think this is going to affect 99% of the cases. This should only affect your results if you have multiple string literals and the earlier ones are part of the later ones.

It will only affect you negatively if you have a weird reliance on multiple regex literals in a single level of an Terminal where the later ones are partial superset of the earlier ones and you want the earlier ones to match.

PR #982, opened 2021-08-30
ast_utils: Added support for WithMeta (Initial)

erezsh:


lark/ast_utils.py
```diff
@@ -19,15 +19,17 @@ class AsList(object):
...
     """
 
-def camel_to_snake(name):
-    return re.sub(r'(?<!^)(?=[A-Z])', '_', name).lower()
+class WithMeta(object):
+    """Abstract class
 
-def _call(func, _data, children, _meta):
-    return func(*children)
+    Subclasses will be instanciated the Meta instance of the tree. (see ``v_args`` for more detail)
```
MegaIng, line 25: This is not a correct sentence I think.

MegaIng: Also, at what position will it be provided?

MegaIng (changes requested):

lark/ast_utils.py
erezsh: The same position `v_args` does. Which means first argument, but second if using `AsList`.

And yes, it's not ideal. I plan to switch them around for v1.0 so meta is always first.

erezsh: Why isn't it correct?

MegaIng: I think there is at least a `with` missing

erezsh: Right, I missed that :)

Thanks!

MegaIng (approved):

Issue #983, opened 2021-08-30
Matched token missing one character and a few other questions


jjaakko:
I am using Lark 0.11.3 and Python 3.9.6.

1) When running the script below, the value of the FORMULA token will be `\(formula\` while the regular expression for the token is intended to capture a sequence surrounded with backslash and parenthesis such as `\(formula\)` so the captured value is missing the right parenthesis `)`. Why is this? Have I misunderstood something how regexes work with Lark? I tested the regex with https://regex101.com/ and it worked there.

```
from lark import Lark

my_grammar = """
start: thing+

thing: text_or_formula+
text_or_formula: formula|text
formula: FORMULA
text: TEXT_LINE
TEXT_LINE: /((?!\\\()).)+/
FORMULA: /\\\(.+\\\)/

%import common.NEWLINE
"""

text = "text text text \(formula\) text"


def main():
    parser = Lark(
        my_grammar, ambiguity="resolve", parser="earley"
    )
    parse_tree = parser.parse(text)
    print(parse_tree.pretty())
    
    print(parse_tree)


if __name__ == "__main__":
    main()

```
2) If I try to use priorities for rules to simplify the syntax and change the grammar part of the code to the following, everything gets to be picked up as TEXT_LINE although I am trying to specify that when ever the parser could interpret something as FORMULA it should do that. Why the priorities are not working as I expect?

```
my_grammar = """
start: thing+

thing: text_or_formula+
text_or_formula: formula|text
formula.100: FORMULA
text: TEXT_LINE
TEXT_LINE: /.+/
FORMULA: /\\\(.+\\\)/

%import common.NEWLINE
"""
```

3) Any principles, best practices or tips how to maybe make the grammar better? This is the first time I am trying to use Lark.

erezsh:
Regarding (1), it might be due to the need for double-escaping (once for lark, once for the regex)

Try this instead: `FORMULA: "(" /.+/ ")"`

Generally your grammar looks fine. Maybe the rule `text_or_formula` is a bit unnecessary.

I'll try to take a look at (2) tomorrow you don't figure it out by then.

jjaakko:
Thanks for the quick reply! Regarding (1) I tried with your suggestion. It works and makes the syntax more readable! However, I  think the fact that `FORMULA: /\\\(.+\\\)/` does not work _might_ indicate a bug.

[...]

PR #984, opened 2021-08-31
Prefer more specific lexings

chanicpanic:
This PR is an idea I had after looking at #983 to choose the more specific lexing when using Earley's ambiguity resolution.

The motivation may be better understood by an example.

Consider the following:

```python
from lark import Lark

grammar = r"""
start: (group | anything)+
group.1: GROUP
anything: ANYTHING
GROUP: "(" /.+/ ")"
ANYTHING: /.+/
"""

parser = Lark(grammar, lexer='dynamic_complete')
tree = parser.parse("ab(c)de")
print(tree.pretty())
```

While the priority given to `group` forces a derivation in which `group` is matched, there is still unhandled ambiguity. The parser can produce the following trees (and others) nondeterministically.

```
start
  anything      ab
  group (c)
  anything      de
```

```
start
  anything      ab
  group (c)
  anything      d
  anything      e
```

However, the first tree is preferable because it uses a minimal number of `anything` matches. The second has two consecutive `anything`'s which is unnecessary. We tend to like derivations with fewer terminals that match longer strings.

Thus, this PR makes Earley resolve ambiguity by considering rule priorities, and then by choosing derivations with the fewest number of terminals so that the first tree is always returned.

Overall, the change may provide more intuitive behavior for users and reduce the amount of nondeterminism they encounter. As a side bonus, node priorities are no longer restricted to numbers which would allow for more flexibility in custom prioritizers.

I am interested to hear thoughts on the idea and whether it would be a reasonable default.

### DIFF
--- a/lark/parsers/earley_forest.py
+++ b/lark/parsers/earley_forest.py
@@ -143,7 +143,7 @@ class PackedNode(ForestNode):
         ambiguously. Hence, we use the sort order to identify
         the order in which ambiguous children should be considered.
         """
-        return self.is_empty, -self.priority, self.rule.order
+        return self.is_empty, self.priority, self.rule.order
 
     @property
     def children(self):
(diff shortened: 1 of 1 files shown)


MegaIng:
While I agree that something like this will help, this should certainly be configurable: Sometimes, we want the exact opposite behavior, e.g. create as many terminals as possible (see #965)

chanicpanic:
Oh yes, I see how this not always the desired behavior. Perhaps a better approach would be allowing users to pass in a prioritizer (an instance of a `ForestVisitor` subclass). Then we could provide some useful library prioritizers such as this one.

MegaIng:
While that is the most general solution, I would prefer a solution where can just pass a simple score function (similar to the one defined in the other PR), that directly operators on some easy to understand elements (Dream would be a Tree, but that is probably very hard).

chanicpanic:
Right, a Tree would be hard. Plus, part of the advantage of prioritizing the SPPF is the ability to handle ambiguity without needing to construct the entire ambiguous tree. I'll see if I can think of any easier to understand intermediate representations or alternative solutions.

erezsh:
@chanicpanic Now that #991 is merged, should we close this PR?

PR #985, opened 2021-09-03
Stubs improvements

ornariece:
- Added missing `token` attribute to the `UnexpectedToken` stub
- Added a stub for `DedentError`
- Added a stub for `merge_transformers`
- Improved/fixed the type of the `examples` parameter of the `match_examples` method of `UnexpectedInput` (Mapping vs. Dict)


erezsh:
Thanks! Any chance you can rebase these into 1 commit?

ornariece:
> 
> 
> Thanks! Any chance you can rebase these into 1 commit?
i am definitely no expert when it comes to git, but i think i can't because there has been merging commits in-between my commits? please correct me if i'm wrong

erezsh:
`master` is no longer using stubs!

Sorry for the inconvenience, but if you'd like to re-submit these changes as type annotations, we will be happy to accept them.

ornariece:
@erezsh in that case, would you be ok with marking the packages with a ``py.typed`` file to allow static type checking, as per https://www.python.org/dev/peps/pep-0561/?

erezsh:
I think so, as long as there are no adverse effects.

Open a separate issue or PR, and we can discuss it there.

PR #987, opened 2021-09-04
Grammars: Remove use of /i/ in middle of terminals, to support Lark.js

erezsh:


ThatXliner:
Just wondering, why do you have to do this?

erezsh:
@ThatXliner Javascript doesn't support flags in the middle of the regexp. That would mean grammars built using those terminals, wouldn't be able to run in Lark.js

There is no benefit to keep them as flags, other than perhaps an aesthetic preference.

PR #990, opened 2021-09-06
Grammar: Added support for line breaks, using \\

erezsh:


MegaIng (approved):
I am still not sure if this is a good idea. Large rules are inherently slower than smaller rules, and should therefore be avoided. The current system somewhat forces users to use smaller rules.

erezsh:
Why do you say larger rules are inherently slower?

Certainly having a large amount of rules is, but otherwise I don't think that `a b c d` is slower than `ab cd` (where `ab: a b` and `cd: c d`)

It's also useful when you have a lot of imports.
But I suppose it's possible to do it using multiple `%import` statements instead.

MegaIng:
> Why do you say larger rules are inherently slower?

Observation from #949, e.g. slower during construction. But it is probably not relevant unless your rule is literally dozens of expansions. But just merge this. I don't think it is relevant in 99% of cases (and long imports are also a good point).

erezsh:
From the table, it looks like it needs to have over 40 items for the difference to be really noticeable. If someone is going to abuse `\\` that much, they can also abuse the word-wrap feature in their editors :)

Also it shows that longer rules are slightly faster to parse!

Issue #989, opened 2021-09-06
Support Terminal Priorities for Dynamic Earley


chanicpanic:
**Suggestion**
Provide a clear and concise description of what the problem is, and what you would like to happen.

Terminal priorities are currently supported when using a lexer to control the order in which terminals are matched. When using dynamic Earley, the same behavior is not needed as it tries several different lexings. However, in this case, terminal priorities could be used to prefer certain lexings and resolve ambiguity. Additionally, terminal priorities could be used to force minimal or maximal tokens in the chosen tree. (see also #984).

In terms of implementation, terminal priorities could be handled the same way in which rule priorities are handled in Earley. Although, there are a few options:

1. Consider rule and terminal priorities to be equal and resolve ambiguity based on the greatest sum of rule/terminal priorities
2. Resolve ambiguity using rule priorities and use terminal priorities in the case of a tie (like in #984).
3. Resolve ambiguity using terminal priorities and use rule priorities in the case of a tie

**Describe alternatives you've considered**

Option 1. above could be achieved by wrapping individual terminals in rules which have priorities. However, this adds redundancy to the grammar.

erezsh:
Intuitively, I'd say that (1) makes the most sense to me.

It is also possible to emulate either (2) or (3) by scaling the priorities in rules or terminals.

i.e.

1. The rules:terminals priority ratio is 1:1
2. The ratio is 1000:1 (or some other big number)
3. The ration is 1:1000

chanicpanic:
I like the idea of implementing (1) and using ratios to achieve (2) or (3). The implementation of (1) could be done without changing `ForestSumVisitor` simply by giving `Token`'s a priority attribute. Should the emulation of (2) and (3) be provided by an option, or should priority scaling be left to the grammar author?

erezsh:
Well, passing the burden to the grammar author seems reasonable, considering that it's not a very common concern.

Unless you think users might want to switch quickly between these three behaviors?

chanicpanic:
I don't think it's likely that users would want to switch between these behaviors for the same grammar. So, I will just work on an implementation of (1).

PR #991, opened 2021-09-07
Terminal Priorities for Dynamic Earley

chanicpanic:
This is my attempt at supporting terminal priorities for Earley as discussed in #989. A few comments:

- Passing terminal priority information to the parser through the `EarleyRegexpMatcher`is probably not the best approach, but it was a quick solution that did not require changing too many classes.
- Without terminal priority info in the parser constructor, we cannot forgo a forest walk by `ForestSumVisitor`. Although, I don't think the time required to do this walk is significant compared to the time required to construct the tree.
- We may consider changing the default terminal priority to 0 to preserve the current behavior of Earley without terminal priorities. With the default of 1, Earley will choose derivations with the maximal number of tokens.
- The priority options of 'invert' and `None` have not been applied to terminals so far.


erezsh:
I can't wrap my head around why you would need `Token` to contain a priority. Why not just put it next to the regexp?

chanicpanic:
Giving `Token` a priority is helpful for [this](https://github.com/chanicpanic/lark/blob/3754d1671268b2e721181b033291171945ffdd15/lark/parsers/earley_forest.py#L463) part of `ForestSumVisitor` where it reads the priority attribute of a packed node's children. One similar alternative would be creating a new type of leaf `ForestNode`that contains a `Token` and a priority. Ultimately, it means we don't have to perform a type check in `visit_packed_node_out`.

>  Why not just put it next to the regexp?

I don't fully understand what you mean by this, so please expand if you see it as a better solution.

erezsh:
So, looking at the code you linked, I see a reference to `node.rule.options.priority`, so the question is why is there no `node.term.priority` ?

My point is the information is there per terminal type, so copying it for each token seems inefficient. Especially since you can access it with `terminals[token.type].priority`. So I'm asking if there's an algorithmic reason for it, or just that it feels technically easier.

chanicpanic:
At the moment there is no `node.term.priority` because the SPPF contains `Token` objects directly. I could do something like this by creating a `TerminalForestNode`.

There is not an algorithmic reason for it. The issue is that, as far as I know, the `terminals` dict is not available to the Earley parser or forest visitor. I found my approach to be less messy than adding a positional argument to the parser. I am open to suggestions on how to give terminal information to the parser.

erezsh:
Well, the parser gets a term_matcher :

```python
class Parser:
    def __init__(self, parser_conf, term_matcher, resolve_ambiguity=True, debug=False, tree_class=Tree):
```

It could contain the terminals.

chanicpanic:
From `parser_frontends.py`, the term_matcher is a function:

```python
def create_earley_parser__dynamic(lexer_conf, parser_conf, options=None, **kw):
        earley_matcher = EarleyRegexpMatcher(lexer_conf)
        return xearley.Parser(parser_conf, earley_matcher.match, ignore=lexer_conf.ignore, **kw)
```

So could we have the `earley_matcher` contain the terminals and pass the entire object instead of just `match`?

erezsh:
Yes, it's an option. Or we can pass the lexer_conf regardless.

chanicpanic:
I made significant changes to the PR based on passing `lexer_conf` directly to the parser.

tests/test_parser.py
```diff
@@ -1789,6 +1789,57 @@ def test_lexer_prioritization(self):
...
+            start: A B | AB
+            A.2: "a"
+            B: "b"
+            AB: "ab"
+            """
+            l = _Lark(grammar)
+            res = l.parse("ab")
+
+            self.assertEqual(res.children, ['a', 'b'])
+            self.assertNotEqual(res.children, ['ab'])
```
erezsh, line 1806: If this statement failed, it would be a bug in Python, not Lark.

lark/parsers/earley.py
```diff
@@ -37,18 +39,11 @@ def __init__(self, parser_conf, term_matcher, resolve_ambiguity=True, debug=Fals
...
         for rule in parser_conf.rules:
             if rule.origin not in self.predictions:
                 self.predictions[rule.origin] = [x.rule for x in analysis.expand_rule(rule.origin)]
 
-            ## Detect if any rules have priorities set. If the user specified priority = "none" then
-            #  the priorities will be stripped from all rules before they reach us, allowing us to
-            #  skip the extra tree walk. We'll also skip this if the user just didn't specify priorities
-            #  on any rules.
-            if self.forest_sum_visitor is None and rule.options.priority is not None:
-                self.forest_sum_visitor = ForestSumVisitor
```
erezsh, line 50: Wasn't this put here for optimization, for when there were on priorities? Is it not relevant anymore?

erezsh (changes requested):
Looks better!

I added a couple of notes.

Tbh I don't have a good enough understanding of the Earley code, so I have to assume you know what you're doing :)

lark/parsers/earley.py
chanicpanic: Yes, it is an optimization. It can still be applied, but we also need to check that none of the terminals have priorities. However, the default terminal priority is 1 which is a meaningful value to the `ForestSumVisitor`. In other words, if a terminal has a priority of 1, we do not know whether or not it was given a priority in the grammar. If we want to implement this optimization, my suggestion would be change the default terminal priority to 0 (which would also be consistent with the default rule priority of 0).

erezsh: I understand. Then perhaps we should check if the priority != 1 for now?

I agree it would make sense to set both priorities to 0. Unless we actually want to choose derivations with more tokens than rules? I don't really know.

Anyway, we'll be able to change it to 0 in v1.0 which is coming up soon. So maybe we should wait until then?

chanicpanic: Checking priority != 1 is an option. Although, as you mentioned, the default of 1 causes the selection of derivations with more tokens. So either way we are breaking behavior.

MegaIng: Is it breaking? Do we currently specify which one gets selected without any other specification?

chanicpanic: No, we don't specify which one is selected. So, in that sense it is not breaking, but perhaps an unexpected change in behavior.

chanicpanic:
I rebased onto the updated master. In the process I reimplemented the optimization and updated some comments/docs.

erezsh (approved):

lark/parser_frontends.py
```diff
@@ -182,14 +180,14 @@ def match(self, term, text, index=0):
 
 
 def create_earley_parser__dynamic(lexer_conf, parser_conf, options=None, **kw):
-        earley_matcher = EarleyRegexpMatcher(lexer_conf)
-        return xearley.Parser(parser_conf, earley_matcher.match, ignore=lexer_conf.ignore, **kw)
+    earley_matcher = EarleyRegexpMatcher(lexer_conf)
+    return xearley.Parser(lexer_conf, parser_conf, earley_matcher.match, ignore=lexer_conf.ignore, **kw)
```
MegaIng, line 184: If the parser takes `lexer_conf`, we don't need to explicitly provide `lexer_conf.ignore`.

lark/parsers/earley.py
```diff
@@ -42,13 +44,21 @@ def __init__(self, parser_conf, term_matcher, resolve_ambiguity=True, debug=Fals
             if rule.origin not in self.predictions:
                 self.predictions[rule.origin] = [x.rule for x in analysis.expand_rule(rule.origin)]
 
-            ## Detect if any rules have priorities set. If the user specified priority = "none" then
+            ## Detect if any rules/terminals have priorities set. If the user specified priority = "none" then
             #  the priorities will be stripped from all rules before they reach us, allowing us to
```
MegaIng, line 48: `stripped from all rules` and terminals?

MegaIng: Also, it is the value `None`, not the value `"none"`.

MegaIng (changes requested):

lark/parsers/earley.py
chanicpanic: Right now the priority options are not applied to terminals. Is that something I should address in this PR or a separate one?

MegaIng: Depends on how much additional work it is. If it is simple add it here, otherwise we can merge this and then come back.

chanicpanic: It should be pretty simple. I'll add it here.

PR #993, opened 2021-09-12
Various backwards incompatible fixes for v1.0

erezsh:
Mostly obvious changes


MegaIng (approved):
There are a few more scattered across the code base I am sure.

I would suggest keeping track over everything we are doing, especially of everything breaking, in a `CHANGELOG.md` .

erezsh:
Having a changelog is not a bad idea

PR #994, opened 2021-09-12
Token priority is now 0 by default

erezsh:


lark/parser_frontends.py
```diff
@@ -162,7 +162,7 @@ class EarleyRegexpMatcher:
     def __init__(self, lexer_conf):
         self.regexps = {}
         for t in lexer_conf.terminals:
-            if t.priority != 1:
+            if t.priority:
```
MegaIng, line 165: Shouldn't this also be `TOKEN_DEFAULT_PRIORITY `?

erezsh: Probably, yes. But we're going to remove this in a few days

erezsh: I mean, the PR is already there. But I can do it proper anyway, if you think it matters.

MegaIng: if you don't want to make a commit for this, it's fine.

erezsh: Alright. Tbh, even the constant is a bit of an overkill, since we'll probably never change it again. I just didn't want to leave magic numbers in the code, especially since they're so spread out.

Issue #992, opened 2021-09-12
grammar files get opened an unnecessary amount of times, causing an enormous loading time when creating a parser


ornariece:
it seems that, when using a ``FromPackageLoader`` object, a grammar file is opened and read from each time another grammar uses a rule that is imported from that former grammar. this means **opening the same file over and over again**, for each occurence of a rule contained in that file.

while this may not be noticeable for parsers that only use grammar files contained in the same directory (meaning no custom ``FromPackageLoader`` is necessary), it becomes highly problematic when using many ``FromPackageLoader``s, as the time required to construct a parser goes up by an absurd amount.

by placing a ``print(resource_name)`` in the ``get_data()`` function of the python lib ``pkgutil.py``, i was able to count how many times my grammar files were loaded each. for example, the ``common.lark`` grammar provided by lark gets opened 61 (!) times, one of my own grammars 25 times, another 16, etc.

MegaIng:
Can you provided an example grammar(s)? lark should not do more than one import of `common.lark` per grammar file.

ornariece:
i have 46 grammar files, some of which don't use ``common.lark``.


> lark should not do more than one import of `common.lark` per grammar file.

do you mean the ``common.lark`` file gets opened and read from once per grammar file? wouldn't it be better to cache the grammar once and for all?


> Can you provided an example grammar(s)?

let me cook smth up so that i don't have to link my whole project lol

MegaIng:
> wouldn't it be better to cache the grammar once and for all?

Parsing/loading the grammar file is almost certainly not the problem. Also, consider these grammars:

```
// common_rules
name: NAME

%import common.NAME

// a.lark

start_a: "A:" name

%import common_rules.name


// b.lark

start_b: "B:" name

%import common_rules.name


// main.lark

start: start_a | start_b

%import a.start_a
%import b.start_b
```

when you now parse `"A:hello"`, you get the tree `Tree('start', [Tree('start_a', [Tree('a__name', [Token('a__common_rules__CNAME', 'hello')])])])`. Notices how the import path (`main` (no prefix) imports `a` imports `common_rules`) is include in the rule and token names. This would be quite hard if we cached the parsed grammar, for almost no benefit because parsing is so fast.

The slowdown you have is because of many many rules, which is hard to avoid since the prefixes make all those new rules. (e.g. `a__name` and `b__name` are completely distinct rules, each analyzed on their own).

ornariece:
[...]

PR #998, opened 2021-09-14
Load files once

MegaIng:
The changes implemented during the discussion of #992, and a basic implementation of `%include` (typo in commit message :-/). Should I also implement more complex load once stuff in this PR?


ornariece:
@MegaIng say i have 2 grammars A and B that both include a grammar C. now if i include grammars A and B in a grammar D, i get conflict errors as rules defined in grammar C are declared twice. is that wanted?
in fact (not exclusive to %include), why prohibit defining a rule more than once, instead of using the last definition?

MegaIng:
@ornariece Without a major rewrite of `GrammarLoader`, that is not avoidable. I will take `%include` out of this PR, and start on a rework. The rest of this PR works.

The logic is that you can manually specify `%extend` or `%override` which might both be correct behavior for a repeated definition.

MegaIng:
@erezsh This PR is now done. I will put `%include` into a new one.

ornariece:
@MegaIng you added `%include`` back

MegaIng:
HOW? I completely reverted those changes (they aren't even in my local files). I am not sure where git got those changes from...

ornariece:
@MegaIng (currently tweaking my grammars to use ``%include``) 
``%extend`` can't be used on a rule that got defined through an ``%include`` statement (ie. not through direct definition or ``%import``).
the error that is raised is ``lark.exceptions.GrammarError: Can't extend rule name as it wasn't defined before``

MegaIng:
@ornariece That is part of the reason why I removed this implementation of `%include`.

erezsh:
@MegaIng I'm probably missing something, but why not just make it use the same implementation of `%import` but just with an empty namespace?

MegaIng:
@erezsh That leads exactly to the problem @ornariece describes, since imports (and then includes) are added before anything in the grammar is done, meaning that an included grammar could not extend symbols in the main grammar. Also, that would still create a full `GrammarLoader` instance for each included grammar, meaning that no actual import de duplication would happen.

erezsh:
> an included grammar could not extend symbols in the main grammar

I think that's the desirable behavior.

Perhaps these symbols should be refactored into a `shared.lark`, from which they can be extended.

MegaIng:
> I think that's the desirable behavior.

I disagree. Including a grammar should behave no different then just pasting the entire content at that place.

erezsh:
For that behavior, it has to be processed before all other directives.

MegaIng:
@erezsh That is the plan, but that is interdependent of this PR. I will create one for those changes.

erezsh:
Sorry I'm only getting to this now. Any chance you can fix the PR for the recent changes?

It might even need a few more changes for allowing abnf and such? (maybe not. I didn't look that deeply yet)

PR #1000, opened 2021-09-15
Fix for smart_decorator

MegaIng:
- [x] new test to catch edge cases.
- [x] find all edge cases? (@erezsh and @ornariece I would like to know if you find more edge cases)
- [x] rework `smart_decorator`

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -1,8 +1,8 @@
-from typing import TypeVar, Tuple, List, Callable, Generic, Type, Union, Optional
+from typing import TypeVar, Tuple, List, Callable, Generic, Type, Union, Optional, Any
 from abc import ABC
-from functools import wraps
+from functools import wraps, update_wrapper
 
-from .utils import smart_decorator, combine_alternatives
+from .utils import combine_alternatives
 from .tree import Tree
 from .exceptions import VisitError, GrammarError
 from .lexer import Token
@@ -28,7 +28,7 @@ class _Decoratable:
     "Provides support for decorating methods with @v_args"
 
     @classmethod
-    def _apply_decorator(cls, decorator, **kwargs):
+    def _apply_v_args(cls, visit_wrapper):
         mro = getmro(cls)
         assert mro[0] is cls
         libmembers = {name for _cls in mro[1:] for name, _ in getmembers(_cls)}
@@ -41,11 +41,10 @@ class _Decoratable:
                 continue
 
             # Skip if v_args already applied (at the function level)
-            if hasattr(cls.__dict__[name], 'vargs_applied') or hasattr(value, 'vargs_applied'):
+            if isinstance(cls.__dict__[name], _VArgsWrapper):
                 continue
 
-            static = isinstance(cls.__dict__[name], (staticmethod, classmethod))
-            setattr(cls, name, decorator(value, static=static, **kwargs))
+            setattr(cls, name, _VArgsWrapper(cls.__dict__[name], visit_wrapper))
         return cls
 
     def __class_getitem__(cls, _):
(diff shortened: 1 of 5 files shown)


MegaIng:
Ok, this is pretty much finished. The only missing edge case is when you use a staticmethod from one class in another. That is as far as I can tell impossible to differentiate from just a normal function definition. I think this is a complete alternative to #999, and should fix all problems encountered in that area. (Except pickling, but that is a different PR).

tests/test_trees.py
```diff
@@ -171,6 +173,76 @@ def integer2(cls, args):
...
+            # "oc_class_ab_classmethod": ([1, 2], (1, 2)),
+            # "oc_class_ab_staticmethod": ([1, 2], (1, 2)),
+            # "oc_ab_method": ([1, 2], (1, 2)),
+            # "custom_callable": ([1, 2], (1, 2)),
+            "set_union": ([["b"], ["c"]], {"a", "b", "c"}),
+            "static_add": ([1, 2], 3),
+            "partial_reduce_mul": ([[1, 2]], 2),
+        }
+        for method_name, (children, expected_result) in expected.items():
+            result = test_instance.transform(Tree(method_name, children))
```
ornariece, line 253: shouldn't we aslo test the static/class methods with the uninstanciated class? you can have a perfectly functional resulting method but that didn't preserve the staticity

MegaIng: Yes, will add that.

MegaIng:
Can you provide a small example for that?

ornariece:
```python
@v_args(inline=True)
class T(Transformer):
    @staticmethod
    def test():
        pass


@v_args(inline=True)
class T2(Transformer):
    test = T.test


T2.test()
```

MegaIng:
That edge case still exist and is the same one as for non-v_args function, and also the expected behavior from a python standpoint. We can however special case and make it work...

```
class T:
    @staticmethod
    def test():
        pass


class T2:
    test = T.test


T2().test()
```

also doesn't work.

ornariece:
oh right yes, it's the exact same as using a normal function, for which you'd have to redeclare the staticity. i'd keep it that way, as it's what you'd expect from classic python classes

MegaIng:
Yes, It was also a TIL for me, but I wouldn't deviate in that respect from python. The current implementation should now always exactly reflect what python would be doing.

ornariece:
i've run my project (which makes an extensive use of ``v_args`` decorators) with your branch and every output was correct 👍

ornariece:
@erezsh could you review this? it should be ready for merging, unless new edgecases are found

lark/visitors.py
```diff
@@ -402,46 +401,47 @@ def inner(cls, tree):
...
-    f.vargs_applied = True
-    f.visit_wrapper = visit_wrapper
-    return f
+    def __set_name__(self, owner, name):
+        try:
+            f = self.base_func.__set_name__
+        except AttributeError:
+            return
+        else:
+            f(owner, name)
```
erezsh, line 444: This isn't covered by any of the tests

erezsh:
Interesting approach! Definitely seems better.

erezsh:
Also, `examples\advanced\create_ast.py` is failing for this branch.

lark/visitors.py
MegaIng: That is because I don't know when it might be relevant. AFAIK no standard descriptors (except maybe `property`) actually use `__set_name__`. I just added this for completeness. But I will see if I find something.

MegaIng:
What is this error message: `cannot convert dictionary update sequence element #0 to a sequence`? WTF does that mean?

(And `create_ast` was a very simple fix.)

> Definitely seems better

Also, ignoring the tests this is one of those rare PRs that remove more code that they add :-).

MegaIng:
The problem in `dynamic_complete.py` was not caused by this PR, but by the change of the default for `maybe_placeholders`.

lark/visitors.py
MegaIng: I added a test. Not the most useful of situations, but one that breaks without the above code (and never would have worked with the old system).

erezsh:
The test looks okay. Where's the create_ast fix? :)

lark/visitors.py
```diff
@@ -427,7 +427,7 @@ def __init__(self, func: Callable, visit_wrapper: Callable[[Callable, str, list,
     def __call__(self, *args, **kwargs):
         return self.base_func(*args, **kwargs)
 
-    def __get__(self, instance, owner):
+    def __get__(self, instance, owner=None):
```
MegaIng, line 430: This is the `create_ast.py` fix.

erezsh:
Always happy to remove code.

PR #1002, opened 2021-09-18
Fix accepts

### LINKED ISSUE #1001: match_examples failing to match because trying to transform a lark-made Token
``match_examples``, at some point

https://github.com/lark-parser/lark/blob/2335aa63e183c9182d6f8554b0d2d3714fd2286b/lark/exceptions.py#L110

creates a Token, whose value is set to an empty string:

https://github.com/lark-parser/lark/blob/2335aa63e183c9182d6f8554b0d2d3714fd2286b/lark/parsers/lalr_interactive_parser.py#L88

the problem is, this empty string might not be a valid input for whatever transformer method the user defined. in my case, i have a method that uses ``Decimal`` from ``decimal``, and trying to pass an empty string to it results in a ``decimal.InvalidOperation`` being raised. that exception isn't caught by the except statement

https://github.com/lark-parser/lark/blob/2335aa63e183c9182d6f8554b0d2d3714fd2286b/lark/parsers/lalr_interactive_parser.py#L89

and ``match_examples`` fails because of that. with a broader ``except`` statement (say ``except Exception:``), ``mach_examples`` succeeds

MegaIng:
Closes #1001 

Also does a bit of py2 cleanup, and changes the default logger level to `ERROR`. That change is only to keep the output of `.exception` in `Lark.__init__` when loading from cache fails.

### DIFF
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -245,9 +245,16 @@ class UnexpectedToken(ParseError, UnexpectedInput):
             self._accepts = self.interactive_parser and self.interactive_parser.accepts()
         return self._accepts
 
+    @property
+    def _safe_accepts(self):
+        try:
+            return self.accepts
+        except Exception:
+            return self.expected
+
     def __str__(self):
         message = ("Unexpected token %r at line %s, column %s.\n%s"
-                   % (self.token, self.line, self.column, self._format_expected(self.accepts or self.expected)))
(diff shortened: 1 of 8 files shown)


erezsh:
Sorry, this one slipped through the cracks.

Why does accept have to create empty tokens? Can't we just set the string value to "<accepts>", or something of that sort?

PR #1003, opened 2021-09-18
Fix loading from cache

### LINKED ISSUE #997: failure to load grammar from cache
(this is a follow-up on an issue encountered in #992)
```python
import logging

from lark import Lark, logger
from lark.visitors import Transformer, merge_transformers, v_args

logger.setLevel(logging.ERROR)


@v_args(inline=True)
class MergedTransformer(Transformer):
    SOMETHING = str


class MainTransformer(Transformer):
    pass


transformer = merge_transformers(
    MainTransformer(),
    merged=MergedTransformer()
)

parser = Lark(
    'start: "a"',
    parser='lalr',
    transformer=transformer,
    cache='cached_grammar',
    debug=True
)
```
in this minimal example, running the code for a second time (first we need to create the cached grammar) will result in this error being raised:
```python
Failed to load Lark from cache: 'cached_grammar'. We will try to carry on.
  File "C:\Python39\lib\site-packages\lark-1.0.0a0-py3.9.egg\lark\lark.py", line 319, in __init__
    cached_parser_data = pickle.load(f)
AttributeError: 'MergedTransformer' object has no attribute 'str'
```

MegaIng:
Closes #997. The alternative solution via correct setting of `__name__` is not a good idea.

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -3,7 +3,7 @@ import sys, os, pickle, hashlib
 import tempfile
 from typing import (
     TypeVar, Type, List, Dict, Iterator, Callable, Union, Optional,
-    Tuple, Iterable, IO, Any, TYPE_CHECKING
+    Tuple, Iterable, IO, Any, TYPE_CHECKING, Collection
 )
 if TYPE_CHECKING:
     from .parsers.lalr_interactive_parser import InteractiveParser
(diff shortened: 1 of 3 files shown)


erezsh:
Maybe add a test? (We already have a test that mocks the cache, so it shouldn't be a lot of work)

MegaIng:
@erezsh Done.

erezsh:
Thanks!

PR #1006, opened 2021-09-21
Tree Templates (first draft) + py3to2 example

erezsh:


examples/advanced/py3to2.py
```diff
@@ -0,0 +1,93 @@
...
+from lark.tree_templates import TemplateConf, TemplateTranslator
+
+from reconstruct_python import PythonIndenter, PythonReconstructor
+
+
+#
+# 1. Define a Python parser that also accepts template vars in the code (in the form of $var)
+#
+TEMPLATED_PYTHON = r"""
+%import python (single_input, file_input, eval_input, atom, var, stmt, expr, testlist_star_expr, _NEWLINE, _INDENT, _DEDENT, COMMENT, NAME)
```
MegaIng, line 22: Seems like a perfect usecase for an `%include` statement.

examples/advanced/reconstruct_python.py
```diff
@@ -8,10 +8,15 @@
...
 
-from lark import Token
+from lark import Lark, Token
 from lark.reconstruct import Reconstructor
 
-from python_parser import python_parser3
+from python_parser import PythonIndenter
+
+# Official Python grammar by Lark
+python_parser3 = Lark.open_from_package('lark', 'python.lark', ['grammars'],
```
MegaIng, line 17: Do we need `.lark`?

examples/advanced/reconstruct_python.py
```diff
@@ -8,10 +8,15 @@
 
 """
 
-from lark import Token
+from lark import Lark, Token
 from lark.reconstruct import Reconstructor
 
-from python_parser import python_parser3
+from python_parser import PythonIndenter
```
MegaIng, line 14: If we have the python grammar in the lark stdlib, maybe add an compatible PythonIndenter in `lark`?

lark/tree_templates.py
```diff
@@ -0,0 +1,154 @@
...
+
+                res.update(matches)
+
+            return res
+
+
+
+class _ReplaceVars(Transformer):
+    def __init__(self, conf, vars):
+        self.conf = conf
```
MegaIng, line 77: These should also have an `_` underscore

MegaIng (changes requested):
There are a few more things that might be useful, but that should be done in the future. (e.g. some kind of repeated vars).

examples/advanced/py3to2.py
erezsh: I was thinking that too

examples/advanced/reconstruct_python.py
erezsh: What do you mean?

examples/advanced/reconstruct_python.py
erezsh: Good point! Maybe in lark/indenter.py

lark/tree_templates.py
erezsh: You mean `self._conf`?

MegaIng: Yeah. In case the grammar has a rule `conf`

examples/advanced/reconstruct_python.py
MegaIng: Yeah.

examples/advanced/reconstruct_python.py
MegaIng: Do we need to manually add the extension `.lark`

examples/advanced/py3to2.py
erezsh: You didn't submit a PR for this, right?

MegaIng: It is part of #998, but that is not a finished implementation.

erezsh: Hmm okay. Anyway, we can change it to `%include` when it's ready.

MegaIng: Yes.

MegaIng (approved):

PR #1011, opened 2021-10-02
Improved expected tokens returned in UnexpectedToken.

liorgold2:
In some cases, LALR returns a wrong list of expected tokens in the case of `UnexpectedToken`.

For example, processing the input `aaa` in the grammar:

```
start: base base "c"
base: "a" ("b" base)?
```

returns only "c" as a possible replacement for the third "a", while "b" is also a valid option (as in `aabac`)

### DIFF
--- a/lark/parsers/lalr_parser.py
+++ b/lark/parsers/lalr_parser.py
@@ -120,12 +120,18 @@ class ParserState:
         end_state = self.parse_conf.end_state
         callbacks = self.parse_conf.callbacks
 
+        # original_stacks contains a copy of the stacks if they are changed.
+        original_stacks = None
+
         while True:
             state = state_stack[-1]
             try:
                 action, arg = states[state][token.type]
             except KeyError:
                 expected = {s for s in states[state].keys() if s.isupper()}
+                if original_stacks is not None:
+                    # Restore the original state of the stacks
+                    self.state_stack, self.value_stack = original_stacks
                 raise UnexpectedToken(token, expected, state=self, interactive_parser=None)
 
             assert arg != end_state
@@ -138,6 +144,11 @@ class ParserState:
                 return
             else:
                 # reduce+shift as many times as necessary
+
+                if original_stacks is None:
+                    # Make a copy of the stacks before modifying them.
+                    original_stacks = list(state_stack), list(value_stack)
+
                 rule = arg
                 size = len(rule.expansion)
                 if size:
--- a/lark/parsers/lalr_parser.py
+++ b/lark/parsers/lalr_parser.py
@@ -14,14 +14,16 @@ from lark.exceptions import UnexpectedCharacters, UnexpectedInput, UnexpectedTok
 ###{standalone
 
 class LALR_Parser(Serialize):
-    def __init__(self, parser_conf, debug=False):
+    def __init__(self, parser_conf, debug=False, accurate_exceptions=False):
         analysis = LALR_Analyzer(parser_conf, debug=debug)
         analysis.compute_lalr()
         callbacks = parser_conf.callbacks
 
         self._parse_table = analysis.parse_table
         self.parser_conf = parser_conf
-        self.parser = _Parser(analysis.parse_table, callbacks, debug)
+        self.parser = _Parser(
+            parse_table=analysis.parse_table, callbacks=callbacks, debug=debug,
+            accurate_exceptions=accurate_exceptions)
 
     @classmethod
     def deserialize(cls, data, memo, callbacks, debug=False):
@@ -70,9 +72,11 @@ class LALR_Parser(Serialize):
 
 
 class ParseConf:
-    __slots__ = 'parse_table', 'callbacks', 'start', 'start_state', 'end_state', 'states'
+    __slots__ = (
+        'parse_table', 'callbacks', 'start', 'start_state', 'end_state', 'states',
+        'accurate_exceptions')
 
-    def __init__(self, parse_table, callbacks, start):
+    def __init__(self, parse_table, callbacks, start, accurate_exceptions):
         self.parse_table = parse_table
 
         self.start_state = self.parse_table.start_states[start]
@@ -81,6 +85,8 @@ class ParseConf:
 
         self.callbacks = callbacks
         self.start = start
+
+        self.accurate_exceptions = accurate_exceptions
 
 
 class ParserState:
@@ -119,6 +125,7 @@ class ParserState:
         states = self.parse_conf.states
         end_state = self.parse_conf.end_state
         callbacks = self.parse_conf.callbacks
+        accurate_exceptions = self.parse_conf.accurate_exceptions
 
         # original_stacks contains a copy of the stacks if they are changed.
         original_stacks = None
@@ -145,9 +152,10 @@ class ParserState:
             else:
                 # reduce+shift as many times as necessary
 
-                if original_stacks is None:
-                    # Make a copy of the stacks before modifying them.
-                    original_stacks = list(state_stack), list(value_stack)
+                if accurate_exceptions:
+                    if original_stacks is None:
+                        # Make a copy of the stacks before modifying them.
+                        original_stacks = list(state_stack), list(value_stack)
 
                 rule = arg
                 size = len(rule.expansion)
@@ -169,13 +177,14 @@ class ParserState:
                     return value_stack[-1]
 
 class _Parser:
-    def __init__(self, parse_table, callbacks, debug=False):
+    def __init__(self, parse_table, callbacks, debug=False, accurate_exceptions=False):
         self.parse_table = parse_table
         self.callbacks = callbacks
         self.debug = debug
+        self.accurate_exceptions = accurate_exceptions
 
     def parse(self, lexer, start, value_stack=None, state_stack=None, start_interactive=False):
-        parse_conf = ParseConf(self.parse_table, self.callbacks, start)
+        parse_conf = ParseConf(self.parse_table, self.callbacks, start, self.accurate_exceptions)
         parser_state = ParserState(parse_conf, lexer, state_stack, value_stack)
         if start_interactive:
             return InteractiveParser(self, parser_state, parser_state.lexer)
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -197,7 +197,7 @@ class UnexpectedCharacters(LexError, UnexpectedInput):
         message = "No terminal matches '%s' in the current parser context, at line %d col %d" % (self.char, self.line, self.column)
         message += '\n\n' + self._context
         if self.allowed:
-            message += self._format_expected(self.allowed)
+            message += self._format_expected(sorted(self.allowed))
         if self.token_history:
             message += '\nPrevious tokens: %s\n' % ', '.join(repr(t) for t in self.token_history)
         return message
@@ -247,8 +247,9 @@ class UnexpectedToken(ParseError, UnexpectedInput):
         return self._accepts
 
     def __str__(self):
+        expected_str = self._format_expected(sorted(self.accepts or self.expected))
         message = ("Unexpected token %r at line %s, column %s.\n%s"
-                   % (self.token, self.line, self.column, self._format_expected(self.accepts or self.expected)))
+                   % (self.token, self.line, self.column, expected_str))
         if self.token_history:
             message += "Previous tokens: %r\n" % self.token_history
 
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -66,6 +66,7 @@ class LarkOptions(Serialize):
     edit_terminals: Optional[Callable[[TerminalDef], TerminalDef]]
     import_paths: 'List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]'
     source_path: Optional[str]
+    accurate_exceptions: bool
 
     OPTIONS_DOC = """
     **===  General Options  ===**
@@ -136,6 +137,8 @@ class LarkOptions(Serialize):
             A List of either paths or loader functions to specify from where grammars are imported
     source_path
             Override the source of from where the grammar was loaded. Useful for relative imports and unconventional grammar loading
+    accurate_exceptions
+            Provides better errors for the lalr parser, but may result in worse performance.
     **=== End of Options ===**
     """
     if __doc__:
@@ -169,6 +172,7 @@ class LarkOptions(Serialize):
         'use_bytes': False,
         'import_paths': [],
         'source_path': None,
+        'accurate_exceptions': False,
     }
 
     def __init__(self, options_dict):
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -151,7 +151,8 @@ def create_contextual_lexer(lexer_conf, parser, postlex):
 
 def create_lalr_parser(lexer_conf, parser_conf, options=None):
     debug = options.debug if options else False
-    return LALR_Parser(parser_conf, debug=debug)
+    accurate_exceptions = options.accurate_exceptions if options else False
+    return LALR_Parser(parser_conf, debug=debug, accurate_exceptions=accurate_exceptions)
 
 
 create_earley_parser = NotImplemented
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2302,6 +2302,23 @@ def _make_parser_test(LEXER, PARSER):
             self.assertEqual(p.parse("a").children, ['a', None, None])
             self.assertEqual(p.parse("abc").children, ['a', 'b', 'c'])
 
+        @unittest.skipIf(PARSER=='cyk', "Empty rules")
+        def test_expected_tokens(self):
+            grammar = """
+            start: base base "c"
+            base: "a" ("b" base)?
+            """
+
+            l = _Lark(grammar)
+            # The following string should be accepted
+            l.parse("aabac")
+
+            # The following string has an unexpected token
+            with self.assertRaises((UnexpectedToken, UnexpectedCharacters)) as e:
+                l.parse("aaa")
+            # The error message should include B as a possible next token
+            expect_message = "Expected one of: \n\t* B\n\t* C\n"
+            assert str(e.exception).endswith(expect_message)
 
         def test_escaped_string(self):
             "Tests common.ESCAPED_STRING"
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2309,7 +2309,7 @@ def _make_parser_test(LEXER, PARSER):
             base: "a" ("b" base)?
             """
 
-            l = _Lark(grammar)
+            l = _Lark(grammar, accurate_exceptions=True)
             # The following string should be accepted
             l.parse("aabac")
 


lark/parsers/lalr_parser.py
```diff
@@ -120,12 +120,19 @@ def feed_token(self, token, is_end=False):
         end_state = self.parse_conf.end_state
         callbacks = self.parse_conf.callbacks
 
+        # Make a copy of state_stack
+        original_state_stack = list(state_stack)
```
MegaIng, line 124: I don't like unconditionally copying the stacks, but I am not sure if there is a better solution. Maybe at least delay it till the first `Reduce` action?

MegaIng (commented):
Nice catch!

lark/parsers/lalr_parser.py
liorgold2: Changed. Is this better?

MegaIng: It at least reduces the performance hit a bit. But I am still not sure if it is worth it, since it is an unconditional performance hit only for error messages. Can try to run these benchmarks: https://github.com/goodmami/python-parsing-benchmarks for lark?

liorgold2: I'm not sure how to run the benchmarks with my local copy, so what I did is to override lalr_parser.py in the venv directory (+ making sure this file is indeed used).

Before the change:

```
-------------------------------------- benchmark 'arithmetic': 1 tests ---------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     1.8369  1.9139  1.8743  0.0347  1.8891  0.0600       3;0  0.5335       5           1
--------------------------------------------------------------------------------------------------------------

------------------------------------- benchmark 'arithmetic-compile': 1 tests --------------------------------------
Name (time in ms)              Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     9.4613  20.4742  11.0787  2.5286  10.1461  0.4207     10;18  90.2630      98           1
--------------------------------------------------------------------------------------------------------------------

----------------------------------------------- benchmark 'ini': 1 tests -----------------------------------------------
Name (time in ms)              Min       Max      Mean   StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     311.2880  344.4021  332.7991  15.1623  341.6660  24.8145       1;0  3.0048       5           1
------------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'ini-compile': 1 tests -----------------------------------------
Name (time in ms)              Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     9.6849  19.4452  11.0329  2.1035  10.3357  0.2728     10;19  90.6381      96           1
--------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json': 1 tests ------------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     3.5411  3.6372  3.5782  0.0378  3.5741  0.0521       1;0  0.2795       5           1
--------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     15.0438  28.1825  17.5996  3.1443  16.3627  0.4059     10;12  56.8194      62           1
---------------------------------------------------------------------------------------------------------------------

```

After the change:

```
-------------------------------------- benchmark 'arithmetic': 1 tests ---------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     2.0184  2.0845  2.0559  0.0293  2.0669  0.0513       1;0  0.4864       5           1
--------------------------------------------------------------------------------------------------------------

------------------------------------- benchmark 'arithmetic-compile': 1 tests --------------------------------------
Name (time in ms)              Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     9.6690  20.8623  11.5581  2.2187  10.7658  0.3369     12;21  86.5194      93           1
--------------------------------------------------------------------------------------------------------------------

----------------------------------------------- benchmark 'ini': 1 tests -----------------------------------------------
Name (time in ms)              Min       Max      Mean   StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     339.6006  376.2953  357.4534  16.1469  353.9490  29.1214       2;0  2.7976       5           1
------------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'ini-compile': 1 tests -----------------------------------------
Name (time in ms)              Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     9.7191  19.5331  11.6324  1.9314  11.0307  0.2571      9;17  85.9665      91           1
--------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json': 1 tests ------------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     3.9242  3.9995  3.9707  0.0324  3.9854  0.0521       1;0  0.2518       5           1
--------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     16.5646  26.4917  18.4586  2.2015  17.6902  0.4268      7;12  54.1753      58           1
---------------------------------------------------------------------------------------------------------------------

```

Maybe we can do it under a flag (e.g., `accurate_exceptions`)?
This bug is currently preventing us from using the LALR mode, because we need good error messages for the users.

MegaIng: Maybe it can be implemented via a flag, and then an automatic reparsing happens when an error occurs, **and** the users requests to use `.interactive_parser` (this also happens when `.accepts` is accessed to print the error message)?  But I don't think a 10% performance hit for successful execution is worth it. (Thank you for running the benchmark. I think the answer would have been `pip install -e .` to install the local package, but you figured it out anyway.)

liorgold2: I added a flag that's by default False. If someone (like me) cares about the errors enough to lose some performance they can turn it on.

I redid the benchmarks since they were noisy. All the numbers now are higher because I fixed my CPU freq to 2GHz.
By the way, if you really care about performance I think you should remove the asserts in the critical path. See the last benchmark.

Before the change:

```
-------------------------------------- benchmark 'arithmetic': 1 tests ---------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     3.4774  3.5526  3.5111  0.0270  3.5081  0.0255       2;0  0.2848       5           1
--------------------------------------------------------------------------------------------------------------

-------------------------------------- benchmark 'arithmetic-compile': 1 tests --------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     18.8610  33.4660  19.9224  2.7566  19.0908  0.6966       2;3  50.1948      38           1
---------------------------------------------------------------------------------------------------------------------

----------------------------------------------- benchmark 'ini': 1 tests -----------------------------------------------
Name (time in ms)              Min       Max      Mean   StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     577.3088  611.5630  593.7888  13.3596  595.8848  20.0004       2;0  1.6841       5           1
------------------------------------------------------------------------------------------------------------------------

------------------------------------------ benchmark 'ini-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     19.1115  32.0133  20.3093  2.4676  19.3569  1.3748       4;4  49.2386      51           1
---------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json': 1 tests ------------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     6.4420  6.4658  6.4540  0.0092  6.4551  0.0138       2;0  0.1549       5           1
--------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     29.5193  40.9094  31.5422  2.9765  29.9385  2.7600       4;3  31.7035      32           1
---------------------------------------------------------------------------------------------------------------------
```



With a flag:

```
-------------------------------------- benchmark 'arithmetic': 1 tests ---------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     3.6224  3.6333  3.6273  0.0043  3.6268  0.0067       2;0  0.2757       5           1
--------------------------------------------------------------------------------------------------------------

-------------------------------------- benchmark 'arithmetic-compile': 1 tests --------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     19.1647  33.4298  20.5668  2.8158  19.4101  0.9302      4;11  48.6221      50           1
---------------------------------------------------------------------------------------------------------------------

---------------------------------------------- benchmark 'ini': 1 tests ----------------------------------------------
Name (time in ms)              Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     627.5808  636.0194  632.0631  3.2096  631.3121  4.2531       2;0  1.5821       5           1
----------------------------------------------------------------------------------------------------------------------

------------------------------------------ benchmark 'ini-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     19.4110  33.8223  20.4391  2.4277  19.6380  0.3726       4;9  48.9258      41           1
---------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json': 1 tests ------------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     6.6579  6.6870  6.6699  0.0134  6.6634  0.0236       1;0  0.1499       5           1
--------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     29.8766  45.6162  31.5961  3.2599  30.3657  1.1974       4;5  31.6495      32           1
---------------------------------------------------------------------------------------------------------------------
```

Always on (the previous version, without a flag):

```
-------------------------------------- benchmark 'arithmetic': 1 tests ---------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     3.7267  3.7449  3.7380  0.0068  3.7396  0.0063       2;0  0.2675       5           1
--------------------------------------------------------------------------------------------------------------

-------------------------------------- benchmark 'arithmetic-compile': 1 tests --------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     19.3040  32.5226  20.5547  2.6698  19.6383  0.3681       4;9  48.6507      50           1
---------------------------------------------------------------------------------------------------------------------

----------------------------------------------- benchmark 'ini': 1 tests ----------------------------------------------
Name (time in ms)              Min       Max      Mean  StdDev    Median      IQR  Outliers     OPS  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     641.2466  664.4995  648.7045  9.3477  646.7174  10.5931       1;0  1.5415       5           1
-----------------------------------------------------------------------------------------------------------------------

------------------------------------------ benchmark 'ini-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     19.5595  31.3382  20.5078  1.9280  19.8379  0.8130       2;5  48.7620      50           1
---------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json': 1 tests ------------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     6.9633  7.0019  6.9863  0.0175  6.9967  0.0290       1;0  0.1431       5           1
--------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     30.1013  42.2325  32.1045  3.1885  30.5720  2.0026       4;4  31.1483      32           1
---------------------------------------------------------------------------------------------------------------------
```

Removed asserts:

```
-------------------------------------- benchmark 'arithmetic': 1 tests ---------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     3.2740  3.3103  3.2963  0.0140  3.3010  0.0177       1;0  0.3034       5           1
--------------------------------------------------------------------------------------------------------------

-------------------------------------- benchmark 'arithmetic-compile': 1 tests --------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     18.5920  32.9804  19.8039  2.5817  18.8803  0.8268       3;8  50.4952      53           1
---------------------------------------------------------------------------------------------------------------------

---------------------------------------------- benchmark 'ini': 1 tests ----------------------------------------------
Name (time in ms)              Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     565.2981  577.1831  570.2641  4.9078  570.8486  7.7442       2;0  1.7536       5           1
----------------------------------------------------------------------------------------------------------------------

------------------------------------------ benchmark 'ini-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     18.7165  29.9016  19.6981  2.0014  18.9750  0.2352      5;11  50.7662      52           1
---------------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json': 1 tests ------------------------------------------
Name (time in s)             Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
--------------------------------------------------------------------------------------------------------------
test_parse_time[lark]     6.1023  6.1692  6.1446  0.0287  6.1599  0.0442       1;0  0.1627       5           1
--------------------------------------------------------------------------------------------------------------

----------------------------------------- benchmark 'json-compile': 1 tests -----------------------------------------
Name (time in ms)               Min      Max     Mean  StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------
test_compile_time[lark]     28.8005  41.3566  30.8045  3.1636  29.5189  1.9951       3;3  32.4628      29           1
---------------------------------------------------------------------------------------------------------------------
```

erezsh:
I'm not convinced that your solution actually solves all exceptions. It seems more likely to me that you solved your specific case, while still leaving out others that we don't have a test for.

Moreover, this use-case works well when using the contextual lexer. Is there a reason that you can't use it?

Finally, this state-saving behavior can be implemented without changing Lark at all, by using the interactive parser to make your own parsing loop (parse_interactive and so on).

erezsh:
I do appreciate that it would be nice if the LALR parser would return perfect errors that aren't affected by the limitations of the parser. However, that is only possible with Earley.

I think I can see why this particular error is more significant to you than others, but I don't know how many other users will feel the same way.

P.S.

It's probably faster to compute and save the `last_accepts`, instead of the entire last state, which is usually quite bigger.

liorgold2:
@erezsh 

> I'm not convinced that your solution actually solves all exceptions. It seems more likely to me that you solved your specific case, while still leaving out others that we don't have a test for.

What I did works when the LALR interactive parser is used, and it should solve all the cases. The reason is that in order to figure out what are the possible replacement tokens you need to reset to the state before the wrong token and then try running `feed_token` on all of them. This is exactly what `InteractiveParser.accepts` does (except that before the PR it worked on a bad state of the stacks).
By the way, it is possible to add more code before the exception is raised so that it'll work even if the interactive parser is not used.

> Finally, this state-saving behavior can be implemented without changing Lark at all, by using the interactive parser to make your own parsing loop (parse_interactive and so on).

How about moving this code to `InteractiveParser.feed_token`? So that it'll clone the stacks and recover the state if the internal call to `ParserState.feed_token` throws an exception.

> It's probably faster to compute and save the last_accepts, instead of the entire last state, which is usually quite bigger.

I'm not sure exactly that I understand the suggestion, but computing `accepts()` may be expensive because it calls `feed_token` a lot of time + it also clones the stack (multiple times).


Thanks!

erezsh:
The bottom line is that it's a big performance cost, for a very minor improvement.

I think such a PR might make more sense if it took a wider point of view, for example, adding a `backtrack=n` method, which saves the last *n* states, and allows backtracking in case of error. And then, in case it fails, it can backtrack to provide the right error.

Otherwise, I'm not much inclined to add another switch especially for such a specific use, especially since, as I said, you don't need to change Lark in order to easily implement what you need.

> computing accepts() may be expensive because it calls feed_token

Yes, you're probably right about that.

PR #1013, opened 2021-10-04
Transformer_NonRecursive should drop nodes when Discard is raised

### LINKED ISSUE #1012: Transformer_NonRecursive fails on Discard
**Describe the bug**

As documented, `Transformer` catches any `Discard` exceptions and removes the node from the tree. 

I expected `Transformer_NonRecursive` to behave the same way - instead it lets the exception propagate eventually crashing the program.

**To Reproduce**

```
tree = Tree('start', [
	Tree('add', [Token('N', '1'), Token('N', '2'), Token('token_ignore', '4')]),
	Tree('add', [Token('N', '3'), Token('N', '4')]),
	Tree('ignore_tree', [Token('DO', 'NOT PANIC')]),
])

for base in (Transformer, Transformer_InPlace, Transformer_NonRecursive, Transformer_InPlaceRecursive):
	class T(base):
		def add(self, children):
			return sum(children)

		def N(self, token):
			return int(token)

		def ignore_tree(self, children):
			raise Discard

		def token_ignore(self, token):
                    raise Discard
```

adbancroft:
Fixes #1012

### DIFF
--- a/tests/test_trees.py
+++ b/tests/test_trees.py
@@ -375,7 +375,11 @@ class TestTrees(TestCase):
         self.assertEqual(x, t2)
 
     def test_transformer_variants(self):
-        tree = Tree('start', [Tree('add', [Token('N', '1'), Token('N', '2')]), Tree('add', [Token('N', '3'), Token('N', '4')])])
+        tree = Tree('start', [
+            Tree('add', [Token('N', '1'), Token('N', '2'), Token('token_ignore', '4')]),
+            Tree('add', [Token('N', '3'), Token('N', '4')]),
+            Tree('ignore_tree', [Token('DO', 'NOT PANIC')]),
+            ])
         for base in (Transformer, Transformer_InPlace, Transformer_NonRecursive, Transformer_InPlaceRecursive):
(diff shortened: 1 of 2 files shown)


tests/test_trees.py
```diff
@@ -375,7 +375,11 @@ def b(cls, args):
         self.assertEqual(x, t2)
 
     def test_transformer_variants(self):
-        tree = Tree('start', [Tree('add', [Token('N', '1'), Token('N', '2')]), Tree('add', [Token('N', '3'), Token('N', '4')])])
+        tree = Tree('start', [
+            Tree('add', [Token('N', '1'), Token('N', '2'), Token('token_ignore', '4')]),
```
MegaIng, line 379: Please make token names uppercase

MegaIng (changes requested):

MegaIng (approved):

tests/test_trees.py
adbancroft: Done

erezsh:
Thanks!

Issue #318, opened 2019-02-04
Construct a Lark grammar from ABNF format (RFC 5234)


Zac-HD:
[RFC 5234](https://tools.ietf.org/html/rfc5234) describes the standard grammar format for internet standards, such as the notoriously-hard-to-validate email addresses.

Because this standard is a dialect of EBNF and does not allow for embedded code, it should be *relatively* easy to construct a Lark object for a given ABNF grammar - at least easier than converting from Nearley!  Hopefully it's easy enough that runtime conversion in a new `Lark.from_abnf` method (or group of methods) would be practical.

This feature request is based on HypothesisWorks/hypothesis#170, where I eventually realized that parsing ABNF was going to be easier as well as more widely useful upstream.  I'd be happy to work on this with some guidance about where to start, and have already translated the grammar of ABNF from ABNF to Lark's format.

erezsh:
I think it's a nice idea. I think your best bet is to look create a Lark parser that reproduces the output of the GrammarLoader parser : https://github.com/lark-parser/lark/blob/master/lark/load_grammar.py#L663

You might have to do some post-processing (for example, in a Transformer) to make them a perfect fit.

Once you have that working, I'll add an interface to plug it in there when called with ABNF grammar.

While using `Lark.from_abnf` isn't bad (and can be complemented with `Lark.open_abnf`), how about doing this instead?
```python
parser = Lark("... grammar ... ", syntax='abnf')
```
That opens the door to adding other formats in the future.

davaya:
The output of GrammarLoader is an instance of the Grammar class with `rule_defs`, `term_defs` and `ignore` variables.   Is there a "GrammarSave" function that takes a Grammar instance and produces a Lark file or returns a string?

It will be necessary to convert an ABNF file to a Lark file, because there are inevitably features of Lark that are not supported in ABNF.  The workflow would be 1) convert ABNF to a Lark file, then 2) tweak the Lark file to achieve the desired results.  The GrammarSave function is needed both to develop an ABNF loader in the first place and to do grammar optimization once it is available.

erezsh:
@davaya 

I don't understand what the difficulty is, and what you're trying to do to solve it.

> convert an ABNF file to a Lark file

Why does it have to go through files?

> there are inevitably features of Lark that are not supported in ABNF

So give Lark a default (like an empty list, or whatever if appropriate)

davaya:
[...]

PR #1017, opened 2021-10-12
ABNF grammar support

t-higuchi:
This PR implements feature request #318

As proposed in #318,  'syntax' option need to be set to load ABNF grammar: 
  `parser = Lark("... grammar ... ", syntax='abnf')`

It accepts ABNF grammar described in RFC5234 and RFC7405.
Non-standard extension to ABNF is kept minimal. Only '%import' directive is implemented. 

Please see an example (url_parser_abnf.py ) for ABNF-specific decorator usage and post-processing to parse tree.

Some implementation notes:
- ABNFGrammar.compile() is implemented so that it outputs `terminals` and `compiled_rules`  
  just as Grammar.compile() does for lark's EBNF grammar. This approach makes minimal modifications to existing code base.
- A few modifications to existing code is needed since python complains if hyphens are used in ABNF rule names.

### DIFF
--- a/examples/grammars/rfc3986.abnf
+++ b/examples/grammars/rfc3986.abnf
@@ -1,0 +1,87 @@
+; ABNF grammar from RFC3986
+;   Uniform Resource Identifier (URI): Generic Syntax
+;
+; some terminals (e.g. DIGIT, ALPHA, ..) is defined in ABNF core rules in RFC5234.
+;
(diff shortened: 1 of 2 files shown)


erezsh:
@MegaIng I don't mind doing that refactor myself. I'll try to get to it this week.

PR #1018, opened 2021-10-13
Refactor load_grammar to use Terminal/NonTerminal throughout the process

erezsh:
It took a bit of refactoring, but I think I got it done without too many changes.

Maybe I'm missing something, but apparently it's not necessary to change mangle, aliases, or the rest.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -1138,8 +1138,9 @@ class GrammarBuilder:
                     item ,= t2.children
                     if item.data == 'value':
                         item ,= item.children
-                        if isinstance(item, Token) and item.type == 'TERMINAL':
-                            self._ignore_names.append(item.value)
+                        if isinstance(item, Terminal):
+                            # Keep terminal name, no need to create a new definition
+                            self._ignore_names.append(item.name)
                             return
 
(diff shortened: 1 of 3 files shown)


erezsh:
(I didn't move anything to a new module yet, so the diff will be easy to read)

erezsh:
@MegaIng Looks like both the tests and I missed two glaring bugs in the refactor, so if you have the time to give it another look, it's probably a good idea.

Maybe it's impossible to catch everything. But I do wonder if there's a way to do it better, so we won't have to find them by "luck".

erezsh:
@MegaIng You were right about %ignore. I guess coverage is good for something.

I also found another problem that was hidden for a long time.

MegaIng:
Yeah, I can do that. Btw, do we want abnf as a default packaged grammar or as a plugin/example?

erezsh:
I'm not sure. I'm a little worried about bloat, but maybe ABNF is standard enough to warrant being built-in.

What do you think?

erezsh:
I don't see the harm in a few more tests and examples. But I agree additions to the actual library should be minimal.

PR #1019, opened 2021-10-15
Added parse_lark_grammar.py and the syntax parameter/registry

MegaIng:
I think the core design here stands. I would like feedback on that already.

Primarily missing here is tests, examples and docs.


erezsh:
Oops sorry. I meant `<string>`

MegaIng:
That reminds me, we should also because of this start loading files with different extensions.

PR #1028, opened 2021-11-03
Fix `python.number` pattern

0dminnimda:
Python doesn't accept numbers with the `_` in the beginning/end and numbers with more than one `_` in the allowed places:
```python
>>> 69420
69420
>>> 69_420
69420
>>> 69__420
  File "<stdin>", line 1
    69__420
      ^
SyntaxError: invalid decimal literal
>>> 69_420_
  File "<stdin>", line 1
    69_420_
          ^
SyntaxError: invalid decimal literal
>>> _69_420
  File "<stdin>", line 1, in <module>
NameError: name '_69_420' is not defined
```

```python
>>> 03.1415
3.1415
>>> 0_3.14_15
3.1415
>>> 0__3.14_15
  File "<stdin>", line 1
    0__3.14_15
     ^
SyntaxError: invalid decimal literal
>>> 0_3.14__15
  File "<stdin>", line 1
    0_3.14__15
          ^
SyntaxError: invalid decimal literal
>>> 0_3.14_15_
  File "<stdin>", line 1
    0_3.14_15_
             ^
SyntaxError: invalid decimal literal
>>> 0_3._14_15
  File "<stdin>", line 1
    0_3._14_15
       ^
SyntaxError: invalid decimal literal
>>> 0_3_.14_15
  File "<stdin>", line 1
    0_3_.14_15
       ^
SyntaxError: invalid decimal literal
>>> _0_3.14_15
  File "<stdin>", line 1
    _0_3.14_15
    ^^^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?
```

the same goes with complex numbers. And yes, python recognizes `_xxx` as a name, even though `x` is a digit, but it's still not a number, so this doesn't affect us.

The current implementation only filters numbers with `_` in the beginning, so here's the fix for the other cases.

---

Hopefully, we still can make backward-incompatible changes, so it's fine to change `IMAG_NUMBER` to `COMPLEX_NUMBER`

I also tested `\d(?:_?\d+)*` for `DEC_NUMBER` but haven't seen any significant performance changes (everything is within the normal range, considering that I was not using a stable environment).


erezsh:
Thanks! I'll have a look.

> Hopefully, we still can make backward-incompatible changes

We don't need to be backward-compatible with bugs :)

lark/grammars/python.lark
```diff
@@ -228,14 +228,15 @@ COMMENT: /#[^\n]*/
...
 LONG_STRING: /[ubf]?r?(""".*?(?<!\\)(\\\\)*?"""|'''.*?(?<!\\)(\\\\)*?''')/is
 
-DEC_NUMBER: /0|[1-9][\d_]*/i
+DEC_NUMBER: /0|\d(?:_?\d)*/i
 HEX_NUMBER.2: /0x[\da-f]*/i
 OCT_NUMBER.2: /0o[0-7]*/i
-BIN_NUMBER.2 : /0b[0-1]*/i
-FLOAT_NUMBER.2: /((\d+\.[\d_]*|\.[\d_]+)([Ee][-+]?\d+)?|\d+([Ee][-+]?\d+))/
-IMAG_NUMBER.2: /\d+[Jj]/ | FLOAT_NUMBER /[Jj]/
+BIN_NUMBER.2: /0b[0-1]*/i
```
MegaIng, line 234: This pattern is also wrong, same for `OCT` and `HEX` as well.

lark/grammars/python.lark
```diff
@@ -228,14 +228,15 @@ COMMENT: /#[^\n]*/
...
+FLOAT_NUMBER.2: (DEC_NUMBER "." DEC_NUMBER? | "." DEC_NUMBER) (("E" | "e") ["+" | "-"] DEC_NUMBER)?
+              | DEC_NUMBER (("E" | "e") ["+" | "-"] DEC_NUMBER)?
+COMPLEX_NUMBER.2: (DEC_NUMBER | FLOAT_NUMBER) ("+" | "-") (DEC_NUMBER | FLOAT_NUMBER) ("J" | "j")
```
MegaIng, line 237: This is also wrong. Python doesn't have a complex number literal, but instead the addition happens at runtime (or better, they are combined in the compiler). Most notably, this syntax doesn't allow a lone `1j` or `1j+1`, which are both valid in python.

MegaIng (changes requested):

lark/grammars/python.lark
0dminnimda: Hmmm I remember I was testing if I could insert `_` into the bin and I failed (I extrapolated this knowledge to oct and hex), but now I checked again and everything is fine
```python
>>> 0b1_0
2
```
I don't even know how I couldn't write that before :)

I will fix it, thanks!

0dminnimda: And that's interesting that we can do 
```python
>>> 0b_10
2
>>> 0x_10
16
```

lark/grammars/python.lark
0dminnimda: > Python doesn't have a complex number literal, but instead the addition happens at runtime (or better, they are combined in the compiler)

Yeah, that's interesting, in the ast those are two separate constants, while in the byte-code it's already combined
```python
>>> ast.dump(ast.parse("1+5j").body[0].value)
'BinOp(left=Constant(value=1), op=Add(), right=Constant(value=5j))'
>>> dis("1+5j")
  1           0 LOAD_CONST               0 ((1+5j))
              2 RETURN_VALUE
```

> Most notably, this syntax doesn't allow a lone `1j` or `1j+1`, which are both valid in python.

Yep I forgot about it 😅

Well, the reason why I joined it into `COMPLEX_NUMBER` is the convince, but it seems like we can't apply this rule actually.
I never thought of constructions like this 
```python
>>> -1+2j+3+4-5j
(6-3j)
```
We can try to parse this as a complex number but this probably will not be accurate according to CPython's method of parsing

MegaIng: Python has a so called peephole optimizer. After all `1+2` without a `j` is also combined into a single `LOAD_CONST`, despite being two separate entries in the AST. We should mirror that behavior.

0dminnimda: > We should mirror that behavior.

sure

0dminnimda:
This pattern
```lark
FLOAT_NUMBER.2: (DEC_NUMBER "." DEC_NUMBER? | "." DEC_NUMBER) (("E" | "e") ["+" | "-"] DEC_NUMBER)?
              | DEC_NUMBER (("E" | "e") ["+" | "-"] DEC_NUMBER)?
```
cannot parse `5.0_8e-6` it recognizes `5.0` as float, then sees `_` and tries to continue parsing new terminal(?)/rule but there's none that starts with `_`, so it fails


If I change it to
```lark
FLOAT_NUMBER.2: (DEC_NUMBER "." DEC_NUMBER? | "." DEC_NUMBER)
              | (DEC_NUMBER "." DEC_NUMBER? | "." DEC_NUMBER) (("E" | "e") ["+" | "-"] DEC_NUMBER)
              | DEC_NUMBER (("E" | "e") ["+" | "-"] DEC_NUMBER)?
```
then everything works fine.

The reason is regexp, because if I use 
```lark
DIGIT: "0".."9"

DEC_NUMBER: "0" | DIGIT ("_" DIGIT | DIGIT)*
```

instead of previously used
```lark
DEC_NUMBER: /0|\d(?:_?\d)*/i
```
than both patterns for `FLOAT_NUMBER` works as intended.

---

This seems unintentional and therefore worth investigation.
@MegaIng @erezsh should I create an issue for that?

lark/grammars/python.lark
0dminnimda: resolved in 8a67efbd5ede2ec2bc76e626bb93dfecd5f16ab9

lark/grammars/python.lark
0dminnimda: resolved in 8a67efbd5ede2ec2bc76e626bb93dfecd5f16ab9

lark/grammars/python.lark
```diff
@@ -225,17 +225,18 @@ _NEWLINE: ( /\r?\n[\t ]*/ | COMMENT )+
...
+STRING: /[ubf]?r?("(?!"").*?(?<!\\)(\\\\)*?"|'(?!'').*?(?<!\\)(\\\\)*?')/i
 LONG_STRING: /[ubf]?r?(""".*?(?<!\\)(\\\\)*?"""|'''.*?(?<!\\)(\\\\)*?''')/is
 
-DEC_NUMBER: /0|[1-9][\d_]*/i
-HEX_NUMBER.2: /0x[\da-f]*/i
-OCT_NUMBER.2: /0o[0-7]*/i
-BIN_NUMBER.2 : /0b[0-1]*/i
-FLOAT_NUMBER.2: /((\d+\.[\d_]*|\.[\d_]+)([Ee][-+]?\d+)?|\d+([Ee][-+]?\d+))/
-IMAG_NUMBER.2: /\d+[Jj]/ | FLOAT_NUMBER /[Jj]/
+DEC_NUMBER: /0|\d(?:_?\d)*/i
```
0dminnimda, line 231: do we really need this `0|` in `DEC_NUMBER`?

0dminnimda: oh I'm sorry, in the original it was right, I changed it to the wrong one 😅

0dminnimda: resolved in 51b1e030dbe4f3c420ba1b265baae2098ab64611

0dminnimda: fully resolved in 9ca28cfd210d7003257dd6232aacead4b0a01241

lark/grammars/python.lark
```diff
@@ -225,17 +225,18 @@ _NEWLINE: ( /\r?\n[\t ]*/ | COMMENT )+
...
-HEX_NUMBER.2: /0x[\da-f]*/i
-OCT_NUMBER.2: /0o[0-7]*/i
-BIN_NUMBER.2 : /0b[0-1]*/i
-FLOAT_NUMBER.2: /((\d+\.[\d_]*|\.[\d_]+)([Ee][-+]?\d+)?|\d+([Ee][-+]?\d+))/
-IMAG_NUMBER.2: /\d+[Jj]/ | FLOAT_NUMBER /[Jj]/
+DEC_NUMBER: "0"+ | "1".."9" ("_"?  "0".."9"                       )*
+HEX_NUMBER.2: "0x"          ("_"? ("0".."9" | "a".."f" | "A".."F"))+
+OCT_NUMBER.2: "0o"          ("_"?  "0".."7"                       )+
+BIN_NUMBER.2: "0b"          ("_"?  "0".."1"                       )+
+FLOAT_NUMBER.2: (DEC_NUMBER "." DEC_NUMBER? | "." DEC_NUMBER) (("E" | "e") ["+" | "-"] DEC_NUMBER)?
```
0dminnimda, line 235: there's no lower/upper case for numbers(?), so the only thing we need to take from `/.../i` is in the hex, where we have `a-f`

0dminnimda:
ok, so I adapted tests from cpython

here's a successful one
```
0_0_0
4_2
1_0000_0000
0b1001_0100
0xffff_ffff
0o5_7_7
1_00_00.5
1_00_00.5e5
1_00_00e5_1
1e1_0
.1_4
.1_4e1
0b_0
0x_f
0o_5
1_00_00j
1_00_00.5j
1_00_00e5_1j
.1_4j
1_2.5+3_3j
.5_6j

0
0xffffffffffffffff
0Xffffffffffffffff
0o77777777777777777
0O77777777777777777
123456789012345678901234567890
0b100000000000000000000000000000000000000000000000000000000000000000000
0B111111111111111111111111111111111111111111111111111111111111111111111

0j
123456789012345678901234567890j

3.14
314.
0.314
000.314
.314
3e14
3E14
3e-14
3e+14
3.e14
.3e14
3.1e4

3.14j
314.j
0.314j
000.314j
.314j
3e14j
3E14j
3e-14j
3e+14j
3.e14j
.3e14j
3.1e4j
```

0dminnimda:
ok, now the pr is tested and ready for review, and if everything is fine ready for merge!

If anyone could spot untested case, just write it here, I'll add it to the message and test it

lark/grammars/python.lark
```diff
@@ -225,17 +225,22 @@ _NEWLINE: ( /\r?\n[\t ]*/ | COMMENT )+
...
-HEX_NUMBER.2: /0x[\da-f]*/i
-OCT_NUMBER.2: /0o[0-7]*/i
-BIN_NUMBER.2 : /0b[0-1]*/i
-FLOAT_NUMBER.2: /((\d+\.[\d_]*|\.[\d_]+)([Ee][-+]?\d+)?|\d+([Ee][-+]?\d+))/
-IMAG_NUMBER.2: /\d+[Jj]/ | FLOAT_NUMBER /[Jj]/
+DEC_NUMBER: "1".."9" _DEC_END | ("_"?  "0"                            )+ /(?![1-9])/
+_DEC_END:                       ("_"?  "0".."9"                       )*
+HEX_NUMBER.2: "0" ("x" | "X")   ("_"? ("0".."9" | "a".."f" | "A".."F"))+
+OCT_NUMBER.2: "0" ("o" | "O")   ("_"?  "0".."7"                       )+
+BIN_NUMBER.2: "0" ("b" | "B")   ("_"?  "0".."1"                       )+
```
0dminnimda, line 235: there's no lower/upper case (just normal one) for numbers(?), so the only thing we need to take from `/.../i` is in the hex, where we have `a-f` and in the prefixes `0something`

lark/grammars/python.lark
```diff
@@ -225,17 +225,23 @@ _NEWLINE: ( /\r?\n[\t ]*/ | COMMENT )+
...
-OCT_NUMBER.2: /0o[0-7]*/i
-BIN_NUMBER.2 : /0b[0-1]*/i
-FLOAT_NUMBER.2: /((\d+\.[\d_]*|\.[\d_]+)([Ee][-+]?\d+)?|\d+([Ee][-+]?\d+))/
-IMAG_NUMBER.2: /\d+[Jj]/ | FLOAT_NUMBER /[Jj]/
+DEC_NUMBER:   "1".."9"        _DEC_END
+          |   "0"             ("_"?  "0"                            )* /(?![1-9])/
+_DEC_END:                     ("_"?  "0".."9"                       )*
+HEX_NUMBER.2: "0" ("x" | "X") ("_"? ("0".."9" | "a".."f" | "A".."F"))+
+OCT_NUMBER.2: "0" ("o" | "O") ("_"?  "0".."7"                       )+
+BIN_NUMBER.2: "0" ("b" | "B") ("_"?  "0".."1"                       )+
```
0dminnimda, line 236: there's no lower/upper case (just normal one) for numbers(?), so the only thing we need to take from `/.../i` is in the hex, where we have `a-f` and in the prefixes `0something`

0dminnimda:
ok, now the pr is tested and ready for review, and if everything is fine ready for merge!

If anyone could spot an untested case, just write it here, I'll add it to [this](https://github.com/lark-parser/lark/pull/1028#issuecomment-961431935) comment and test it

lark/grammars/python.lark
```diff
@@ -225,17 +225,22 @@ _NEWLINE: ( /\r?\n[\t ]*/ | COMMENT )+
...
-OCT_NUMBER.2: /0o[0-7]*/i
-BIN_NUMBER.2 : /0b[0-1]*/i
-FLOAT_NUMBER.2: /((\d+\.[\d_]*|\.[\d_]+)([Ee][-+]?\d+)?|\d+([Ee][-+]?\d+))/
-IMAG_NUMBER.2: /\d+[Jj]/ | FLOAT_NUMBER /[Jj]/
+_SPECIAL_DEC: "0".."9"        ("_"?  "0".."9"                       )*
+DEC_NUMBER:   "1".."9"        ("_"?  "0".."9"                       )*
+          |   "0"             ("_"?  "0"                            )* /(?![1-9])/
+HEX_NUMBER.2: "0" ("x" | "X") ("_"? ("0".."9" | "a".."f" | "A".."F"))+
+OCT_NUMBER.2: "0" ("o" | "O") ("_"?  "0".."7"                       )+
+BIN_NUMBER.2: "0" ("b" | "B") ("_"?  "0".."1"                       )+
```
0dminnimda, line 236: there's no lower/upper case (just normal one) for numbers(?), so the only thing we need to take from `/.../i` is in the hex, where we have `a-f` and in the prefixes `0something`

MegaIng:
Mind creating a new `test_python_grammar` with unittest inside the `tests` folder? If you (and maybe we as well) put so much effort into it, we should also test it to make sure it doesn't break.

If you don't want to create the unittest, I will do it.

Also, the reason I didn't add those tests till now is because I didn't manage to find CPython's tests for their parser. Where did you find them?

lark/grammars/python.lark
MegaIng: Yeah, that should be correct.

0dminnimda: Ok, then I'll resolve it

0dminnimda:
> Also, the reason I didn't add those tests till now is because I didn't manage to find CPython's tests for their parser. Where did you find them?

Well, currently Python uses [PEG parser generator,](https://github.com/we-like-parsers/pegen) so you may be interested in checking it's tests, there's such things like [tests for match cases](https://github.com/we-like-parsers/pegen/blob/main/tests/python_parser/data/pattern_matching.py) (I know you're trying to megre pr with them)

Cpython also have it's own tests and all of them live in the [Lib/test](https://github.com/python/cpython/tree/main/Lib/test). I myself took cases from [test_int.py](https://github.com/python/cpython/blob/main/Lib/test/test_int.py)
and [test_grammar.py](https://github.com/python/cpython/blob/main/Lib/test/test_grammar.py).
(It's definitely not the most rigorous testing, for example there's [test_float.py](https://github.com/python/cpython/blob/main/Lib/test/test_float.py) or [test_complex.py](https://github.com/python/cpython/blob/main/Lib/test/test_complex.py) and I probably should add test cases from there...)

0dminnimda:
> Mind creating a new `test_python_grammar` with unittest inside the `tests` folder?

oh, ok, `python.lark` don't have a `start` inside, I was just worried that if I just use, for example, `file_input*` as `start` then nothing would work ...

MegaIng:
No `start` also matches the behavior of the actual python grammar: Depending on what exactly you want to parse, use `file_input` (without `*`), `eval_input` or `single_input`

0dminnimda:
> No `start` also matches the behavior of the actual python grammar: Depending on what exactly you want to parse, use `file_input` (without `*`), `eval_input` or `single_input`

there's bug, if I use `file_input` and put only one stmt (like `5`), then the parser will break ...

it even can't parse
```
if x:
    5
```
and even if it could I'd still wouldn't rely on that and test each rule\terminal one-by-one

MegaIng:
Are you putting a newline at the end?

0dminnimda:
> Are you putting a newline at the end?
> 
> >

I shouldn't have to

MegaIng:
It's the same weird obligation that actual python grammar also has. The point is that any higher level parsing function can just always append a newline with no loss.

The problem is that it becomes annoying to express the correct behavior when you can't rely on every line ending in a newline.

erezsh:
@0dminnimda It's not ideal, but making the last newline optional requires support for the EOF feature, which we're keeping back because the existing implementation has issues. It would also make the parser a little heavier, so it's not all positive, even if it worked.

0dminnimda:
@erezsh ok, it looks like we are now ready to merge, I'm waiting for you.

erezsh:
I wonder why you chose to test by specifying a different `start` for each number format, instead of just starting from `number` and making sure the result is the right token type?

The latter approach also has the advantage of testing for collision in token regexp/priority.

0dminnimda:
> I wonder why you chose to test by specifying a different `start` for each number format, instead of just starting from `number` and making sure the result is the right token type?

I didn't think about this approach.
It seems better, gonna implement it

0dminnimda:
@erezsh I updated tests, any other comments?

0dminnimda:
@erezsh could you please review and possibly merge this PR?

tests/test_python_grammar.py
```diff
@@ -0,0 +1,221 @@
+from unittest import TestCase, main
+
+from lark import Lark
+from lark.indenter import PythonIndenter
+from lark.exceptions import UnexpectedCharacters, UnexpectedToken
+
+
+python_parser = Lark.open_from_package(
```
MegaIng, line 8: I think we should delay this till the tests actually get called. When we at some point break Lark's init, we would probably prefer if the dedicated tests fail and not the import itself.

erezsh:
@0dminnimda Looks fine. But please rebase it. It should be at most 3 commits, instead of 23.

(I can rebase it for you, if you prefer)

MegaIng (approved):
Looks good for the most part, only one small thing left.

tests/test_python_grammar.py
erezsh: Good point. It should be in the class `__init__`.

0dminnimda:
> (I can rebase it for you, if you prefer)

Why not squash and merge?

erezsh:
@0dminnimda Squash is done by rebasing, no? Either way, whatever gets it down to a few commits.

tests/test_python_grammar.py
MegaIng: `__init__`? It should probably be `setUpClass`.

0dminnimda: definitely `setUpClass`

0dminnimda:
> Squash is done by rebasing, no?

No and that's the point.

> Either way, whatever gets it down to a few commits.

Well yeah, probably one commit is the best option

0dminnimda:
Ok, @MegaIng do you have any other comments?

MegaIng (approved):

0dminnimda:
Ok, now we're probably ready to merge, @erezsh

erezsh:
Rebased and merged: https://github.com/lark-parser/lark/commit/2ec9636fad34ec43456b637c7c14cfa0de95a1a6

erezsh:
Thanks for the PR!

0dminnimda:
> Rebased and merged: [2ec9636](https://github.com/lark-parser/lark/commit/2ec9636fad34ec43456b637c7c14cfa0de95a1a6)

Yeah, that's why I was talking about squash and merge, because commit that'll end up in the repo gonna be assigned to me, you know, fair credit/blame distribution

erezsh:
Sorry, I do the merges via the shell, because I want to run extra tests on the way, and fix up any remaining issues.

The commits are still signed to your name, and now the merge commit will xref to this PR.

0dminnimda:
> The commits are still signed to your name. So you will still get both credit and blame

Huh, that's true. I'm sorry, I was confused by 2ec9636fad34ec43456b637c7c14cfa0de95a1a6, but apparently there are 518bd3ae976261f8052f18ad6d17961dfd5b107c and 553eb41b538c410e305a3137645166356a8c9b94, so everything is fine, I guess :)

erezsh:
No worries. I'll check if there's a way to connect the merge to a PR from the shell, for next time. After all, I want contributors to feel appreciated and rewarded. (as much as open source allows :)

Issue #1029, opened 2021-11-06
Transformer.__default__ not called in tree-less LALR mode


RoadrunnerWMC:
**Describe the bug**

The `__default__` method on `Transformer`s is never called when using "Tree-less LALR" mode.

**To Reproduce**

Here's an example based on "ab.lark" from the unit tests. `__default__()` should be called for every rule here, but when using tree-less LALR mode (change `BUILD_TREE` to False), it's not called at all.

```py
from lark import Lark, Transformer

BUILD_TREE = True  # Change to False to try tree-less LALR mode

grammar = r"""
    start: expr

    expr: A B
        | A expr B

    A: "a"
    B: "b"

    %import common.WS
    %ignore WS
"""

class AbTransformer(Transformer):
    def __default__(self, data, children, meta):
        print(f'__default__({data!r}, {children!r}, {meta!r})')
        return data

def parse(x):
    if BUILD_TREE:
        parser = Lark(grammar, parser='lalr')
        return AbTransformer().transform(parser.parse(x))
    else:
        parser = Lark(grammar, parser='lalr', transformer=AbTransformer())
        return parser.parse(x)

print(parse('a a a b b b'))  # should print "start"
```

erezsh:
A fix is pending. See https://github.com/lark-parser/lark/pull/1036

PR #1036, opened 2021-11-16
Fix for Transformer.__default__ not called in tree-less LALR mode (Issue #1029)

erezsh:
Example of usage:

```python
from lark import Lark, Transformer

BUILD_TREE = False  # Change to False to try tree-less LALR mode

grammar = r"""
    start: expr

    expr: A B
        | A expr B

    A: "a"
    B: "b"

    %import common.WS
    %ignore WS
"""

class AbTransformer(Transformer):
    def __default__(self, data, children, meta):
        print(f'__default__({data!r}, {children!r}, {meta!r})')
        return data

def parse(x):
    if BUILD_TREE:
        parser = Lark(grammar, parser='lalr')
        return AbTransformer().transform(parser.parse(x))
    else:
        parser = Lark(grammar, parser='lalr', transformer=AbTransformer())
        return parser.parse(x)

print(parse('a a a b b b'))  # should print "start"
```

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -352,6 +352,8 @@ class ParseTreeBuilder:
     def create_callback(self, transformer=None):
         callbacks = {}
 
+        default_callback = getattr(transformer, '__default__', self.tree_class)
+
         for rule, wrapper_chain in self.rule_builders:
 
             user_callback_name = rule.alias or rule.options.template_source or rule.origin.name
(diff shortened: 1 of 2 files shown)


MegaIng (approved):
Sorry, I missed this apparently. Looks good.

Would be nice if we could provide meta, but I don't think there is an easy way. Maybe we can provide a method for the tree class that (re-)calculates a meta attribute based on the content (maybe recursively)?

MegaIng (changes requested):
Wait, forgot to say: A test case would be nice

PR #1039, opened 2021-11-17
Remove "Python 2 & 3 compatible" from README.md

Nightblade:
Since 1.0 isn't Python 2 compatible (according to Reddit post) which makes "& 3" redundant too :)


erezsh:
Thanks!

PR #1049, opened 2021-12-08
Use __name__ in FromPackageLoader in test_grammar.py

pfhayes:
This is the recommended way to call FromPackageLoader from its docs, and makes running the tests a bit more portable


MegaIng (approved):
Good find, this slipped through when originally implementing this. I thought this was changed later on to fix tests failing in in some situations, but that is apparently not the case.

PR #1054, opened 2021-12-13
Changed indentation and blockcodes.

Patitotective:
Replaced indentation code blocks to use \`\`\` and removed indentation on lark (grammar) code blocks.

### DIFF
--- a/docs/json_tutorial.md
+++ b/docs/json_tutorial.md
@@ -18,13 +18,13 @@ Knowledge assumed:
 ## Part 1 - The Grammar
 
 Lark accepts its grammars in a format called [EBNF](https://www.wikiwand.com/en/Extended_Backus%E2%80%93Naur_form). It basically looks like this:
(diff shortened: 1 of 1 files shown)


MegaIng:
Why?

Patitotective:
For me it is clearer having those \`\`\` and in the first code block the syntax highlight wasn't working, now it does.   
In lark grammar code blocks there were indentation before each line which made you remove them when you copy it.

MegaIng:
The problem is that fenced code blocks are generally less supported, since they aren't as old as intended code blocks and instead a somewhat common extension. But it's probably fine.

erezsh:
Sorry, I don't see the value in this PR.

I hope that won't dissuade you from future contributions.

Patitotective:
I understand :upside_down_face:
Thanks for taking the time to reply me. :+1:

PR #1057, opened 2021-12-14
Various fixes for mypy, to reduce error count

erezsh:
Some of these "break" the typing interface, so this PR belongs in v1.1

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -540,7 +540,7 @@ class Lark(Serialize):
             return cls(f, **options)
 
     @classmethod
-    def open_from_package(cls: Type[_T], package: str, grammar_path: str, search_paths: Tuple[str, ...]=("",), **options) -> _T:
+    def open_from_package(cls: Type[_T], package: str, grammar_path: str, search_paths: List[str]=[""], **options) -> _T:
         """Create an instance of Lark with the grammar loaded from within the package `package`.
         This allows grammar loading from zipapps.
 
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -540,7 +540,7 @@ class Lark(Serialize):
             return cls(f, **options)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


lark/lark.py
```diff
@@ -540,7 +540,7 @@ def open(cls: Type[_T], grammar_filename: str, rel_to: Optional[str]=None, **opt
             return cls(f, **options)
 
     @classmethod
-    def open_from_package(cls: Type[_T], package: str, grammar_path: str, search_paths: Tuple[str, ...]=("",), **options) -> _T:
+    def open_from_package(cls: Type[_T], package: str, grammar_path: str, search_paths: List[str]=[""], **options) -> _T:
```
MegaIng, line 543: Why not `Sequence[str]`?

lark/load_grammar.py
```diff
@@ -804,9 +804,9 @@ class FromPackageLoader:
     """
 
     pkg_name: str
-    search_paths: Tuple[str, ...]
+    search_paths: List[str]
 
-    def __init__(self, pkg_name: str, search_paths: Tuple[str, ...]=("", )) -> None:
+    def __init__(self, pkg_name: str, search_paths: List[str]=[""]) -> None:
```
MegaIng, line 809: Same here

MegaIng (changes requested):

lark/lark.py
erezsh: I can change it to Sequence. Anything else?

MegaIng: No, looks good otherwise :-)

MegaIng (approved):

erezsh:
It was surprisingly difficult to install Python 3.6

MegaIng (approved):
Well, let's hope that Python3.6 wont be used for that much longer anyway. But I am not entirely sure if that is a justified hope...

PR #1064, opened 2022-01-04
Have ininstance check match type of argument

plannigan:
Since not all `Mapping`s are `dict`s, mypy didn't understand the type of `example` on the following lines.


erezsh:
Looks good to me.

MegaIng (approved):

PR #1065, opened 2022-01-04
Ignore mypy errors related to hasattr with literal attribute names

plannigan:


lark/exceptions.py
```diff
@@ -107,18 +107,21 @@ def match_examples(self, parse_fn: 'Callable[[str], Tree]',
...
+                            if ut.token == self.token:  # type: ignore[attr-defined]
                                 logger.debug("Exact Match at example [%s][%s]" % (i, j))
                                 return label
 
                             if token_type_match_fallback:
                                 # Fallback to token types match
-                                if (ut.token.type == self.token.type) and not candidate[-1]:
+                                if (ut.token.type == self.token.type) and not candidate[-1]:  # type: ignore[attr-defined]
                                     logger.debug("Token Type Fallback at example [%s][%s]" % (i, j))
                                     candidate = label, True
```
plannigan, line 126: Since this and the previous block are both attempting to achieve the same goal, it is odd that they use different strategies (checking first vs letting it fail). It would probably be "better" pick a single strategy and use it for both, but I wanted to keep this PR to just addressing the mypy errors instead of changing code.

erezsh:
I don't like the idea of adding workarounds to solve mypy's temporary deficiencies.

But I agree we should be consistent in our approaches in the code.

> An alternate option would be to use isinstance() checks for the exceptions 

You mean isinstance on UnexpectedInput? I'm okay with that. I don't think it would be a breaking change.

@MegaIng , opinion?

plannigan:
> You mean isinstance on UnexpectedInput? I'm okay with that. I don't think it would be a breaking change.

Yes. The first block that deals with `accepts` would be `isinstance(x, UnexpectedToken)`. The second block that deals with `token` would be `isinstance(x, (UnexpectedToken, UnexpectedEOF))`.

I agree that it likely should not be considered a breaking change. Reading the docstring for `match_examples()`, while not explicit, it seems fairly clear that the expectation is that it only supports exceptions that are part of lark. The fact that it did support other custom sub-classes of `UnexpectedInput` is an implementation detail which can be changed.

plannigan:
I guess I had gotten distracted by the other PRs that I forgot this was still open and needed the small update. Should be good now.

erezsh:
No worries. I appreciate that you're taking the time to contribute!

PR #1068, opened 2022-01-07
Specify minimum expected type in class methods

plannigan:
Since the type was generic, mypy couldn't know what the `__init__()` looked like. The first concrete class that it does have information on that is a parent is `object`, which doesn't take any arguments.

Resolves
```
ark/lexer.py:180: error: Too many arguments for "object"  [call-arg]
lark/lark.py:540: error: Too many arguments for "object"  [call-arg]
lark/lark.py:558: error: Too many arguments for "object"  [call-arg]
```

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -226,7 +226,7 @@ _VALID_PRIORITY_OPTIONS = ('auto', 'normal', 'invert', None)
 _VALID_AMBIGUITY_OPTIONS = ('auto', 'resolve', 'explicit', 'forest')
 
 
-_T = TypeVar('_T')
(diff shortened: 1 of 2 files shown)


erezsh:
I think the point was to support subclassing. Because if I define `class Foo(Lark)`, then `cls` will be Foo, not Lark.

plannigan:
Ok. That makes sense. Should be able to restore the `TypeVar`, but bind it to `Lark`.

erezsh:
That seems right. Any chance you can flatten them? (just to keep the commit logs a bit cleaner, for blame and such)

MegaIng (approved):

PR #1071, opened 2022-01-08
Remove unneeded python version compatibility functionality

plannigan:
Since the minimum supported version is now 3.6, it is safe to directly depend on [textwrap.indent()](https://docs.python.org/3/library/textwrap.html#textwrap.indent) since it was added in 3.3.

Resolves
```
lark/tools/__init__.py:6: error: All conditional function variants must have identical signatures  [misc]
```

Also remove code to make `filter()` produce an iterator in python 2.

### DIFF
--- a/lark/tools/__init__.py
+++ b/lark/tools/__init__.py
@@ -1,10 +1,6 @@
 import sys
 from argparse import ArgumentParser, FileType
-try:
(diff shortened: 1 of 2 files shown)


erezsh:
Good catch!

MegaIng (approved):

PR #1074, opened 2022-01-09
Python grammar fixes

erezsh:
After these commits, we can parse the entire Python10 stdlib.


erezsh:
@evtn I made a small change to your changes. Feel free to express your opinion.

lark/grammars/python.lark
```diff
@@ -111,7 +114,8 @@ suite: simple_stmt | _NEWLINE _INDENT stmt+ _DEDENT
 ?comparison: expr (comp_op expr)*
 star_expr: "*" expr
 
-?expr: or_expr
+?expr: or_expr | assign_expr
+assign_expr: NAME ":=" expr
```
MegaIng, line 118: Chaining assign_expr isn't allowed.

MegaIng (changes requested):

lark/grammars/python.lark
evtn: I think assign_expr should have `test` on the right side, and I'm not sure if it belongs in `expr` itself
```suggestion
assign_expr: NAME ":=" test
```

MegaIng: Yes, this is true. No, it shouldn't be inside of expression, it needs to be on a higher level than the logical operators and comparison.

lark/grammars/python.lark
```diff
@@ -111,7 +114,8 @@ suite: simple_stmt | _NEWLINE _INDENT stmt+ _DEDENT
 ?comparison: expr (comp_op expr)*
 star_expr: "*" expr
```
evtn, line 115:
```suggestion
star_expr: "*" expr | assign_expr
```
It looks like a right place for `assign_expr`

lark/grammars/python.lark
```diff
@@ -111,7 +114,8 @@ suite: simple_stmt | _NEWLINE _INDENT stmt+ _DEDENT
...
 
-?expr: or_expr
+?expr: or_expr | assign_expr
```
evtn, line 117:
```suggestion
?expr: or_expr
```

And this doesn't look like a right place for assign_expr

lark/grammars/python.lark
```diff
@@ -190,7 +194,7 @@ arguments: argvalue ("," argvalue)*  ("," [ starargs | kwargs])?
 
 starargs: stararg ("," stararg)* ("," argvalue)* ["," kwargs]
 stararg: "*" test
```
evtn, line 196:
```suggestion
stararg: "*" test | assign_expr -> assign_expr
```

I'm not sure about this one, but that's roughly how it is in the [docs](https://docs.python.org/3/reference/expressions.html#calls)

lark/grammars/python.lark
evtn: As [here](https://docs.python.org/3/reference/expressions.html#expression-lists)

MegaIng: No, this would make `a := 5 > 6` mean something different that the original grammar from what I can see.

It has to be on a similar level to `lambdadef` from what I can see.

erezsh:
I'm all for being more precise and to adhere to the official version. But keep in mind we're not trying to *validate* Python, just to parse it correctly when it's valid.

MegaIng:
Yes, but the current version is I think wrong. I didn't try it, but I am pretty sure that the current placement of `assign_expr` gives the wrong result for `a := 0 or 5` or `a := 5 > 7`

evtn:
I've checked it, and:
- assignment expression (if placed in `star_expr`) cannot be parenthesized (current grammar expects `test` or comma-separated list of `test_or_star_expr` in parens), which should work in Python
- if placed as a third option in `test`, it works fine, but unparenthesized top-level `assign_expr` is forbidden by Python, as well as unparenthesized `assign_expr` as argument value. 

If we're not trying to validate Python, I'd go with adding `assign_expr` in `test` like this:
```
?test: or_test ("if" or_test "else" test)?
     | lambdef
     | assign_expr

assign_expr: NAME ":=" test
```

evtn (approved):

lark/grammars/python.lark
```diff
@@ -105,6 +105,10 @@ suite: simple_stmt | _NEWLINE _INDENT stmt+ _DEDENT
 
 ?test: or_test ("if" or_test "else" test)?
      | lambdef
+     | assign_expr
+
+assign_expr: NAME ":=" expr
```
evtn, line 110: Whoops, I've missed this in the review
It should be 
```suggestion
assign_expr: NAME ":=" test
```

evtn (changes requested):

MegaIng (approved):

PR #1083, opened 2022-01-17
Experimental Plugins option

erezsh:
Allows overwriting internal Lark classes from the outside.


erezsh:
@MegaIng Any comments?

I need to merge it to make this work: https://github.com/lark-parser/lark_cython

PR #1087, opened 2022-01-23
setup.py: Small cleanup

erezsh:


erezsh:
It's a tiny change but just making sure, I dunno why we were importing regex there in the first place.

MegaIng (approved):
I think when originally adding regex support, all `import re` got blindly turned into this. No real reason AFAICS

PR #1088, opened 2022-01-23
Annotate base_func

plannigan:
It appears that mypy didn't know the instance variable existed yet because it hadn't had a value assigned to it yet. Another option would be to simply add an annotation to tell mypy what the type should be, but that ran into [this error.](https://github.com/python/mypy/issues/708) 

Resolves
```
lark/visitors.py:462: error: Cannot determine type of "base_func"  [has-type]
```

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -457,10 +457,13 @@ class _VArgsWrapper:
     Otherwise, we use the original function mirroring the behaviour without a __get__.
     We also have the visit_wrapper attribute to be used by Transformers.
     """
+    base_func: Callable
+
     def __init__(self, func: Callable, visit_wrapper: Callable[[Callable, str, list, Any], Any]):
         if isinstance(func, _VArgsWrapper):
             func = func.base_func
-        self.base_func = func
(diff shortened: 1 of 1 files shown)


lark/visitors.py
```diff
@@ -458,11 +458,11 @@ class _VArgsWrapper:
     We also have the visit_wrapper attribute to be used by Transformers.
     """
```
erezsh, line 459: A simpler solution is to just add this, no?

```python
    base_func: Union[Callable, Callable]      # Union needed because mypy is dumb
```

plannigan: I suspect that the union is a different way to work around the issue I linked to. If you'd rather not change the code, I'd prefer an explicit explicitly ignoring the specific error instead of tricking mypy.

erezsh: Yes, I find the new code less readable. Instance variables should be assigned only once, unless there's a really good reason.

I'm okay with ignoring the specific error.

erezsh (approved):

MegaIng (approved):

PR #1091, opened 2022-01-28
Clean up tree templates implementation to reduce mypy errors

plannigan:
Mostly type annotation changes, but also includes a few code tweaks.

Based on how `Tree` was used in the file, I'm fairly certain that the intended use case is `Tree[str]`. However, it is possible that this can be mage even more precise with `ParseType` (and alias for `Tree[Token]`).

Resolves
```
lark/tree_templates.py:35: error: Item "Tree[Any]" of "Union[Any, Tree[Any]]" has no attribute "startswith"  [union-attr]
lark/tree_templates.py:36: error: Item "Tree[Any]" of "Union[Any, Tree[Any]]" has no attribute "lstrip"  [union-attr]
lark/tree_templates.py:123: error: Item "str" of "Union[Tree[Any], str]" has no attribute "iter_subtrees"  [union-attr]
lark/tree_templates.py:155: error: Argument 1 to "translate" has incompatible type "Union[Tree[Any], str]"; expected "Template"  [arg-type]
lark/tree_templates.py:155: error: Argument 2 to "translate" has incompatible type "Union[Tree[Any], str]"; expected "Template"  [arg-type]
```

### DIFF
--- a/lark/tree_templates.py
+++ b/lark/tree_templates.py
@@ -4,12 +4,13 @@ A tree templates is a tree that contains nodes that are template variables.
 
 """
 
-from typing import Union, Optional, Mapping
+from typing import Union, Optional, Mapping, Dict, Tuple, Iterator
 
 from lark import Tree, Transformer
 from lark.exceptions import MissingVariableError
 
-TreeOrCode = Union[Tree, str]
+TreeOrCode = Union[Tree[str], str]
+_TEMPLATE_MARKER = '$'
 
 
 class TemplateConf:
@@ -21,7 +22,7 @@ class TemplateConf:
     def __init__(self, parse=None):
         self._parse = parse
 
-    def test_var(self, var: Union[Tree, str]) -> Optional[str]:
+    def test_var(self, var: Union[Tree[str], str]) -> Optional[str]:
         """Given a tree node, if it is a template variable return its name. Otherwise, return None.
 
         This method may be overridden for customization
@@ -30,17 +31,20 @@ class TemplateConf:
             var: Tree | str - The tree node to test
 
         """
-        if isinstance(var, str) and var.startswith('$'):
-            return var.lstrip('$')
+        if isinstance(var, str):
+            return _get_template_name(var)
 
-        if isinstance(var, Tree) and var.data == 'var' and len(var.children) > 0:
-            first_child = var.children[0]
-            if isinstance(first_child, str) and first_child.startswith('$'):
-                return first_child.lstrip('$')
+        if (
+            isinstance(var, Tree)
+            and var.data == "var"
+            and len(var.children) > 0
+            and isinstance(var.children[0], str)
+        ):
+            return _get_template_name(var.children[0])
 
         return None
 
-    def _get_tree(self, template: TreeOrCode):
+    def _get_tree(self, template: TreeOrCode) -> Tree[str]:
         if isinstance(template, str):
             assert self._parse
             template = self._parse(template)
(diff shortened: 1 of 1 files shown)


lark/tree_templates.py
```diff
@@ -141,16 +144,14 @@ def translate(t1: Template, t2: Template, tree: TreeOrCode):
...
 
 
-
 class TemplateTranslator:
     """Utility class for translating a collection of patterns
     """
 
-    def __init__(self, translations: Mapping[TreeOrCode, TreeOrCode]):
-        assert all( isinstance(k, Template) and isinstance(v, Template) for k, v in translations.items() )
+    def __init__(self, translations: Mapping[Template, Template]):
```
plannigan, line 151: Since the assert statement checks that the dictionary actually contains `Template`, I considered that the source of truth and corrected the annotations to match.

lark/tree_templates.py
```diff
@@ -73,9 +76,9 @@ def _match_tree_template(self, template, tree):
             return res
 
 
-
 class _ReplaceVars(Transformer):
-    def __init__(self, conf, vars):
+    def __init__(self, conf: TemplateConf, vars: Mapping[str, Tree[str]]) -> None:
+        super().__init__()
```
plannigan, line 81: Since `Transformer` has an `__init__()`, it is best to call it in case it does something more important in the future.

lark/tree_templates.py
erezsh: Yeah, you're right.

But why remove the assert?

lark/tree_templates.py
```diff
@@ -29,24 +30,26 @@ def test_var(self, var: Union[Tree, str]) -> Optional[str]:
...
 
         """
-        if isinstance(var, str) and var.startswith('$'):
-            return var.lstrip('$')
+        def _get_template_name(value: str) -> Optional[str]:
+            return value.lstrip(_TEMPLATE_MARKER) if value.startswith(_TEMPLATE_MARKER) else None
 
-        if isinstance(var, Tree) and var.data == 'var' and var.children[0].startswith('$'):
-            return var.children[0].lstrip('$')
+        if isinstance(var, str):
```
erezsh, line 36: Incorrect.

If it doesn't start with `$`, it should return None, since it's not a template variable.

But why did you change this code in the first place?

erezsh (changes requested):

lark/tree_templates.py
plannigan: I guess my thought process was: if the type annotations are correct, type checkers will catch cases where the wrong type is used. However, that assumes everyone is using type checking, which is not accurate. I'll add it back

lark/tree_templates.py
plannigan: `_get_template_name()` will return `None` if the value doesn't start with `$`.

The code had to be updated because the `Tree` portion of this function assumed that the child was a `str`. Since I was touching the code, I opted to remove some of the code duplication.

I just went back to the test cases and noticed that there are [no test cases that cover this file](https://app.codecov.io/gh/lark-parser/lark/blob/master/lark/tree_templates.py). It looks like `py3to2` is the only use of this file in the repository. Based on that, it looks like the annotation for `Template` is not correct (should take `TreeOrCode`). I think I'll take a step back to add test cases, then take a second pass at this.

erezsh: Oh right, my bad.

Anyway, `_get_template_name()` would be better as a global function instead of a closure.

Yes, it's a pretty new feature, not super documented yet either.

lark/tree_templates.py
```diff
@@ -129,7 +136,7 @@ def search(self, tree: TreeOrCode):
             if res:
                 yield subtree, res
 
-    def apply_vars(self, vars: Mapping[str, Tree]):
+    def apply_vars(self, vars: Mapping[str, Tree[str]]) -> Tree[str]:
```
plannigan, line 139: This annotation for `vars` was here originally but conflicts with the annotations I've added based on how the code was written. The key issue is that `TemplateConf._match_tree_template()` returns a dictionary where the value could be a `str` or `Tree[str]`. This then ripples up here because `translate()` uses this result as the `vars` argument. This could be a case that the annotation on `vars` was not completely accurate. Or, since it was one of the few annotations used initially, that it is the intended API. If it is the second, then there would need to a second pass on the implementation to have everything line up.

PR #1096, opened 2022-02-05
Add test cases for tree templates

plannigan:
Also include minor fixes for clear issues

The test cases probably aren't the best. The were mostly written by going backwards for code coverage instead of starting with a strong understanding of the intended API.

While I was working on this, I did notice that most methods are annotated to take in `TreeOrCode`. `_get_tree()` is then used to parse the value if the argument is a `str`. However, `TemplateConf` isn't guaranteed to have a parse function. I guess the idea is that if you already working with just `Tree`s, you don't need one. But the current API makes it very easy to get an `AssertionError` if the classes are not create properly & type annotations aren't going to be able to protect against that.

### DIFF
--- a/lark/tree_templates.py
+++ b/lark/tree_templates.py
@@ -7,8 +7,10 @@ A tree templates is a tree that contains nodes that are template variables.
 from typing import Union, Optional, Mapping
 
 from lark import Tree, Transformer
+from lark.exceptions import MissingVariableError
 
 TreeOrCode = Union[Tree, str]
+
 
 class TemplateConf:
     """Template Configuration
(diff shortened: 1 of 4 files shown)


lark/tree_templates.py
```diff
@@ -84,7 +84,9 @@ def __default__(self, data, children, meta):
 
         var = self._conf.test_var(tree)
         if var:
-            return self._vars[var]
+            value = self._vars.get(var)
```
erezsh, line 87: If the variable doesn't exist, shouldn't we throw an exception instead?

erezsh:
Looks good, thanks!

lark/tree_templates.py
plannigan: Yeah, that makes sense. Since it wasn't documented I had to make a guess. I'll switch it to raise a sub-class of `LarkError`.

erezsh (approved):

MegaIng (approved):

PR #1099, opened 2022-02-08
specify ignored mypy error by using `type: ignore[error]` in lark/tree.py and lark/utils.py

hf-kklein:
instead of plain ignore

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -108,7 +108,7 @@ class Tree(Generic[_Leaf_T]):
         for subtree in queue:
             subtrees[id(subtree)] = subtree
             # Reason for type ignore https://github.com/python/mypy/issues/10999
-            queue += [c for c in reversed(subtree.children)  # type: ignore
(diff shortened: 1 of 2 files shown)


erezsh:
Just curious, what is the benefit of this change?

erezsh:
Well, I guess.

It's a little different with exceptions though, since they can jump out of nested functions. Whereas here the "catcher" is tied directly to the line itself.

erezsh (approved):

MegaIng (commented):
I am not that experienced with mypy. Maybe we should ask someone else as well?

erezsh:
I have. I'll wait to see what he thinks. (though it's probably a non-issue)

hf-kklein:
>  I'll wait to see what he thinks. 

he uses the same style already ;)

https://github.com/lark-parser/lark/blob/c780b3bd4634f8f6b6dc32b69ce70fd4f31bc0d5/lark/visitors.py#L466

erezsh:
Fair enough.

PR #1100, opened 2022-02-08
Fix MyPy Warnings in lark/tools/__init__.py

hf-kklein:
See single commit messages for details.

In the root directory I ran
```
mypy lark --show-error-codes
```

### DIFF
--- a/lark/tools/__init__.py
+++ b/lark/tools/__init__.py
@@ -2,7 +2,7 @@ import sys
 from argparse import ArgumentParser, FileType
 from textwrap import indent
 from logging import DEBUG, INFO, WARN, ERROR
-from typing import Optional, Literal
+from typing import Optional
 import warnings
 
 from lark import Lark, logger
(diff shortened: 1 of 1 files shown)


lark/tools/__init__.py
```diff
@@ -22,17 +23,17 @@
...
+lalr_argparser.add_argument('-o', '--out', type=FileType('w', encoding=encoding), default=sys.stdout, help='the output file (default=stdout)')
+lalr_argparser.add_argument('grammar_file', type=FileType('r', encoding=encoding), help='A valid .lark file')
+
+for flag in flags:
+    if isinstance(flag, tuple):
+        options.append(flag[1])
+        lalr_argparser.add_argument('-' + flag[0], '--' + flag[1], action='store_true')
+    elif isinstance(flag, str):
+        options.append(flag)
+        lalr_argparser.add_argument('--' + flag, action='store_true')
```
erezsh, line 37: Now flags that are neither, will simply be ignored. That's not great. Instead, we should throw an error.

hf-kklein: c34b3c576fd63549b3541b004232a9e561b4831b

PR #1102, opened 2022-02-08
PEP 8: Minor Code Style Improvements

hf-kklein:
I honestly just fixed them because they showed up in my IDE.


erezsh (approved):
Sure, why not

PR #1106, opened 2022-02-08
Include `py.typed` file in package

### LINKED ISSUE #1105: Lark 1.1.0 does not provide a `py.typed` file
**Describe the bug**

Dependabot proposed to upgrade from lark 1.0.0 to lark 1.1.0 in a library that uses lark.
But in the branch that with upgrade v1.1.0 the type checks with mypy break with the following error message for both lark and lark submodules:
> error: Skipping analyzing "lark": module is installed, but missing library stubs or py.typed marker  [import]

**To Reproduce**

See https://github.com/Hochfrequenz/ahbicht/pull/121

I think this can be fixed by adding a `py.typed` file and including it in the package.

hf-kklein:
Might fix #1105
See https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,6 @@ universal = 1
 
 [metadata]
 license_file = LICENSE
+
+[options.package_data]
+* = py.typed


erezsh:
Hmm maybe, not sure. 

You can run `setup.py build` and see if it works.

If it doesn't work, it might be easier to just add it to `MANIFEST.in`

erezsh:
Looked it up. Probably better to add to `package_data` in `setup.py`

like

```json
{ ... 'lark': ['py.typed'] } }

```

PR #1104, opened 2022-02-08
Trivial Renamings for Readability

hf-kklein:

PR #1110, opened 2022-02-08
rename `n` to `child` when iterating over `children`

hf-kklein:


hf-kklein:
obsoletes #1104

erezsh:
Thanks!

PR #1111, opened 2022-02-08
InteractiveParser: Added iter_parse() method, for easier instrumentation

erezsh:


erezsh:
You don't have to review, but thought I'd give you the chance to.

MegaIng (approved):
This makes `exhaust_lexer` expensive for large input text files. Not sure how noticeable this will be in practice.

erezsh:
Well, it's basically obsolete now that 'iter_parse()` exists, so we could just deprecate it.

PR #1115, opened 2022-02-16
Remove redefinitions related to standalone parser

plannigan:
* Use a dictionary to look up valid parser creators
* Split parsing frontend argument validatoin from parsing frontend creation
* Convert MakeParsingFrontend uses into functions

### DIFF
--- a/lark/common.py
+++ b/lark/common.py
@@ -1,15 +1,28 @@
 from copy import deepcopy
+import sys
 from types import ModuleType
 from typing import Callable, Collection, Dict, Optional, TYPE_CHECKING
 
 if TYPE_CHECKING:
     from .lark import PostLex
+    from .lexer import Lexer
+    from typing import Union, Type
+    if sys.version_info >= (3, 8):
+        from typing import Literal
+    else:
+        from typing_extensions import Literal
+    if sys.version_info >= (3, 10):
+        from typing import TypeAlias
+    else:
+        from typing_extensions import TypeAlias
 
 from .utils import Serialize
 from .lexer import TerminalDef, Token
 
 ###{standalone
 
+_ParserArgType: 'TypeAlias' = 'Literal["earley", "lalr", "cyk", "auto"]'
+_LexerArgType: 'TypeAlias' = 'Union[Literal["auto", "basic", "contextual", "dynamic", "dynamic_complete"], Type[Lexer]]'
 _Callback = Callable[[Token], Token]
 
 class LexerConf(Serialize):
@@ -24,6 +37,7 @@ class LexerConf(Serialize):
     g_regex_flags: int
     skip_validation: bool
     use_bytes: bool
+    lexer_type: Optional[_LexerArgType]
 
     def __init__(self, terminals: Collection[TerminalDef], re_module: ModuleType, ignore: Collection[str]=(), postlex: 'Optional[PostLex]'=None, callbacks: Optional[Dict[str, _Callback]]=None, g_regex_flags: int=0, skip_validation: bool=False, use_bytes: bool=False):
         self.terminals = terminals
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -1,3 +1,5 @@
+from typing import Any, Callable, Dict, Tuple
+
 from .exceptions import ConfigurationError, GrammarError, assert_config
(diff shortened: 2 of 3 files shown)


lark/common.py
```diff
@@ -1,15 +1,28 @@
...
+    else:
+        from typing_extensions import TypeAlias
 
 from .utils import Serialize
 from .lexer import TerminalDef, Token
 
 ###{standalone
 
+_ParserArgType: 'TypeAlias' = 'Literal["earley", "lalr", "cyk", "auto"]'
+_LexerArgType: 'TypeAlias' = 'Union[Literal["auto", "basic", "contextual", "dynamic", "dynamic_complete"], Type[Lexer]]'
```
plannigan, line 25: Extract type definition to a place where they can be used in multiple places. Since we aren't making `typing_extensions` a runtime dependencies, the definitions need to be a string to defer the evaluation. Which then requires the definitions to be explicitly marked as `TypeAlias`s, requiring an additional `typing_extensions` usage during type checking.

lark/common.py
```diff
@@ -24,6 +37,7 @@ class LexerConf(Serialize):
     g_regex_flags: int
     skip_validation: bool
     use_bytes: bool
+    lexer_type: Optional[_LexerArgType]
```
plannigan, line 40: Wasn't initially annotated and the initial assignment to `None` made `mypy` think the value could only be `None`.

lark/parser_frontends.py
```diff
@@ -22,19 +24,15 @@ def lex(self, lexer_state, parser_state):
         return CustomLexerWrapper
 
 
-class MakeParsingFrontend:
-    def __init__(self, parser_type, lexer_type):
-        self.parser_type = parser_type
-        self.lexer_type = lexer_type
+def _deserialize_parsing_frontend(data, memo, lexer_conf, callbacks, options):
```
plannigan, line 27: `MakeParsingFrontend.deserialize()` wasn't using the instance variables, so it could be converted into a standalone function.

lark/parser_frontends.py
```diff
@@ -49,11 +47,11 @@ def __init__(self, lexer_conf, parser_conf, options, parser=None):
         if parser:  # From cache
             self.parser = parser
         else:
-            create_parser = {
-                'lalr': create_lalr_parser,
-                'earley': create_earley_parser,
-                'cyk': CYK_FrontEnd,
-            }[parser_conf.parser_type]
+            create_parser = _parser_creators.get(parser_conf.parser_type)
```
plannigan, line 50: This previously would always find a value for the given key, but then would raise an exception when `NotImplemented` was used as a `Callable` (which isn't supported). Now there is a descriptive exception when the given key isn't supported.

lark/parser_frontends.py
```diff
@@ -233,10 +228,19 @@ def _apply_callback(self, tree):
...
-        assert isinstance(lexer_conf, LexerConf)
-        assert isinstance(parser_conf, ParserConf)
-        parser_conf.parser_type = self.parser_type
-        lexer_conf.lexer_type = self.lexer_type
-        return ParsingFrontend(lexer_conf, parser_conf, options)
+_parser_creators['earley'] = create_earley_parser
+_parser_creators['cyk'] = CYK_FrontEnd
+
+
+def _construct_parsing_frontend(
```
plannigan, line 235: `MakeParsingFrontend.__call__()` was using the instance variables. However the only use of a class was in a single function, which immediately used it as a `Callable`. So I feel like it is cleaner to make all of the values arguments to a single function.

erezsh (approved):

erezsh:
It looks alright to me. One question - how do you get to the exception "not supported in standalone mode" ? In realistic use

MegaIng (approved):

MegaIng:
The exception could I think instead also be an `assert False`, since I don't think there should be a way to reach that. The `parser` options should be loaded from the `DATA` value put into the file. To change it you would have to manually override it, just passing it as an option shouldn't work

plannigan:
Yeah, in standalone mode you would need to do what you described or explicitly call `ParsingFrontend()` with a `parser_conf` value which isn't valid. Based on that an assert makes sense. I'll touch that up.

PR #1117, opened 2022-02-21
Added Tree.rich() method

erezsh:


MegaIng (approved):
IDK rich, so I can't offer suggestions/feedback for that.

Maybe an extra mode extra install in the form `lark[rich]` similar to how `regex` is handle? Although I am not sure how useful this would be.

Also, based on this code, rich might be a good basis for an extended version of this showcasing how to use the ``Interpreter`` class to do top down visiting, but I am not sure about that.

Are we testing `Tree.pretty`? If yes, we should try to create similar tests for `Tree.rich`  as well IMO. Maybe just making sure that all children are visited?

erezsh:
I think that because it's just a utility function, and not something that will be used in production code, we don't have to be too judicious about it. Also we don't have tests for pretty() afaik.

I don't know if the extra install is worth it. Maybe if we added more rich methods to other objects, like Token and InteractiveParser.

MegaIng (commented):
Don't forget to update the docs file

erezsh:
Thanks!

MegaIng (approved):

PR #1120, opened 2022-02-26
change MAXREPEAT to int

### LINKED ISSUE #1119: Caching does not work with recursive pattern
**Describe the bug**

I tried to leverage caching to speed up the loading time of my grammar, but it didn't work, so I started to dig into the code and found that it can't load the pickled file

```shell
  ...
  File ".../lark/lark.py", line 326, in __init__
    raise e
  File ".../lark/lark.py", line 321, in __init__
    cached_parser_data = pickle.load(f)
TypeError: __new__() missing 1 required positional argument: 'name'
```

I could trace it down to an issue with `MAXREPEAT` and it happens, when a recursive pattern is used.

**To Reproduce**

```python
from lark import Lark


def main():
    json_grammar = r"""
    start: element*
    
    element: value
    
    ?value: "null"                      -> null
          | QUOTED_INTERPOLATION        -> string
    
    QUOTED_INTERPOLATION: /'(?R)'/
    
    %import common.NEWLINE
    %import common.WS_INLINE
    
    %ignore NEWLINE
    %ignore WS_INLINE
    """

    json_parser = Lark(json_grammar, start="start", parser="lalr", regex=True, cache=True)

    tree = json_parser.parse(
        """
        null
        """
    )
    print(tree.pretty())


if __name__ == "__main__":
    main()
```

gruebel:
Fixes #1119 

Happy to add a test, but not sure, how exactly to test it 😅

### DIFF
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -137,6 +137,7 @@ def get_regexp_width(expr):
             # we manually test for the most important info (whether the empty string is matched)
             c = regex.compile(regexp_final)
             if c.match('') is None:
+                # MAXREPEAT is a none pickable subclass of int, therefore needs to be converted to enable caching
                 return 1, int(sre_constants.MAXREPEAT)
             else:
                 return 0, int(sre_constants.MAXREPEAT)
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -137,9 +137,9 @@ def get_regexp_width(expr):
             # we manually test for the most important info (whether the empty string is matched)
             c = regex.compile(regexp_final)
             if c.match('') is None:
-                return 1, sre_constants.MAXREPEAT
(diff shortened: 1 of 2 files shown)


erezsh:
Seems like a reasonable change.

As for tests, see `tests/test_cache.py`

It should be pretty simple.

gruebel:
@erezsh I added some explanation to the test in regards of the use of `self.assertLogs()`

erezsh (approved):

MegaIng (changes requested):
I would suggest adding a comment in `utils.py` why this is necessary, i.e. mention that `MAXREPEAT` is a non pickable int subclass and therefor needs to be converted.

gruebel:
@MegaIng done 🙂

MegaIng (approved):

PR #1129, opened 2022-03-20
Simplify lexer: Use Match.lastgroup instead of lastindex

erezsh:
No idea why we didn't do that before.

I think there's a reason? But I can't remember what it was.


MegaIng:
I don't think I ever looked at this code closely enough to notice this. Seems reasonable.

erezsh:
I just wonder if there are some edge cases where if users write strange regexps, it returns something different.

MegaIng:
From the documentation of `re` it sounds like those should always match up.

PR #1137, opened 2022-04-18
Move iter_subtrees_topdown into standalone

camgunz:
Hiya! When trying to use `iter_subtrees_topdown` from a generated parser, I found it wasn't there. It looks like that's because it's defined outside the standalone template delimiters, so I just nudged it up some, which fixed the issue for me.


erezsh:
Thanks for the PR

PR #1138, opened 2022-04-23
Fix confusing import in examples

JonasLoos:
Somehow `extend_python.py` imported `PythonIndenter` not from `lark.indenter`, but from `python_parser.py`, which then imported it from `lark.indenter`. As far as I see, this doesn't make any sense.

This PR changes `extend_python.py` to directly import from `lark.indenter` and doesn't change any functionality (as far as I understand).

### DIFF
--- a/examples/advanced/extend_python.py
+++ b/examples/advanced/extend_python.py
@@ -8,7 +8,7 @@ to add new syntax to the example Python grammar.
 """
 
 from lark.lark import Lark
-from python_parser import PythonIndenter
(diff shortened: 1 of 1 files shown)


MegaIng (approved):
The reason for this is historic. `PythonIndenter` wasn't always in `lark.indenter`

JonasLoos:
> The reason for this is historic. `PythonIndenter` wasn't always in `lark.indenter`

Ah ok, makes sense. 
Thanks

PR #1140, opened 2022-04-26
Adjust imports for Python 3.11

The-Compiler:
sre_parse and sre_constants are deprecated in 3.11, see https://github.com/python/cpython/issues/91308

This fixes a `DeprecationWarning` when running `python3 -Wall -c "import lark"` with Python 3.11.0a7.

I couldn't run the tests on Python 3.11 due to https://github.com/PiotrDabkowski/Js2Py/issues/282, but it looks like at least lark works without deprecation warnings with this change.

### DIFF
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -113,8 +113,13 @@ try:
 except ImportError:
     regex = None
 
-import sre_parse
-import sre_constants
+try:  # Python 3.11
+    import re._parser as sre_parse
+    import re._constants as sre_constants
+except ImportError:
+    import sre_parse
+    import sre_constants
+
 categ_pattern = re.compile(r'\\p{[A-Za-z_]+}')
 
 def get_regexp_width(expr):
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -113,10 +113,10 @@ try:
 except ImportError:
     regex = None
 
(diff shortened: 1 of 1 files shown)


lark/utils.py
```diff
@@ -113,8 +113,13 @@ def deserialize(cls, data, namespace, memo):
 except ImportError:
     regex = None
 
-import sre_parse
-import sre_constants
+try:  # Python 3.11
```
henryiii, line 116: I recommend using `sys.version_info` instead of `try` on `import`, it serves as documentation to the user, it enables the workaround to be dropped when that version of Python is dropped, and it can be used by static analysis tools like mypy and pyupgrade.

erezsh: You make a good point, thanks

henryiii: True, though older alpha / beta / rcs are not worth supporting after a new one comes out. Especially alphas once a beta is out. They are essentially just development snapshots.

Build 0.8.0 doesn’t support alpha 6 either.

erezsh:
Thanks for contributing!

Zac-HD:
Any timeline on a release?  As far as I can tell this is the last thing stopping me from testing with `-Werror` on Python 3.11.

PR #1148, opened 2022-05-23
fix: remove Python 2 legacy packaging code

henryiii:
A small update to packaging code; universal=True/1 is only used to replace the `py3` tag with `py2.py3`, which is incorrect since 1.0. Also adding `python_requires` - this should always be set as soon as a python version is dropped, as pip will back-solve to find the final release that has a passing `python_requires` for the current Python version. 

If you are interested, I'd recommend moving to static configuration (everything in setup.cfg or pyproject.toml). I'd recommend reading https://scikit-hep.org/developer/packaging or https://packaging.python.org/en/latest/tutorials/packaging-projects/ for the classic way to do this, or https://scikit-hep.org/developer/pep621 for the new way. (We are working on transitioning packaging.python.org to the PEP 621 method as well in the near future).

My initial goal was to see if mypyc would provide the same speedup for CPython that PyPy provided, but there are some missing static types[^1]. :'( I did compare a few versions of Python 3 to see if the faster CPython project really is working. I've got a fast computer and I wasn't able to generate a matching json file (2.9 MB for me), but here are the relative timings for the slowest parser if you were curious:

| Version       | Time |
|---------------|------|
| CPython 3.6   | 46.34 secs |
| CPython 3.9   | 42.05 secs |
| CPython 3.10  | 37.25 secs | 
| CPython 3.11b1| 29.31 secs |
| PyPy 3.7      | 8.46 secs |

[^1]: there might actually be enough to try it!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,8 +1,3 @@
-[global]
-zip_safe=
-
-[bdist_wheel]
-universal = 1
-
 [metadata]
 license_file = LICENSE
+python_requires = >=3.6


erezsh:
Also, very nice! I had no idea CPython has been getting so much faster with each minor version!

henryiii:
(I have tooling that helps with setup.py -> setup.cfg, and I know of tooling that helps with setup.cfg -> pyproject.toml PEP 621)

henryiii:
FYI, setup.cfg Setuptools is supposed to be able to pull a version attribute without running the file (and since you have no deps, even if it fell back to exec'ing the file, it should still work).

erezsh:
> there are some typing errors; you can't compile with typing errors

They all seem solvable. Let me know if you want me to have a look.

> setup.cfg Setuptools is supposed to be able to pull a version attribute without running the file 

How? And does that apply to pyproject too?

henryiii:
I can try this after #1149 if you'd like to see it. Basically it looks something like this:

```ini
version = attr: lark.__version__
```

setuptools will try to read it from the AST before falling back on importing `lark` to get it (which would also work since there's no deps).

erezsh:
> However, flit does this automatically without configuration

I'm not familiar with flit. It's supposed to be a poetry replacement, right?

So it would automatically know to read the version of `lark.__version__`?

henryiii:
Technically, I meant `flit_core`, which is a simple PEP 517 backend. PEP 517 described the system which builders (like `build` and `pip`) use to talk to backends (PyPA projects `setuptools`, `flit_core`, and `hatchling`, and non-PyPA projects `pdm-pep517`, `poetry_core`, `whey`, `trampolim`, and others). The backend's job is to create the SDist or the wheel. Of all those, only Poetry is missing support for standard configuration language introduced in PEP 621, so transitioning between backends is really trivial unless it's Poetry[^2].

If you produce a pure Python wheel, that doesn't use the backend at all to install, so it's not something that affects users much, mostly a developer choice based on what you find nice and easy to configure.

Reminder: I'd like to keep the current backend for now, until I play with mypyc, which is tied to setuptools (and hatchling has an adaptor), but likely won't be easy to tie to other backend. I plan on working on a PEP to address this, probably in the fall, but we are stuck with the status quo at the moment, which is extensions are tied to a backend.

---

Flit has the built-in feature that it pulls `<package>.__version__` if version is left dynamic. It also can pull description from the module docstring, too. Other backends tend to have ways to get the version from Git (which is what I usually do, avoids any source changes on releases, fewer ways to make mistakes).

---

Almost every existing backend comes with a command-line tool as well. Flit (and Poetry) existed before there was standard tooling for it; I think that's primarily why there is a `flit` command. Though you don't need to use it (I don't). Nowadays, with `pipx` and `build` and `twine`, there's really no need to use a custom command to build SDists, wheels, and upload them. I also really like using the right tool for the job, so I mostly ignore what all-in-one CLI tools are doing.

Most of the CLI tools are trying to be "all-in-one", the question is how much is all on one? Flit does the least and is really easy to just use standard tooling instead. Poetry and PDM do environment management (and PDM follows standards, so I like it better - and you can use any build backend with PDM, not just it's own, which is awesome and I'd like to see that happen with other systems). The far end of the spectrum would be Hatch, which tries to also do multi-environment management (ala Tox/Nox) too[^1].

[^2]: Poetry has promised to support this in Poetry 2, but also promised that 2 is along way off.

[^1]: Hatch is currently missing lockfiles, though - Ofek really has been trying to follow standards, and the standardization attempt for lockfiles fell through.

erezsh:
Thanks for the overview! So far, just using `setup.py` seems to work for Lark.  But I'll be happy to switch to something nicer when it's ready for use. I'll keep flit in mind.

PR #1149, opened 2022-05-24
fix(types): some typing updates

henryiii:
Added mypy configuration, and fixing all but one of the typing issues.

The remaining one I'm not sure about:

```
lark/tree_templates.py:150: error: Argument 1 to "apply_vars" of "Template" has incompatible type "Dict[str, Union[Tree[str], str]]"; expected "Mapping[str, Tree[str]]"  [arg-type]
```

I'm using tox since that's already set up. I personally prefer nox; better defaults, no custom language, explains what it's doing while it's doing it.

- fix(types): address several type issues, add type check job
- ci: update and use tox for mypy


henryiii:
@erezsh Need input on the best path to take to solve the last remaining mypy error.

erezsh:
I think all the return types should just be `Tree`, not `TreeOrCode`.  It was only supposed to be used for function arguments, for convenience.

lark/lark.py
```diff
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 import sys, os, pickle, hashlib
 import tempfile
+import types
```
erezsh, line 4: Is this needed?

.github/workflows/mypy.yml
```diff
@@ -4,16 +4,8 @@ jobs:
...
         with:
-          python-version: 3.8
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install mypy
+          submodules: recursive
       - name: Lint with mypy
-        run: mypy -p lark || true
+        run: pipx run tox -e type
```
erezsh, line 11: Magic!

tox.ini
```diff
@@ -1,16 +1,30 @@
...
-skip_missing_interpreters=true
+envlist = py36, py37, py38, py39, py310, pypy3, type
+skip_missing_interpreters = true
 
 [testenv]
 whitelist_externals = git
 deps =
     -rtest-requirements.txt
+passenv =
+    TERM
```
erezsh, line 10: What does this do?

henryiii: This allows mypy to detect that color is supported by the running terminal (well, assuming it is), and causes it to print colored output instead of uncolored output.

.github/workflows/mypy.yml
henryiii: I _love_ that GHA supports pipx as a first-class package manager!

lark/lark.py
henryiii: Yes, to type `re_module: types.ModuleType` on line 258. Otherwise `re_module` doesn't support being assigned two different Modules.

tox.ini
henryiii: Nox doesn't filter environment variables, but Tox does. TERM shouldn't ever be filtered, since tools might check it to see what you are running in (like mypy does).

erezsh:
Looks good to me! 

@MegaIng Might be nice to have another pair of eyes.

lark/tree_templates.py
```diff
@@ -55,7 +55,7 @@ def _get_tree(self, template: TreeOrCode) -> Tree[str]:
     def __call__(self, template: Tree[str]) -> 'Template':
         return Template(template, conf=self)
 
-    def _match_tree_template(self, template: TreeOrCode, tree: TreeOrCode) -> Optional[Dict[str, TreeOrCode]]:
+    def _match_tree_template(self, template: TreeOrCode, tree: Tree) -> Optional[Dict[str, Tree]]:
```
MegaIng, line 58: The annotation for the argument `tree` isn't quite correct based on the rest of the function, or the test `template == tree` in the `if isinstance(template, str)` case below is not needed.

MegaIng (changes requested):
Looks really nice for the most part, just the annotations in `tree_templates` aren't quite correct I think.

lark/tree_templates.py
MegaIng: Actually, the problem wasn't with these functions returning `TreeOrCode` (although it's a bit of a misnomer), it's the problem of everything else not accepting `TreeOrToken` as an input in the type signature.

erezsh: @henryiii I think the only thing needed it to change back `tree: Tree` into `tree: TreeOrCode`

henryiii: Then it's taking in `Tree | str` and returning `Tree` without converting `str -> Tree`. So MyPy (rightly) complains.

henryiii: `match` and `search` still take `TreeOrCode`. Just not `_match_tree_template`.

MegaIng: The return type should also be `Tree | str`, that was correct before. Wherever else that causes problems when receiving arguments need to be changed to accept `Tree | str`. For example, the template `Tree('data', ["$var"])` applied to `Tree('data', ["value"])` should output `{'$var': "value"}`.

henryiii: Okay, I can go that route instead, was based on the initial comment. Trying again.

erezsh: @MegaIng  Hmm you may have a point!

erezsh: (@henryiii I don't mind fixing it up myself, if you prefer)

henryiii: I think it might need to be a bound TypeVar in at least a few places, since the input matches the output. `translate` does seem to require a tree, at least by the time it starts operating on `.data` and `.children`. Translate is also missing a return annotation.

> I don't mind fixing it up myself, if you prefer

That might be best, I'll revert my change here, then you can edit the PR (or I can type ignore it, then you can follow up, whatever you like).

erezsh:
Fixed `tree_templates.py`, see PR #1150

lark/tree_templates.py
erezsh: No problem, fixed it in #1150.

Sorry for the mixup! It's definitely tricky code, and I wasn't the one who originally added types to it. And in fact, it caught a real edge case!

erezsh:
@henryiii  You can merge my branch here, and then I'll merge this PR. Or you can close this PR and I'll just merge the new one. Whatever works for you.

erezsh:
No problem. Thanks for the contribution!

PR #1150, opened 2022-05-24
Fix for PR #1149

erezsh:


lark/tree_templates.py
```diff
@@ -49,15 +53,21 @@ def _get_tree(self, template: TreeOrCode) -> Tree[str]:
...
         return Template(template, conf=self)
 
-    def _match_tree_template(self, template: TreeOrCode, tree: TreeOrCode) -> Optional[Dict[str, TreeOrCode]]:
+    def _match_tree_template(self, template: TreeOrCode, tree: Branch) -> Optional[MatchResult]:
+        """Returns dict of {var: match} if found a match, else None
+        """
         template_var = self.test_var(template)
         if template_var:
+            if not isinstance(tree, Tree):
+                raise TypeError(f"Template variables can only match Tree instances. Not {tree}")
```
MegaIng, line 70: This should be either `type(tree)` or `{tree!r}`

lark/tree_templates.py
```diff
@@ -9,14 +9,18 @@
...
 
 class TemplateConf:
     """Template Configuration
 
     Allows customization for different uses of Template
+
+    parse() must return a Tree instance.
     """
 
     def __init__(self, parse=None):
```
MegaIng, line 26: This could be annotated with `Callable[[str], Tree]`

MegaIng (changes requested):

lark/tree_templates.py
erezsh: I agree, but there's a lot of "could be" to add there. We can do that in a different PR.

lark/tree_templates.py
erezsh: Fixed.

PR #1151, opened 2022-05-24
style: pre-commit basic config

henryiii:
This adds a basic pre-commit-config, a tox job running it (to help new contributors - usually you should run pre-commit directly to take advantage of it's start-up speed), and a GitHub job. Pre-commit is from Anthony Sottile, the developer behind flake8, pytest, tox, and other tools. You can read more about it at https://pre-commit.com or https://scikit-hep.org/developer/style. You can see it on many of the PyPA repos, like build, pip, pipx, cibuildwheel, etc. If you want, there's also a pre-commit.ci service that will auto-update your pre-commit file and auto-fix PRs, and is ultra-fast.

If you like this, I can add a few more checks, starting with the basics. MyPy can be moved to pre-commit as well. I won't add black or isort unless you ask for them, but I think you'll really like pyupgrade.

By the way, you can exclude files, even per-check, so let me know if there's any that intentionally unusual (like wrong file endings) for testing purposes. I can also ignore SVGs, etc, quite easily.

- chore: add pre-commit basic checks
- style: run pre-commit - basic checks

FYI, I mostly run `pre-commit run -a` like you might run nox or tox, but you can also run `pre-commit install` and have it run in "precommit hook mode", where it checks on commit unless you pass `-n`.


erezsh:
Why did it modify the SVGs?

> MyPy can be moved to pre-commit as well

It sounds like a good fit for pre-commit. But moved from where?

pyupgrade looks nice! (with `--keep-percent-format`, because it doesn't support f-strings, it seems). Though I'm not a fan of pre-commits that change files beyond whitespace. But as long as it's a manual run, it's a nice addition!

Is there a way to split the pre-commit, so that checks run automatically (like with install), but modifications need to be run manually?

Another question - if the files get modified by the CI, who is signed on the commit?

examples/tests/no_newline_at_end.lark
```diff
@@ -1 +1 @@
-start: "a"
\ No newline at end of file
+start: "a"
```
henryiii, line 1: This should be excluded! :)

henryiii:
> Why did it modify the SVGs?

Probably for consistent end lines or extra whitespace.

> But moved from where?

Currently it's sitting in tox.ini.

> because it doesn't support f-strings, it seems

What doesn't support f-strings? Pyupgrade will convert `%` to f-strings if there's no possible output change and the expression in the f-string is very simple. I know of some other tools that are a bit more aggressive that can be used to convert the rest. f-strings are faster than `%` or `.format`[^1].

> Though I'm not a fan of pre-commits that change files beyond whitespace

There's not that much you can do beyond whitespace and a few minor things in Python. We don't have an equivalent to Rubocop for Python. Most tools are just linters only.

> But as long as it's a manual run, it's a nice addition

Pre-commit is just the name of the tool. It's not a very good name.

> Is there a way to split the pre-commit, so that checks run automatically (like with install), but modifications need to be run manually?

You can set stages on all hooks. There is a manual hook stage, but you have to specify it when running pre-commit. I'm not sure if you can specify a "non-precommit" state that runs manually without adding `--hook-stage manual`. I'd generally not bother, and just don't use the precommit mode if you don't want it running. (As I mentioned, I rather rarely install and use it as a git precommit hook).

It's worth trying out, though, if you haven't. It's very well done - it handles partial checkings, for example, by copying to a temp directory and running there, etc. I'd recommend trying out everything on the page I shared, one at a time, eventually. I'll only be adding this (I was having issues with line endings on my last PR), pyupgrade, and maybe some linting checks - I'm really after improving the typing coverage a bit so I can measure the impact of mypyc. MyPy and Black both use mypypc to get nice speedups. Style formatting (Black, isort, etc) are a personal choice, and not a usually a good contribution unless asked for.

> It's better that the CI won't make any changes

You can still activate pre-commit.ci, but leave changes off. We do that in pypa/build. You still get the fast checks & weekly updates.

---

I'll add a couple of exclusions, and repush, and we can go from there. Probably in an hour or two.

[^1]: Python 3.11 allows very simple `%` formatting to be as fast as f-strings.

erezsh:
Thanks you for your contribution!

Sorry that it took so long to review. My mind's been on other projects.

PR #1152, opened 2022-05-25
fix(types): add some types based on MonkeyType

henryiii:
Typing is a bit too sparse to run mypyc, so I've tried to fill in a couple of modules (lark.py and utils.py) with the help of monkeytype. This is based on what was seen running the tests more than what things should be, so input and cleanup would be helpful!

### DIFF
--- a/lark/parsers/lalr_parser.py
+++ b/lark/parsers/lalr_parser.py
@@ -30,7 +30,7 @@ class LALR_Parser(Serialize):
         inst.parser = _Parser(inst._parse_table, callbacks, debug)
         return inst
 
-    def serialize(self, memo = None) -> Dict[str, Any]:
+    def serialize(self, memo: Any = None) -> Dict[str, Any]:
         return self._parse_table.serialize(memo)
 
     def parse_interactive(self, lexer, start):
--- a/lark/parsers/lalr_parser.py
+++ b/lark/parsers/lalr_parser.py
@@ -3,8 +3,9 @@
 # Author: Erez Shinan (2017)
 # Email : erezshin@gmail.com
 from copy import deepcopy, copy
+from typing import Dict, Any
 from ..lexer import Token
-from ..utils import Serialize
+from ..utils import Serialize, SerializeMemoizer
 
 from .lalr_analysis import LALR_Analyzer, Shift, Reduce, IntParseTable
 from .lalr_interactive_parser import InteractiveParser
@@ -29,7 +30,7 @@ class LALR_Parser(Serialize):
         inst.parser = _Parser(inst._parse_table, callbacks, debug)
         return inst
 
-    def serialize(self, memo):
+    def serialize(self, memo = None) -> Dict[str, Any]:
         return self._parse_table.serialize(memo)
 
     def parse_interactive(self, lexer, start):
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -2,7 +2,7 @@ import unicodedata
 import os
 from functools import reduce
 from collections import deque
-from typing import Callable, Iterator, List, Optional, Tuple, Type, TypeVar, Union, Dict, Any, TYPE_CHECKING
+from typing import Callable, Iterator, List, Optional, Tuple, Type, TypeVar, Union, Dict, Any, Sequence, TYPE_CHECKING
 
 ###{standalone
 # Can remove for python 3.7+ and from __future__ import annotations
@@ -172,7 +172,7 @@ def get_regexp_width(expr: str) -> Union[Tuple[int, int], List[int]]:
 _ID_START =    'Lu', 'Ll', 'Lt', 'Lm', 'Lo', 'Mn', 'Mc', 'Pc'
 _ID_CONTINUE = _ID_START + ('Nd', 'Nl',)
 
-def _test_unicode_category(s: str, categories: Union[Tuple[str, str, str, str, str, str, str, str], Tuple[str, str, str, str, str, str, str, str, str, str]]) -> bool:
+def _test_unicode_category(s: str, categories: Sequence[str]) -> bool:
     if len(s) != 1:
         return all(_test_unicode_category(char, categories) for char in s)
     return s == '_' or unicodedata.category(s) in categories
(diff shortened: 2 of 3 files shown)


lark/parsers/lalr_parser.py
```diff
@@ -29,7 +30,7 @@ def deserialize(cls, data, memo, callbacks, debug=False):
         inst.parser = _Parser(inst._parse_table, callbacks, debug)
         return inst
 
-    def serialize(self, memo):
+    def serialize(self, memo = None) -> Dict[str, Any]:
```
henryiii, line 33: Adding `= None` here to make mypy happy, otherwise the subclass doesn't match the parent.

lark/utils.py
```diff
@@ -93,18 +110,18 @@ class SerializeMemoizer(Serialize):
...
+    def __init__(self, types_to_memoize: List[Union[Type[TerminalDef], Type[Rule]]]) -> None:
         self.types_to_memoize = tuple(types_to_memoize)
         self.memoized = Enumerator()
 
-    def in_types(self, value):
+    def in_types(self, value: Serialize) -> bool:
         return isinstance(value, self.types_to_memoize)
 
-    def serialize(self):
+    def serialize(self) -> Dict[int, Any]:  # type: ignore[override]
```
henryiii, line 120: This time I just ignored the mismatch.

lark/utils.py
```diff
@@ -134,7 +151,8 @@ def get_regexp_width(expr):
             raise ImportError('`regex` module must be installed in order to use Unicode categories.', expr)
         regexp_final = expr
     try:
-        return [int(x) for x in sre_parse.parse(regexp_final).getwidth()]
+        # TODO: bug? Returns an int?
+        return [int(x) for x in sre_parse.parse(regexp_final).getwidth()]   # type: ignore[attr-defined]
```
henryiii, line 155: `sre_parse.parse(regexp_final).getwidth()` returns an int, which you can't iterate over. Not sure what was intended here.

lark/utils.py
```diff
@@ -154,47 +172,51 @@ def get_regexp_width(expr):
 _ID_START =    'Lu', 'Ll', 'Lt', 'Lm', 'Lo', 'Mn', 'Mc', 'Pc'
 _ID_CONTINUE = _ID_START + ('Nd', 'Nl',)
 
-def _test_unicode_category(s, categories):
+def _test_unicode_category(s: str, categories: Union[Tuple[str, str, str, str, str, str, str, str], Tuple[str, str, str, str, str, str, str, str, str, str]]) -> bool:
```
henryiii, line 175: Ahh, didn't see this. 😆 
```suggestion
def _test_unicode_category(s: str, categories: Tuple[str, ...]) -> bool:
```

henryiii: Or

```suggestion
def _test_unicode_category(s: str, categories: Sequence[str]) -> bool:
```

lark/utils.py
henryiii: Ahah, bug in typeshed! https://github.com/python/typeshed/blob/cb5b31cf15da14c63cb6467b5a29315471173ced/stdlib/sre_parse.pyi#L68

lark/utils.py
```diff
@@ -154,47 +172,51 @@ def get_regexp_width(expr):
...
     """Given a list (l) will removing duplicates from the list,
        preserving the original order of the list. Assumes that
        the list entries are hashable."""
+       # Types seend are Tree, SlottedTree, Terminal, NonTerminal
     dedup = set()
-    return [x for x in l if not (x in dedup or dedup.add(x))]
+    # This returns None, but that's expected
+    return [x for x in l if not (x in dedup or dedup.add(x))]  # type: ignore[func-returns-value]
+    # 2x faster (ordered in PyPy and CPython 3.6+, gaurenteed to be ordered in Python 3.7+)
+    # return list(dict.fromkeys(l))
```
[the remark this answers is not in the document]
erezsh: I didn't write this function, but looks like for someone else it was the opposite. Don't know why.

lark/parsers/lalr_parser.py
erezsh: It's probably more correct to say the return type is `Dict[Any, Any]`? But I'm rusty on that code, didn't touch it in a long time.

lark/utils.py
erezsh: Well, it's quite an obscure feature. Most re libraries in other languages don't even have it.

lark/parsers/lalr_parser.py
MegaIng: No, `Dict[str, Any]` is correct, this is being fed into json.

henryiii:
```suggestion
    def serialize(self, memo: Any = None) -> Dict[str, Any]:
```

lark/utils.py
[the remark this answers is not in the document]
erezsh: Did you test it on heavily duplicated lists?

It's possible that fromkeys() keeps overwriting existing values, making it slower in that case.

(Just guessing, no idea if it's true)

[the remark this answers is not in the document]
henryiii: Yes, 1,000,000 items with 100 unique values.

lark/utils.py
henryiii: Fixed in typeshed now, but missed today's mypy 0.960 release.

henryiii: Oops, can't resolve until I fix the comment (tomorrow).

erezsh:
@henryiii I'm sorry it took me so long to review this PR! I will try to be more prompt in the future.

I have two issues with this PR:

- `pickle.dump / load` do not limit the input to `BytesIO`, so I don't think that we should either.
- `utils.py` should not be aware of any Lark-specific types, it's a violation of concerns.
  - If we need more specificity for functions like `bfs()` or `classify()`, it should be done through generics. e.g. `bfs( Sequence[T] ) -> Iterator[T] `
  - If serialize/deserialize can't be expressed effectively using generics (though I think they should?), it's better to move them to a separate module like `serialize.py`

I created a new PR where I revert only these changes. Let me know if you agree with it - https://github.com/lark-parser/lark/pull/1191

henryiii:
That's fine, I'm assuming this will be an iterative process, and that's fine. :)

PR #1157, opened 2022-06-01
Moved the metadata into `PEP 621`-compliant `pyproject.toml`.

KOLANICH:


erezsh:
Thanks, but I think we'll stick to setup.py until pyproject.toml can extract the version from `lark/__init__.py`.

erezsh:
@KOLANICH I think it's great that you can do `lark.__version__` and get the version, very helpful for debugging. I'm actually annoyed that a lot of packages don't do that these days.

KOLANICH:
`setuptools_scm` has a feature to generate a python source code file for a version extracted from git tags. I can utilize it instead of fetching version from `__init__.py`.

erezsh:
That sounds like a roundabout solution, but I'm willing to consider it.

henryiii:
See https://scikit-hep.org/developer/pep621#versioning for an example with hatchling (which works with setuptools_scm). Most of https://scikit-hep.org/developer/packaging#git-tags-official-pypa-method is valid if you want to stick with setuptools, just ignore the setup.cfg mentions. :)

pyproject.toml
```diff
@@ -1,7 +1,76 @@
 [build-system]
-requires = ["setuptools>=42"]
+requires = ["setuptools>=61.2.0", "wheel"]
```
henryiii, line 2:
```suggestion
requires = ["setuptools>=61.2.0"]
```

You don’t need wheel - it’s added by setuptools via PEP 517 and it’s likely to be going away at some point.

pyproject.toml
```diff
@@ -1,7 +1,76 @@
...
+
+[project.entry-points.pyinstaller40]
+hook-dirs = "lark.__pyinstaller:get_hook_dirs"
+
+[project.optional-dependencies]
+regex = ["regex"]
+nearley = ["js2py"]
+atomic_cache = ["atomicwrites"]
+
+[tool.setuptools]
```
henryiii, line 59: Hatching handles this better, IMO, and works even if some limits setuptools<60 in a non isolated environment (cough, NumPy, cough)

KOLANICH:
>if some limits setuptools<60 in a non isolated environment (cough, NumPy, cough)

I heard something about it (only a bit, I have not yet lurked about the issue), but I think it is an issue of `numpy`. If some package uses `<` constraints, it is that package that is broken (it doesn't mean the ones that require `<=` shouldn't lover the version on RHS of the expression).

>You don’t need wheel - it’s added by setuptools via PEP 517 and it’s likely to be going away at some point.

Thanks for noting this, I'll fix it.

erezsh:
Cool, that looks elegant! Do you think the transition will be entirely seamless, or are there any details to be aware of?

KOLANICH:
>Do you think the transition will be entirely seamless, or are there any details to be aware of?

It depends on your assumptions on Python and `setuptools` versions users have. Since `setuptools` has dropped old Python versions.

erezsh:
This is the first time I hear of `python3 -m build`.

Is there a guide or tutorial for transitioning from setup.py, for those of us who are out of scope? ;)

KOLANICH:
Here they are: https://pypa-build.readthedocs.io/en/latest/

`python3 -m build` is a backend-neutral way to trigger build (now `setuptools` is just one of the backends). [`PEP 517`](https://peps.python.org/pep-0517/) and [`PEP 518`](https://peps.python.org/pep-0518/) are used to specify the backend, be it `setuptools`, or something else (like `flit` or `hatchling`). Those backends just have no `setup.py`. [`PEP 621`](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata) is just a backend-neutral format for metadata instead of the total mess that happenned after `PEP 517` got implemented in `pip`, when each backend used own format. I cannot say the mess is fully cleaned, but PEP 621 is surely an improvement.

henryiii:
There’s also https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

https://scikit-hep.org/developer is a good place for modern packaging suggestions. Also I’m fond of `hatch new --init`, which converts setup.py/setup.cfg directly to hatchling.

Many PyPA packages have moved, like packaging and wheel. So it’s pretty safe.

henryiii:
And personally I almost always use `pipx run build`. :)

erezsh:
Hello! Sorry it took me this long to look into this. Thank you for the info and the links. I found scikit hep especially interesting.

I'm generally in favor of the change. 

Just one question - if I understand correctly, users are now expected to run "pytest" instead of "setup.py test". Does it mean we should also include pytest in the dev-requirements?

Or is it up to the user to install it, or use pipx run pytest?

henryiii:
FYI, scikit-hep.org/developer has been mostly migrated to https://learn.scientific-python.org/development/

Yes, `setup.py test` has been deprecated for years. You do want pytest in dev-requirements (I usually have a `[test]` extra, so a user can install `-e.[test]`).

Usually you don't want use pipx for pytest, since pytest imports your package, its requirements, and plugins; it's more like a library than an app. In some simple cases it does work, but it's generally designed to run alongside your package.

pyproject.toml
henryiii: FYI, from what I gather, the next version of setuptools is making `tool.setuptools` non-provisional.

KOLANICH:
My approach to testing is writing test suites using vanilla `unittest` and sometimes with standalone libs not related to `pytest` (to allow them be run without `pytest` which is large and slow and a dependency, and `pytest` CLI tool is smart enough to understand such tests), but to run tests using `pytest` on CI in order to generate reports (on GitLab CI these reports can be integrated into GUI by pointing the report file in the config). Locally I usually use `pytest --lf` or just `./tests/tests.py` depending on what is faster (`pytest` initialization is quite slow).

erezsh:
So, everyone okay with merging this as-is?  We can add a [test] extra in a separate PR.

@MegaIng Feel free to chime in too, if you have something to add.

MegaIng:
I don't have a lot of concrete experience with this stuff and therefore don't have anything to add.

I am okay with merging.

erezsh:
Thanks everyone!

PR #1171, opened 2022-07-27
Create lark/grammars/__init__.py

chanicpanic:
`lark.grammars` is declared as a package in [setup.py](https://github.com/lark-parser/lark/blob/9e38c4dbb8b0414300d24b8de8bb2852269f08be/setup.py#L9), but `lark/grammars` is missing an `__init__.py`.

Without it, attempting something like `Lark.open_from_package("lark.grammars", "lark.lark")` fails.


erezsh:
Thanks!

Just fyi, what I usually do is `Lark.open_from_package("lark", "grammars/lark.lark")`

But there's no harm in having it both ways.

PR #1179, opened 2022-08-12
Add user to cache filename; better handle cache load/save failures

### LINKED ISSUE #1178: `lark.Lark` `cache=True` default cache filename can be problematic on multi-user systems
**Describe the bug**

The default `lark.Lark` caching mechanism can fail when utilized on a multi-user system with more than one user due to the default permissions of the cached file. (See the traceback/reproduction steps below.)

**To Reproduce**

* Using a lark grammar with [caching enabled](https://github.com/pcdshub/whatrecord/blob/cdeb432b81c7f1f9e0ee9841a50190126a3bba23/whatrecord/db.py#L656)
* User 1 runs the program and lark generates the cache file:
     `-rw------- 1 user1 group1 28250 Aug  3 09:41 /tmp/.lark_cache_42d962555f1b4b923579d2cfd7bfbbfc_3_9.tmp`
* User 2 runs the program and lark tries to open from the cache. Read access fails due to permissions on the file:

```
...
  File "env_path/python3.9/site-packages/whatrecord/db.py", line 646, in from_string
    grammar = lark.Lark.open_from_package(
  File "env_path/python3.9/site-packages/lark/lark.py", line 566, in open_from_package
    return cls(text, **options)
  File "env_path/python3.9/site-packages/lark/lark.py", line 315, in __init__
    with FS.open(cache_fn, 'rb') as f:
  File "env_path/python3.9/site-packages/lark/utils.py", line 232, in open
    return open(name, mode, **kwargs)
PermissionError: [Errno 13] Permission denied: '/tmp/.lark_cache_bf910f0d49cfefbc0576ce70de8a7dbd_3_9.tmp'
```

I wonder if the default cache filename format should perhaps also include a username? Read-only permissions for group/other could be okay, too.

klauer:
## Context

Closes #1178 

## Description

This is my attempt at fixing the issue I reported (#1178) regarding cache files generated by different users on the same system.

Consider this a proposal - I'd be happy to adjust it to your preference.

Three main changes were made:
* Added `$USER` (via `getpass.getuser()`) to the cache filename (77d8277)
* Moved the `try/except` block to include cache load / `open(cache_fn)` (1c36098)
   * This means that if, for any reason, the cache file fails to load, the normal loading procedure can continue
* Wrapped the cache save section in a `try/except` block and a similar `logger.exception` message (835e6923239ee2ecee9a6dc6637a2cd73065e38a)
   * I believe cache save failures should not be fatal

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+import getpass
 import sys, os, pickle, hashlib
 import tempfile
 import types
@@ -308,7 +309,7 @@ class Lark(Serialize):
                     if self.options.cache is not True:
                         raise ConfigurationError("cache argument must be bool or str")
                         
-                    cache_fn = tempfile.gettempdir() + '/.lark_cache_%s_%s_%s.tmp' % (cache_md5, *sys.version_info[:2])
+                    cache_fn = tempfile.gettempdir() + "/.lark_cache_%s_%s_%s_%s.tmp" % (getpass.getuser(), cache_md5, *sys.version_info[:2])
 
                 if FS.exists(cache_fn):
                     logger.debug('Loading grammar from cache: %s', cache_fn)
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -328,7 +328,7 @@ class Lark(Serialize):
                     # The cache file doesn't exist; parse and compose the grammar as normal
                     pass
                 except Exception: # We should probably narrow done which errors we catch here.
-                    logger.exception("Failed to load Lark from cache: %r. We will try to carry on." % cache_fn)
+                    logger.exception("Failed to load Lark from cache: %r. We will try to carry on.", cache_fn)
                     
                     # In theory, the Lark instance might have been messed up by the call to `_load`.
                     # In practice the only relevant thing that might have been overriden should be `options`
@@ -424,11 +424,14 @@ class Lark(Serialize):
 
         if cache_fn:
             logger.debug('Saving grammar to cache: %s', cache_fn)
-            with FS.open(cache_fn, 'wb') as f:
-                assert cache_md5 is not None
-                f.write(cache_md5.encode('utf8') + b'\n')
-                pickle.dump(used_files, f)
-                self.save(f, _LOAD_ALLOWED_OPTIONS)
+            try:
+                with FS.open(cache_fn, 'wb') as f:
+                    assert cache_md5 is not None
+                    f.write(cache_md5.encode('utf8') + b'\n')
+                    pickle.dump(used_files, f)
+                    self.save(f, _LOAD_ALLOWED_OPTIONS)
+            except Exception:
+                logger.exception("Failed to save Lark to cache: %r.", cache_fn)
 
     if __doc__:
         __doc__ += "\n\n" + LarkOptions.OPTIONS_DOC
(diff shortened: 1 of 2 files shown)


erezsh:
> Moved the try/except block to include cache load 

The one thing that bugs me is that now it will always say "Loading grammar from cache ...", even when the cache file doesn't exist. At least, we should change it to "checking for cache for grammar", or something of the sort.

@MegaIng Mind also taking a look?

lark/lark.py
```diff
@@ -308,28 +309,30 @@ def __init__(self, grammar: 'Union[Grammar, str, IO[str]]', **options) -> None:
                     if self.options.cache is not True:
                         raise ConfigurationError("cache argument must be bool or str")
                         
-                    cache_fn = tempfile.gettempdir() + '/.lark_cache_%s_%s_%s.tmp' % (cache_md5, *sys.version_info[:2])
+                    cache_fn = tempfile.gettempdir() + "/.lark_cache_%s_%s_%s_%s.tmp" % (getpass.getuser(), cache_md5, *sys.version_info[:2])
```
MegaIng, line 312: I am not sure what the gurantees are about the value returned from `getpass.getuser()`. It could potentially return a string that results in an invalid file name, but we could also wait till someone actually reports a problem with this. In a similar vein, `getuser` might fail on some weird OSes and raise an exception.

lark/lark.py
```diff
@@ -308,28 +309,30 @@ def __init__(self, grammar: 'Union[Grammar, str, IO[str]]', **options) -> None:
...
 
-                if FS.exists(cache_fn):
```
MegaIng, line 313: Removing the call to `FS.exists` might break the test suite. The `MockFS.open` function does not result in an exception being generated if the file is missing.

MegaIng (commented):
@erezsh The behavior of always printing `loading from cache` might just be only when running tests and not when actually running the code with a real FileSystem. I would think the call to `FS.open` should already cause the Exception before getting to the logger.debug statement.

lark/lark.py
klauer: * Per https://docs.python.org/3/library/getpass.html#getpass.getuser it can raise an exception if attempts at finding a username fail (the exact exception may change in the future: https://github.com/python/cpython/pull/29739) I could wrap `getuser` in a try/except block if that would be preferable?
* As to the invalid filename, I suppose it could be a possibility. On all operating systems I'm aware of, users generally have directories that include the full username.
* Alternatively, perhaps these cached files don't belong in `/tmp` (as in `gettempdir`) but rather the OS-specific user-cache directory (xdg or otherwise, using [appdirs](https://github.com/ActiveState/appdirs)'s `user_cache_dir` for example )

lark/lark.py
klauer: A call to `file.exists()` and then subsequently `open()` is redundant and susceptible to a race condition.

It looks like the test suite would need reworking to appropriately handle this change. 
Would you like prefer that a test suite fix be added to this PR, or would you prefer the `exists()` check to be re-added?

MegaIng: No, what you implemented is pretty much the same I also came to conclude to be the best solution.

lark/lark.py
MegaIng: Whether anything should be done here depends on whether or not this might fail on some platforms and how often this happens. I would say keep it like this, and we take another look if someone reports this breaking caching for them. But we should keep this in the back of our minds.

MegaIng (approved):

klauer:
Anything else I can do to help move this along?

erezsh:
@klauer  Sorry, I've been a little busy. I'll try to look into it this weekend.

erezsh:
Thanks for the PR!

PR #1181, opened 2022-08-17
fix: Token now pattern matches correctly

### LINKED ISSUE #1180: Token is inconsistent in py3.10 pattern matching
**Describe the bug**

`Token` is not working correctly with structural pattern matching.

While you can create tokens as:

``` python
t = Token(type_="TEST", value="test")
```

you have to pattern match on:

``` python
match t:
    case Token(type="TEST"):
        pass
```

As pattern matching tries to access the attributes by names provided to the pattern.

I find this solution fairly confusing and ugly, as well as hard to work with type checkers and triggers a nasty bug in pylance (https://github.com/microsoft/pylance-release/issues/3197).

**To Reproduce**

``` python
t = Token(type_="TEST", value="test")

match t:
    case Token(type_="TEST"):
        print("with underscore") # expected match
    case Token(type="TEST"):
        print("without underscore") # actual match
```

**Potential solutions**
The simplest solution would be to rename `type`/`type_` to be consistent between the `__init__` argument and the attribute, but this might be a fairly big change to the API. There might be a possibility to add custom pattern matching logic so that matching works on `Token(type_=...)`, but I'm not very familiar with the pattern matching API yet.

marcinplatek:
I've added fixes as per https://github.com/lark-parser/lark/issues/1180#issuecomment-1214921614. I don't think that the backward compatibility is possible in this case, as the original argument is positional, e.g.

``` python
Token(type_="TEST", value="test") # this will break no matter what, as positional arguments are not set properly

Token("TEST", "test") # this doesn't require any backwards compatibility changes
```

Please correct me if I'm missing something.

I've also added a bunch of tests (which only are run if the version is 3.10 or higher).
I wasn't sure how to use assert statements in structural pattern matching, but it does the job.

closes #1180

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -147,6 +147,8 @@ class Token(str):
     """
     __slots__ = ('type', 'start_pos', 'value', 'line', 'column', 'end_line', 'end_column', 'end_pos')
 
+    __match_args__ = ('type', 'value', 'start_pos', 'line', 'column', 'end_line', 'end_column', 'end_pos')
+
     type: str
     start_pos: int
     value: Any
@@ -156,9 +158,9 @@ class Token(str):
     end_column: int
     end_pos: int
 
-    def __new__(cls, type_, value, start_pos=None, line=None, column=None, end_line=None, end_column=None, end_pos=None):
+    def __new__(cls, type, value, start_pos=None, line=None, column=None, end_line=None, end_column=None, end_pos=None):
         inst = super(Token, cls).__new__(cls, value)
-        inst.type = type_
+        inst.type = type
         inst.start_pos = start_pos
         inst.value = value
         inst.line = line
@@ -178,6 +180,10 @@ class Token(str):
     @classmethod
     def new_borrow_pos(cls: Type[_T], type_: str, value: Any, borrow_t: 'Token') -> _T:
         return cls(type_, value, borrow_t.start_pos, borrow_t.line, borrow_t.column, borrow_t.end_line, borrow_t.end_column, borrow_t.end_pos)
+    
+    @property
+    def type_(self) -> str:
+        return self.type
 
     def __reduce__(self):
         return (self.__class__, (self.type, self.value, self.start_pos, self.line, self.column))
--- a/tests/__main__.py
+++ b/tests/__main__.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, print_function
 
 import unittest
 import logging
+import sys
 from lark import logger
 
 from .test_trees import TestTrees
@@ -25,6 +26,9 @@ except ImportError:
 from .test_logger import Testlogger
 
 from .test_parser import *  # We define __all__ to list which TestSuites to run
+
+if sys.version_info >= (3, 10):
+    from .test_pattern_matching import TestPatternMatching
 
 logger.setLevel(logging.INFO)
 
--- a/tests/test_pattern_matching.py
+++ b/tests/test_pattern_matching.py
@@ -1,0 +1,52 @@
+from unittest import TestCase, main
+
+from lark import Token
+
+
+class TestPatternMatching(TestCase):
+    token = Token('A', 'a')
+
+    def setUp(self):
+        pass
+
+    def test_matches_with_string(self):
+        match self.token:
+            case 'a':
+                pass
+            case _:
+                assert False
+
+    def test_matches_with_str_positional_arg(self):
+        match self.token:
+            case str('a'):
+                pass
+            case _:
+                assert False
+    
+    def test_matches_with_token_positional_arg(self):
+        match self.token:
+            case Token('a'):
+                assert False
+            case Token('A'):
+                pass
+            case _:
+                assert False
+    
+    def test_matches_with_token_kwarg_type(self):
+        match self.token:
+            case Token(type='A'):
+                pass
+            case _:
+                assert False
+    
+    def test_matches_with_token_kwarg_type_(self):
+        match self.token:
+            case Token(type_='A'):
+                pass
+            case _:
+                assert False
+            
+
+
+if __name__ == '__main__':
+    main()


MegaIng (commented):
It is possible to provide backwards compatibility by not making `type` and `value` be required positional arguments, but instead use `*args` and `**kwargs` and have the actual supported signature(s) be only present in `@overload` declarations. But I am not sure if this should be done, you are correct that this is more work than I originally thought.

Also, the `update` method should also be updated with the changed parameter name.

marcinplatek:
> Also, the update method should also be updated with the changed parameter name.

Fixed, also added setter for `.type_`.

marcinplatek:
> It is possible to provide backwards compatibility by not making type and value be required positional arguments, but instead use *args and **kwargs and have the actual supported signature(s) be only present in @overload declarations. But I am not sure if this should be done, you are correct that this is more work than I originally thought.

Please let me know if you think this should be added, I'm happy to figure it out.

erezsh:
I was thinking we'll just add a new param, `type_=None`, and change `inst.type = type` to `inst.type = type_ if type_ is not None else type`

The downside is we'll have to set type and value to None too. But it doesn't break anything.

Btw, why not just do `__match_args__ = __slots__ = (...)` ?

marcinplatek:
> I was thinking we'll just add a new param, type_=None, and change inst.type = type to inst.type = type_ if type_ is not None else type

That would make the typing inconsistent between `new` and the parameters.

The args thing seems to be ok.

> Btw, why not just do __match_args__ = __slots__ = (...) ?

Does order of `__slots__` matter? Because I think that `__match_args__` should be in the same order as args to `new` and `update`

> P.S. we'll also need a deprecation warning for any use of type_

On it.

lark/lexer.py
```diff
@@ -147,6 +147,8 @@ class Token(str):
     """
     __slots__ = ('type', 'start_pos', 'value', 'line', 'column', 'end_line', 'end_column', 'end_pos')
 
+    __match_args__ = ('type', 'value', 'start_pos', 'line', 'column', 'end_line', 'end_column', 'end_pos')
```
MegaIng, line 150: I missed this the first time: We shouldn't have everything in `__match_args__`, if you want to match by `start_pos` or `line` for some reason, just use keyword arguments. Allowing everything to be matched via positional arguments would just encourage unmaintainable code. IMO only `type` and `value` should be in `__match_args__`

marcinplatek:
Hey,
I've figured out that the easiest way to use `*args` and `**kwargs` with `@overload`, was to create methods for the old and new signatures and call them from the overloaded one.

Otherwise, I would have to manually write all the parsing of `*args` and `**kwargs` including errors etc. which would've been a huge mess.

Also, this way will allow you to replace the code with the new version easily.

One caveat is that the errors will contain one of the wrapped functions names (still should be easily readable), e.g.:


```
TypeError: Token._deprecated_new() got multiple values for argument 'type_'
```


Also `DeprecationWarning`s are supressed by default. I'm not sure if you want to change this, so I left it as it is.

lark/lexer.py
marcinplatek: fixed

marcinplatek:
Also, I'm pretty sure it would be nice to add proper typings to `Token.__new__` and to fix the typings for `Token` (most of the stuff there should be `Optional`), but it's not really in scope of the PR.

erezsh:
Why did you write
```python
    def __new__(cls, *args, **kwargs):
        if "type_" in kwargs:
            return cls._deprecated_new(*args, **kwargs)

        return cls._future_new(*args, **kwargs)
```

Instead of just - ?
```python
    def __new__(cls, *args, **kwargs):
        if "type_" in kwargs:
            kwargs["type"] = kwargs["type_"]

        return cls.same_new(*args, **kwargs)
```

> Also DeprecationWarnings are supressed by default.

What do you mean?

> most of the stuff there should be Optional

That's true! We should fix that.

marcinplatek:
> Why did you write
> 
>     def __new__(cls, *args, **kwargs):
>         if "type_" in kwargs:
>             return cls._deprecated_new(*args, **kwargs)
> 
>         return cls._future_new(*args, **kwargs)
> Instead of just - ?
> 
>     def __new__(cls, *args, **kwargs):
>         if "type_" in kwargs:
>             kwargs["type"] = kwargs["type_"]
> 
>         return cls.same_new(*args, **kwargs)

Both should be fine, but in the first one, errors will be more readable:

```
TypeError: Token._deprecated_new() got multiple values for argument 'type_'
```

makes more sense if you put in something like `Token("a", "A", type_="a")`, as opposed to (which would be the error in your version):

```
TypeError: Token._deprecated_new() got multiple values for argument 'type'
```

It's supposed to be removed, so I wouldn't worry too much about code duplication.

> > Also DeprecationWarnings are supressed by default.
> 
> What do you mean?

The warnings don't really show until we change the configs:
https://stackoverflow.com/a/20960427/7416999

At least they didn't show to me, when I tested this.

> Earlier you wrote about adding a type_ parameters.
> 
> > That would make the typing inconsistent between new and the parameters.
> 
> Can you please explain that a little bit more?

Unless we want to allow `Token.type` and `Token.value` to be optional, which would be a relatively big, and unnecessary change (as the `@overload` solution works), having `type` and `value` as optional in `__new__` would lead to undefined behavior or require us to add runtime errors like `TypeError("at least one of type and type_ has not to be None")`. Having types properly statically checked might save some errors from happening on runtime.

marcinplatek:
> That's true! We should fix that.

Fixed types and added typings to `Token.__new__`.

erezsh:
> errors will be more readable:

That's a bit of a nitpick. I can easily add this, and get an even better error -

```python
def __new__(cls, *args, **kwargs):
    if "type_" in kwargs:
        if "type" in kwargs:
             raise TypeError("Error: using both 'type' and the deprecated 'type_' as arguments.")
        kwargs["type"] = kwargs["type_"]

    return cls.same_new(*args, **kwargs)
```

> The warnings don't really show until we change the configs:

That link is from 2014. But when I test it in Python 3.10 -

```python
>>> import warnings
>>> warnings.warn('test', DeprecationWarning)
<stdin>:1: DeprecationWarning: test
```

marcinplatek:
> > errors will be more readable:
> 
> That's a bit of a nitpick. I can easily add this, and get an even better error -
> 
> ```
> def __new__(cls, *args, **kwargs):
>     if "type_" in kwargs:
>         if "type" in kwargs:
>              raise TypeError("Error: using both 'type' and the deprecated 'type_' as arguments.")
>         kwargs["type"] = kwargs.pop("type_")
> 
>     return cls.same_new(*args, **kwargs)
> ```

I don't know if this is necessarily a better error, as it does nothing to address my use case (which while valid, I admit is really minor). It's just a different error.

Nevertheless, I've added it, as I believe it makes a lot of sense here.

As I've said, both versions are fine to me, so I've changed to the one you prefer.

> That link is from 2014. But when I test it in Python 3.10 -
> 
> ```
> >>> import warnings
> >>> warnings.warn('test', DeprecationWarning)
> <stdin>:1: DeprecationWarning: test
> ```

That link is 100% valid still. Please look at python docs regarding warnings.
Also, you can test it by importing from a module that warns DeprecationWarning. As far as I understand, warning filters are different in the interactive shell.

The issue can be solved by adding:

```
warnings.simplefilter('default', DeprecationWarning)
```

to the top of this file, but it breaks the standalone version, which I really don't understand. So, I'm leaving this the way it is.

erezsh:
Thanks!

I see, you're right, the behavior changes for imported modules. I'll look into adding it.

lark/lexer.py
```diff
@@ -147,37 +148,106 @@ class Token(str):
...
             value if value is not None else self.value,
             self
         )
 
     @classmethod
     def new_borrow_pos(cls: Type[_T], type_: str, value: Any, borrow_t: 'Token') -> _T:
         return cls(type_, value, borrow_t.start_pos, borrow_t.line, borrow_t.column, borrow_t.end_line, borrow_t.end_column, borrow_t.end_pos)
+    
+    @property
+    def type_(self) -> str:
```
MegaIng, line 245: After thinking about this a bit more: Do we actually need this property for backwards compatibility? The `type_` didn't exists before, and adding a attribute that is deprecated-on-release is a bit pointless.

If we decide to keep it, it should raise a DeprecationWarning.

MegaIng (changes requested):

MegaIng:
In regards to Deprecation Warnings: They should be silenced by default, if someone cares they need to add it themselves We shouldn't do that in the library.

Issue #1182, opened 2022-08-22
Stand-alone tool raises ResourceWarning


shawnbrown:
I noticed that the standalone tool triggers a `ResourceWarning` when the warnings filter is set to "default" (`-W default`).

When redirecting the stdout, one warning is raised for the input file in the following example:

```shell
$ python -W default -m lark.tools.standalone json.g > json_parser.py
# /home/..../Projects/lark/lark/tools/standalone.py:196: ResourceWarning: unclosed file <_io.TextIOWrapper name='json.g' mode='r' encoding='utf-8'>
#   main()
```

When using the `--out` option, two warnings are raised--one for input and another for output:

```shell
$ python -W default -m lark.tools.standalone --out=json_parser.py json.g
# /home/..../Projects/lark/lark/tools/standalone.py:196: ResourceWarning: unclosed file <_io.TextIOWrapper name='json.g' mode='r' encoding='utf-8'>
#   main()
# /home/..../Projects/lark/lark/tools/standalone.py:196: ResourceWarning: unclosed file <_io.TextIOWrapper name='json_parser.py' mode='w' encoding='utf-8'>
#   main()
```

This seems to be happening because `argparse` is opening the FileType objects but not closing them. I don't think there's an existing test for the `main()` function and I'm not sure where I'd add one (otherwise I would have made a pull-request).

But I think this warning can be reliably silenced by explicitly closing these objects at the end of the `main()` function (see last two lines):

```python
# lark/tools/standalone.py

...

def main():
    make_warnings_comments()
    parser = ArgumentParser(prog="prog='python -m lark.tools.standalone'", description="Lark Stand-alone Generator Tool",
                            parents=[lalr_argparser], epilog='Look at the Lark documentation for more info on the options')
    parser.add_argument('-c', '--compress', action='store_true', default=0, help="Enable compression")
    if len(sys.argv) == 1:
        parser.print_help(sys.stderr)
        sys.exit(1)
    ns = parser.parse_args()

    lark_inst, out = build_lalr(ns)
    gen_standalone(lark_inst, out=out, compress=ns.compress)

    ns.out.close()           # <- CLOSE HERE!
    ns.grammar_file.close()  # <- CLOSE HERE!
```

erezsh:
Makes sense to me! Can you submit a PR for this?

shawnbrown:
PR submitted: #1183 (I'll close this on merge).

PR #1183, opened 2022-08-22
Close file-like objects to address ResourceWarning.

shawnbrown:
This is needed because ArgumentParser is opening its FileType arguments
but not closing them. When the program terminates, the default warnings
filter was emitting a "ResourceWarning: unclosed file" message.

See issue #1182.


MegaIng (approved):

erezsh:
Thanks!

PR #1190, opened 2022-09-15
refactor: add 'usedforsecurity=False' arg to hashlib.md5 usage

### LINKED ISSUE #1187: Add `usedforsecurity=False` argument to `hashlib.md5` instances to support running on FIPS-enabled systems
**Describe the bug**

Packages using `lark` fail when running in a FIPS-enabled environment due to use of the `hashlib.md5` function. MD5 is not a secure algorithm to begin with, but it isn't really being used for security purposes here.

To support running in a FIPS environment, we can add `usedforsecurity=False` to the `hashlib.md5` constructor. This argument is only available in Python 3.9+, but if implemented like `hashlib.new('md5', usedforsecurity=False)` it will not raise an exception on older versions of Python.

**To Reproduce**

I see it when trying to install packages with [poetry](https://github.com/python-poetry/poetry) on a FIPS-enabled system (running RHEL 8.5).

```
    [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS

  at ~/.pyenv/versions/3.9.13/lib/python3.9/site-packages/poetry/core/_vendor/lark/load_grammar.py:1315 in do_import
      1311│                         text = f.read()
      1312│             except IOError:
      1313│                 continue
      1314│             else:
    → 1315│                 h = hashlib.md5(text.encode('utf8')).hexdigest()
      1316│                 if self.used_files.get(joined_path, h) != h:
      1317│                     raise RuntimeError("Grammar file was changed during importing")
      1318│                 self.used_files[joined_path] = h
      1319│
```

cquick01:
This argument was introduced in Python 3.9, but using `hashlib.new()` prevents <3.9 versions from throwing an exception.

Tested running locally against 3.7, 3.9, and 3.10.

Closes #1187

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
         
 from .exceptions import ConfigurationError, assert_config, UnexpectedInput
 from .utils import Serialize, SerializeMemoizer, FS, isascii, logger
-from .load_grammar import load_grammar, FromPackageLoader, Grammar, verify_used_files, PackageResource
+from .load_grammar import load_grammar, FromPackageLoader, Grammar, verify_used_files, PackageResource, md5_digest
 from .tree import Tree
 from .common import LexerConf, ParserConf, _ParserArgType, _LexerArgType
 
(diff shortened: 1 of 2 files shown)


erezsh:
I didn't realize we call md5 from 3 different places. Would you mind refactoring it into a single function, like `load_grammar.md5_digest()`, and call that instead?

erezsh:
Thanks! Also, mypy fails with 

```
lark/load_grammar.py:1419: error: Unexpected keyword argument "usedforsecurity" for "new"  
```

Can you check if there's a way to fix this? If not, adding an "ignore" comment is good enough for me.

erezsh:
Looks good to me. I'll wait a bit to see if @MegaIng has anything to add.

MegaIng (approved):
I don't really have anything to add. IMO a version check is cleaner than a type ignore.

erezsh:
@cquick01 Thank you for your contribution!

jjtroberts:
Thank you all for working on this fix. When might the next release be cut?

erezsh:
It's probably about time for a new release! I'll try to do it early next week.

PR #1200, opened 2022-10-14
Add `__all__` to `__init__`

aspizu:


MegaIng (approved):

MegaIng (approved):
I personally find this `__all__` definition unnecessary, but if it silents type checkers, oh well

aspizu:
https://stackoverflow.com/questions/44834/what-does-all-mean-in-python

erezsh:
@MegaIng I agree, it's quite superfluous, but at the same time it seems to help, and I have no serious objection against it.

PR #1202, opened 2022-10-24
Update lexer option docs

benber86:
Current docs list `basic` as a lexer option, but this raises a configuration error exception:

`lark.exceptions.ConfigurationError: Got 'basic', expected one of ('standard', 'contextual', 'dynamic', 'dynamic_complete')`

Updated the docs to reflect that `standard` should be used instead.


MegaIng (approved):

erezsh:
@benber86 The correct option is "basic", ever since v1. What version of Lark are you using?

PR #1203, opened 2022-10-25
PropagatePositions: Allow any object to carry the metadata, by returning it in __lark_meta__()

erezsh:


MegaIng:
Didn't we recently talk about not inventing `__dunder__` methods? I think we decided to avoid that.

MegaIng:
Hmm, I can't find it. I am personally not that invested, I think someone else argued to follow the python spec, but I also think it's not that important.

MegaIng (approved):

PR #1209, opened 2022-11-02
setup.cfg: Replace deprecated license_file with license_files

mgorny:
Replace the deprecated `license_file` key with `license_files`. The latter has the same semantics and is available since setuptools 42.0.0 (that lark already requires).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,3 @@
 [metadata]
-license_file = LICENSE
+license_files = LICENSE
 python_requires = >=3.6


erezsh:
Is there a reason to change this? Is `license_file` ever going to get dropped?

mgorny:
> Is there a reason to change this?

Right now it triggers verbose deprecation warnings that trigger our QA pipelines.

> Is `license_file` ever going to get dropped?

Hard to tell. I've asked setuptools a more general question about timeline for dropping deprecated stuff, and got no answer. However, they definitely did remove some deprecated stuff in the past.

erezsh:
> verbose deprecation warnings 

Fair enough.

erezsh:
No problem, thanks for making Lark a tiny bit better.

PR #1216, opened 2022-11-20
Fix AmbiguousExpander (Issue #1214)

### LINKED ISSUE #1214: Missing ambiguities when using inline rules
**Describe the bug**

With some grammars, Earley parser does not return all ambiguities with `ambiguity='explicit'`. 
However, when removing the inlining underscore, all ambiguities are returned. 

Potentially related to #536. 

**To Reproduce**

```
start: _field+

_field: f1 | f2 | f3

f1: INT
f2: INT "M"?
f3: INT "M"

%ignore WS
%import common.WS
%import common.INT
```

When parsing string `1M 2` with `ambiguity='explicit'`, the result misses ambiguities `f2 f2` and `f3 f1`. 

```
>> t = parser.parse('1M 2')
>> print(t.pretty())

_ambig
  start
    f2	1
    f1	2
  start
    f3	1
    f2	2
```

However, removing `_` from `field` fixes it, at the expense of the additional tree level which I wanted to avoid. 

```
start: field+

field: f1 | f2 | f3

f1: INT
f2: INT "M"?
f3: INT "M"

%ignore WS
%import common.WS
%import common.INT
```

```
>> t = parser.parse('1M 2')
>> print(t.pretty())

start
  _ambig
    field
      f2	1
    field
      f3	1
  _ambig
    field
      f1	2
    field
      f2	2
```

It's great that the workaround works, at least in this case. But with a complex grammar including inlined rules at several levels, it's difficult to spot the issue before realising that some results are missing in production.

chanicpanic:
Fixes #1214.
Surprisingly, this bug has been around since 0.6.6.

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -8,7 +8,7 @@ from .visitors import _vargs_meta, _vargs_meta_inline
 
 ###{standalone
 from functools import partial, wraps
-from itertools import repeat, product
+from itertools import product
 
 
 class ExpandSingleChild:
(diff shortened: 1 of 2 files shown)


erezsh:
Wow, that's almost 4 years ago! Wild.

Thanks for fixing it!!

PR #1218, opened 2022-11-26
Update README.md

shomodj:
String quoted in example is not working, it should have ```r```


MegaIng:
The `r` does not make a difference, the two strings are exactly the same.

PR #1225, opened 2022-12-05
Use generator instead of list expand or add method

jmishra01:
We can use generator to create pretty string of Tree in place of list. This PR reduce memory usage.

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -70,18 +70,15 @@ class Tree(Generic[_Leaf_T]):
     def _pretty(self, level, indent_str):
         yield indent_str*level
         yield self._pretty_label()
-        
         if len(self.children) == 1 and not isinstance(self.children[0], Tree):
-            for i in '\t', '%s' % (self.children[0],), '\n':
-                yield i
+            yield from (i for i in ('\t', '%s' % (self.children[0],), '\n'))
         else:
             yield '\n'
             for n in self.children:
                 if isinstance(n, Tree):
                     yield from n._pretty(level+1, indent_str)
                 else:
-                    for i in indent_str*(level+1), '%s' % (n,), '\n':
-                        yield i
+                    yield from (i for i in (indent_str*(level+1), '%s' % (n,), '\n'))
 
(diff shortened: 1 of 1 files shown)


erezsh:
I like this change.

You can turn a few more of the "yield"s to "yield from"s.

jmishra01:
@erezsh Done!

erezsh:
What I meant is that you don't need any for loop..

```python
yield from (i for i in ('\t', '%s' % (self.children[0],), '\n'))
```

can be

```python
yield from ('\t', '%s' % (self.children[0],), '\n')
```

etc.

jmishra01:
Thank you for review, I change few lines,  use yield with f-string method.

lark/tree.py
```diff
@@ -149,13 +148,15 @@ def iter_subtrees_topdown(self):
         Iterates over all the subtrees, return nodes in order like pretty() does.
         """
         stack = [self]
+        stack_append = stack.append
```
erezsh, line 151: I assume you added this for performance. Did you benchmark it on a realistic use-case?

erezsh: Did you try also changing `Tree` to a local variable (`_Tree` or such)?

jmishra01: Not such type of thing, but I create child class of it, and enhance some part for better performance. 
Also working on multi-threading, with cythonise some part of code and logic.

erezsh: I meant specifically inside this function, to skip the global lookup.

jmishra01: > Also, you're aware of https://github.com/lark-parser/lark_cython?

Yes. 
I recommend, we should start implement lark in C++ or Rust, so with WASM and  Cython (or Pybind11), we can use optimise version of it in Python and JS world.

jmishra01: I'm interest to take both the tasks, if you allow.

erezsh:
Thanks!

lark/tree.py
jmishra01: Sure, I will take care of it.

PR #1227, opened 2022-12-05
LALR analysis performance improvement

jmishra01:
Use dictionary comprehension to improve performance of lalr analysis.

### DIFF
--- a/lark/__init__.py
+++ b/lark/__init__.py
@@ -14,7 +14,7 @@ from .tree import ParseTree, Tree
 from .utils import logger
 from .visitors import Discard, Transformer, Transformer_NonRecursive, Visitor, v_args
 
-__version__: str = "1.1.4"
+__version__: str = "1.1.5"
 
 __all__ = (
     "GrammarError",


MegaIng:
What is the performance improvement of this? I.e. what is the measurable increase?

erezsh:
@jmishra01 Generally, when creating a PR claiming to provide a speed improvement, you should also provide some evidence that it is indeed the case. At the very least, the numerical results of your benchmark with and without this single change.

I think this change is a bit cleaner, so I'm inclined to accept it. But I'm not convinced it's actually faster.

jmishra01:
@erezsh Understood! 
Please check attached image.
List or dictionary comprehension always faster than normal loop method.

image

MegaIng:
But what is the impact on the whole of lark? Isolated pieces of code can be optimized to an absurd amount without actually improving anything in practice (in your example, `dict.fromkeys` would probably be faster still)

erezsh:
If we rename the PR to "style improvements for code", can we agree it's acceptable?

erezsh:
Also @jmishra01 please flatten this to one single commit.

MegaIng:
For style I would also prefer `dict.fromkeys(X, 0)` instead of `dict(zip(X, [0]*len(X)))`. But yes, otherwise it's fine.

jmishra01:
> @jmishra01 Looks like a sync issue. Make sure to fetch and rebase over the latest master.

Kindly review again.

erezsh:
@jmishra01 I agree with @MegaIng that `dict.fromkeys(X, 0)` is the best of the 3 versions.

PR #1232, opened 2022-12-18
Updated Python grammar list literal to support `[*x]`

evtn:
Previous version used `_testlist_comp` which allowed for either one `test` or 2 and more `test_or_star_expr`.
This version also allows a list with one `star_expr` which is valid both in Python and in official Python grammar.
Moreover it merges rules used in `set` and `list` (since those terminals differ only in that set literal cannot be empty)

### DIFF
--- a/lark/grammars/python.lark
+++ b/lark/grammars/python.lark
@@ -197,11 +197,11 @@ AWAIT: "await"
 ?atom: "(" yield_expr ")"
      | "(" _tuple_inner? ")" -> tuple
      | "(" comprehension{test_or_star_expr} ")" -> tuple_comprehension
-     | "[" _testlist_comp? "]"  -> list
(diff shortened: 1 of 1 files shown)


MegaIng (approved):
It would be really nice if we could find a syntax test bench for python somewhere to catch all these edge cases. But I don't think that exists.

evtn:
Well, there is a suite that is used by PyPy, which is MIT-licensed and probably will be a good place to start
https://github.com/mozillazg/pypy/tree/master/pypy/interpreter/pyparser/test (this is a mirror, but still)

I probably could spend some time on this, since my current project requires a fully working Python grammar anyways

erezsh:
Thank you, @evtn ! If you would like to set up tests for the Python grammar, we'll be happy to include them in our test suite!

PR #1235, opened 2023-01-02
Fixed off-by-one bytes bug in qscintilla when using unicode

traverseda:
If we just grab self.parent().text is causes off by one errors in all downstream tokens, if you have unicode glyphs that are longer than one byte.


erezsh:
Using the latest version of PyQt5, I'm not getting any error when using the symbol. (I just updated the version, see https://github.com/lark-parser/lark/pull/1243)

However, I am getting suspiciously slow response time when editing a string in which that symbol exists. But your PR doesn't seem to affect it either way, it happens in both cases.

PR #1251, opened 2023-02-13
Replace md5 hashing with sha256.

### LINKED ISSUE #1250: Lark doesn't run on FIPS enabled systems on older versions of python 3
**Describe the bug**

Due to md5 ciphers being disabled on FIPS enabled systems, Lark won't work due to the md5_digest method in the load_grammer module. On Python 3.9 and up this will work, but not all users running FIPS are on 3.9 and since Lark purports to support python as low as 3.6, we should probably make the module FIPS compliant by simply removing MD5 hashing. This change would build on the work already put into place by #1187 

**To Reproduce**

On a RHEL 7 system running Python 3.6, when opening a grammar file with `lark_parser = Lark.open(GRAMMAR_FILE)`, I get the following error (note that the top of the stack is redacted):

```
...
  File "/usr/lib/python3.6/site-packages/lark/lark.py", line 567, in open
    return cls(f, **options)
  File "/usr/lib/python3.6/site-packages/lark/lark.py", line 349, in __init__
    self.grammar, used_files = load_grammar(grammar, self.source_path, self.options.import_paths, self.options.keep_all_tokens)
  File "/usr/lib/python3.6/site-packages/lark/load_grammar.py", line 1410, in load_grammar
    builder.load_grammar(grammar, source)
  File "/usr/lib/python3.6/site-packages/lark/load_grammar.py", line 1250, in load_grammar
    self.do_import(dotted_path, base_path, aliases, mangle)
  File "/usr/lib/python3.6/site-packages/lark/load_grammar.py", line 1317, in do_import
    h = md5_digest(text)
  File "/usr/lib/python3.6/site-packages/lark/load_grammar.py", line 1423, in md5_digest
    return hashlib.md5(s.encode('utf8')).hexdigest()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for fips
```

wilrodriguez:
Fixes #1250

This change is intended to replace the md5 hashing currently utilized to validate grammar caches with sha256 hashing, which will help mitigate situations where md5 is being blocklisted, such as by FIPS-mode

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -20,7 +20,7 @@ if TYPE_CHECKING:
 
 from .exceptions import ConfigurationError, assert_config, UnexpectedInput
 from .utils import Serialize, SerializeMemoizer, FS, isascii, logger
-from .load_grammar import load_grammar, FromPackageLoader, Grammar, verify_used_files, PackageResource, md5_digest
+from .load_grammar import load_grammar, FromPackageLoader, Grammar, verify_used_files, PackageResource, sha256_digest
 from .tree import Tree
 from .common import LexerConf, ParserConf, _ParserArgType, _LexerArgType
 
@@ -288,7 +288,7 @@ class Lark(Serialize):
             grammar = read()
 
         cache_fn = None
-        cache_md5 = None
+        cache_sha256 = None
         if isinstance(grammar, str):
             self.source_grammar = grammar
             if self.options.use_bytes:
@@ -303,7 +303,7 @@ class Lark(Serialize):
                 options_str = ''.join(k+str(v) for k, v in options.items() if k not in unhashable)
                 from . import __version__
                 s = grammar + options_str + __version__ + str(sys.version_info[:2])
-                cache_md5 = md5_digest(s)
+                cache_sha256 = sha256_digest(s)
 
                 if isinstance(self.options.cache, str):
                     cache_fn = self.options.cache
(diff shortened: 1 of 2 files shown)


erezsh:
I have no problem with this change, but we have to make sure that in attempting to fix one edge case, we're not creating new breaking changes for other edge cases. (i.e. other python versions or operating systems)

wilrodriguez:
I was mildly concerned about that as well. That said, sha256 is supported as part of the core library in all the versions of python that lark currently supports and this hashing library is only used within lark. The only downside I'd see is that it would need to recache files after the upgrade to the new version, but that's hardly much of a concern. This method should only be used internally. It's certainly not like it provides any novel functionality that a user wouldn't be able to more easily get from just calling hashlib directly. So, I'm not sure it's really anything to worry about.

MegaIng:
The recaching already happens since the lark version is part of what get's hashed.

wilrodriguez:
Even better.

erezsh:
@wilrodriguez Thanks for contributing!

Erotemic:
Small comment on this. I'm +1 for anything modern over md5, and sha256 on modern machines is faster than md5. I'm not sure how much speed is a factor here, but there are hashers that are faster than sha256 - especially if we don't need them to be cryptographic. However some of these faster hashers are not FIPS validated (although blake3 probably should be): 

image

Here is a small benchmark: 

```python
import ubelt as ub
import timerit

ti = timerit.Timerit(100_000, bestof=100, verbose=2)

hashers = ['md5', 'sha256', 'blake3', 'xxh64']

data = 'foobar' * 10_000

for hasher in hashers:
    for timer in ti.reset(f'{hasher=}'):
        with timer:
            ub.hash_data(data, hasher=hasher)

mean_rankings = ub.udict(ti.rankings['mean']).sorted_values()
print(ub.urepr(mean_rankings, nl=2, align=':'))
```

And the larger benchmark that produced the image is here: https://github.com/Erotemic/timerit/blob/dev/1.0.2/examples/benchmark_hashers.py

Likely not a big deal, but I thought I'd chime in.

PR #1252, opened 2023-02-16
Fix 1154

### LINKED ISSUE #1154: exception str() failed in `UnexpectedCharacters` because of 'PatternRE' has no attribute 'raw'
**Describe the bug**

Lark v1.1.2

The standalone parser generated with `python -m lark.tools.standalone` appears to have an issue with the `PatternRE`/`PatternStr` subclasses because when I get an `UnexpectedCharacters` error, it's exception formatting fails because of 

```
kraken.lang._generated_parser.UnexpectedCharactersTraceback (most recent call last):
  File ".venvs/3.9/lib/python3.9/site-packages/kraken/lang/_generated_parser.py", line 206, in __str__
    message += self._format_expected(self.allowed)
  File ".venvs/3.9/lib/python3.9/site-packages/kraken/lang/_generated_parser.py", line 143, in _format_expected
    expected = [d[t_name].user_repr() if t_name in d else t_name for t_name in expected]
  File ".venvs/3.9/lib/python3.9/site-packages/kraken/lang/_generated_parser.py", line 143, in <listcomp>
    expected = [d[t_name].user_repr() if t_name in d else t_name for t_name in expected]
  File ".venvs/3.9/lib/python3.9/site-packages/kraken/lang/_generated_parser.py", line 1187, in user_repr
    return self.pattern.raw or self.name
AttributeError: 'PatternRE' object has no attribute 'raw'
```

**To Reproduce**

* Take some Lark grammar
* Generate a standalone parser
* Feed it some input that results in an `UnexpectedCharacters` error

If it helps, I attached the standalone parser that was generated on my side:

[_generated_parser.py.zip](https://github.com/lark-parser/lark/files/8782534/_generated_parser.zip)

MegaIng:
Fixes #1154 

The problem is that the `.raw` attribute wasn't being serialized. Now it is, so it should remove this particular problem. Also contains a test that should detect general problems in the error message generation as well.


erezsh:
LGTM

@ornariece Does this fix your use-case?

ornariece:
@erezsh yup it does! thanks for the quick fix @MegaIng.

PR #1261, opened 2023-03-09
Added strict-mode, enabled using the strict=True flag

erezsh:


MegaIng (approved):
Looks good, a few more tests might be nice.

For a future PR, it might be nice to transform `LarkOptions` and the two `Config` classes into `dataclasses`. Would remove some code duplication

erezsh:
Definitely, I'm going to add some tests.

Sure, I'm pro dataclasses. But they have to be backported for Python 3.6.

MegaIng:
There is such a backport, I am using it in `interegular`.

erezsh:
Yes, I know. That's why I said I'm okay with it.

PR #1262, opened 2023-03-09
Freeze interegular version to 0.2.4 (until fixed)

erezsh:


MegaIng:
This is now fixed with `0.2.6` and I also setup CI for `interegular` with the same python versions as `lark`: https://github.com/MegaIng/interegular/blob/master/.github/workflows/tests.yml

erezsh:
@MegaIng Cool. You can create a PR to use the new version.

PR #1265, opened 2023-03-09
Docs: Added section about strict-mode to how-to-use

erezsh:


MegaIng (approved):
Looks good!

Side note: I am currently looking through the hundreds of lark grammars available on github and found out that sometimes the current example generation algorithm can sometimes take multiple GBs of memory and take a few dozen seconds to actually find something, even if the intersection test was very fast. I gotta change that...

erezsh:
Cool, well done on taking the initiative and testing them.

I think it will come down to either optimizing the algorithm, or setting a maximum number of steps, after which it exits without an example.

MegaIng:
I will actually do both I think. Stopping after a certain number of steps is easy and reliable no matter how the exact regex(es) look like. Optimizations can only provide so much I think, it will always be possible to create a very slow example I think.

PR #1267, opened 2023-03-16
Added progress bar recipe

MegaIng:
As suggested on gitter.


erezsh:
Looking good! Just one thing, I think we should emphasize that this only works for LALR. Right now, it might seem like it works for both parsers.

MegaIng:
Aha true.

PR #1270, opened 2023-04-17
Update README.md

BrandonBahret:
Added a section for conference talks.

I enjoy watching conference talks whenever I begin using new packages. It would be useful to create a list of such talks in the readme or elsewhere.

[Using Lark to Parse Text - Robin Reynolds-Haertle (PyCascades 2023)
](https://www.youtube.com/watch?v=CeOtqlh0UuQ)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ Most importantly, Lark will save you time and prevent you from getting parsing h
 - Blog post: [How to write a DSL with Lark](http://blog.erezsh.com/how-to-write-a-dsl-in-python-with-lark/)
 - [Gitter chat](https://gitter.im/lark-parser/Lobby)
 
+### Talks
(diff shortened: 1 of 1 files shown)


MegaIng (approved):

erezsh (changes requested):
Very cool. But I don't think that is the right place for the link.

Makes more sense to me to place it in the documentation: https://github.com/lark-parser/lark/blob/master/docs/index.rst

MegaIng:
@erezsh I consider Readme a good idea since it's basically a project using Lark IMO. But also, we can link it in both places. It is also a decent tutorial.

erezsh:
In what way is it a project? It looks like only a tutorial.

erezsh:
Added link

PR #1291, opened 2023-06-25
Update parse_tree_builder.py

jmishra01:
List function is not required, and rule check in callback dict should be after loop

### DIFF
--- a/lark/parse_tree_builder.py
+++ b/lark/parse_tree_builder.py
@@ -331,7 +331,7 @@ class ParseTreeBuilder:
         self.ambiguous = ambiguous
         self.maybe_placeholders = maybe_placeholders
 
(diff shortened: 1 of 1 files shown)


erezsh:
Hello! Thanks for the PR. You're right about moving the rule check up in the loop. But I prefer that you avoid changing lines only for cosmetic reasons, especially when it's very subjective, like removing blank lines that separate code blocks.

jmishra01:
Revert changing lines changes.

jmishra01:
@erezsh Thanks. Understood the reason behind calling `list()`. I'm closing this PR.

erezsh:
@jmishra01 You can re-open with just the rule-check. I'll be happy to merge that in.

Issue #1293, opened 2023-06-28
propagate_positions isn't well documented


ningit:
**Describe the bug**

The keyword argument `propagate_positions` of [Lark](https://lark-parser.readthedocs.io/en/latest/classes.html#lark.Lark) constructor is not obeyed. The attributes `line`, `column`, `end_line`, and `end_column` are not available in the `Tree` objects returned by `Lark.parse`.

**To Reproduce**

In the following code, the last two lines produce an `AttributeError: 'Tree' object has no attribute 'line'`, while they should return 1.

```python
import lark

grammar = '''
	%import common.WS
	%ignore WS

	%import common.INT -> NUMBER

	start: expr

	expr: NUMBER
	    | NUMBER "+" NUMBER
'''

parser = lark.Lark(grammar, parser='lalr', propagate_positions=True)
tree = parser.parse('1 + 2')

print(tree.children[0].children[0].line)
print(tree.children[0].line)
print(tree.line)
```

erezsh:
It's tree.meta

ningit:
Many thanks and sorry. Now I see this is mentioned in the documentation for the constructor of the [`Tree`](https://lark-parser.readthedocs.io/en/latest/classes.html#lark.Tree) class.

Anyhow, making explicit that these attributes are available through `meta` in the documentation of the `propagate_positions` option itself may avoid confusion and help users.

erezsh:
@ningit  I added a PR https://github.com/lark-parser/lark/pull/1294

ningit:
Great, it is now clearer. I think the PR solves the issue.

erezsh:
Great! Added even more explanations too.

PR #1294, opened 2023-06-29
propagate_positions: Bugfix + Better documentation

erezsh:
Addresses issue #1293

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -87,7 +87,9 @@ class LarkOptions(Serialize):
     transformer
             Applies the transformer to every parse tree (equivalent to applying it after the parse, but faster)
     propagate_positions
-            Propagates (line, column, end_line, end_column) attributes into all tree branches.
+            Propagates positional attributes into the 'meta' attribute of all tree branches.
+            Sets attributes: (line, column, end_line, end_column, start_pos, end_pos,
+                              container_line, container_column, container_end_line, container_end_column)
             Accepts ``False``, ``True``, or a callable, which will filter which nodes to ignore when propagating.
     maybe_placeholders
             When ``True``, the ``[]`` operator returns ``None`` when not matched.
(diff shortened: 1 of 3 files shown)


lark/lark.py
```diff
@@ -87,7 +87,9 @@ class LarkOptions(Serialize):
     transformer
             Applies the transformer to every parse tree (equivalent to applying it after the parse, but faster)
     propagate_positions
-            Propagates (line, column, end_line, end_column) attributes into all tree branches.
+            Propagates positional attributes into the 'meta' attribute of all tree branches.
+            Sets attributes: (line, column, end_line, end_column, start_pos, end_pos,
+                              container_line, container_column, container_end_line, container_end_column)
```
MegaIng, line 92: Is it documented anywhere what the `container_*` versions mean?

erezsh: Hmm it's not.. Just this comment in the code -

```python
            # Calculate positions while the tree is streaming, according to the rule:
            # - nodes start at the start of their first child's container,
            #   and end at the end of their last child's container.
            # Containers are nodes that take up space in text, but have been inlined in the tree.
```

lark/tree.py
```diff
@@ -47,7 +47,9 @@ class Tree(Generic[_Leaf_T]):
         data: The name of the rule or alias
         children: List of matched sub-rules and terminals
         meta: Line & Column numbers (if ``propagate_positions`` is enabled).
-            meta attributes: line, column, start_pos, end_line, end_column, end_pos
+            meta attributes: (line, column, end_line, end_column, start_pos, end_pos,
+                              container_line, container_column, container_end_line, container_end_column)
+            container_* attributes consider all symbols, including those that have been inlined in the tree.
```
MegaIng, line 52: I think an example of the difference might be nice. I have a hard time understanding where this makes a difference.

MegaIng (approved):

lark/tree.py
erezsh: Added.

PR #1302, opened 2023-07-17
Improve documentation of `v_args` `meta` option.

richardxia:
When using `@v_args(meta=True)`, visitor methods will receive a `meta` argument followed by a `children` argument in that order. The documentation previously described the arguments in the opposite order, which can be misleading to someone reading the docs, especially because there was no example of `@v_args(meta=True)`.

This updates the documentation to describe the arguments in the order they are passed in to the visitor methods, and adds a new example for `@v_args(meta=True)`.

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -534,6 +534,7 @@ def v_args(inline: bool = False, meta: bool = False, tree: bool = False, wrapper
 
                 @v_args(meta=True)
                 def mul(self, meta, children):
+                    logger.info(f'mul at line {meta.line}')
                     left, right = children
                     return left * right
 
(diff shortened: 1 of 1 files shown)


erezsh:
Thanks, always glad to improve documentation.

One nitpick - if we already give an example with `meta`, it should at least use that variable in some way? Maybe even add `logger.info(f'mul at line {meta.line}')` or something of the sort.

richardxia:
No worries, just pushed up a new commit that adds an actual usage of the `meta` argument: https://github.com/lark-parser/lark/pull/1302/commits/fc5da2209fce2338de9b2332192e683057389654

MegaIng (approved):

erezsh:
Thanks!

richardxia:
No problem, thank you!

Issue #1304, opened 2023-07-20
Possible bug in PropagatePositions class


stellabelin:
Hi, thanks for your work with this library! When updating to lark 1.1.6 we started getting issues with our parser, and after digging I noticed that it's possible this line inputs the wrong variable: https://github.com/lark-parser/lark/blob/e4df0249c489d20c64222fd293b5cca4c3be4f27/lark/parse_tree_builder.py#L65
It should use `last_meta` instead of `first_meta`, right? When I tried changing it our parser worked as expected.

erezsh:
Hi Stella! I believe you're right. Very embarrassing. Thanks for letting me
know!

erezsh:
Fixed and released 1.1.7

Thanks again!

stellabelin:
Great, thank you the fix and the quick response!

PR #1305, opened 2023-07-20
Bugfix in propagate positions

erezsh:
Solves issue #1304


MegaIng (approved):
Looks fine. Tests would be nice to reduce the chance of this happening again.

Issue #1301, opened 2023-07-17
Docs are ahead of stable release


nat-n:
Hello, firstly thanks for for maintaining this fantastic library.

I've just spent some time trying to work out how to get the strict mode to work as described in the docs, until I realised it hasn't been released yet! Now maybe I should have noticed that latest != stable, but this is easy to miss.

Of course now I know I can get what I want by running: `pip install pip install git+https://github.com/lark-parser/lark interegular`, but 15 minutes ago I was starting to suspect a regression in lark was to blame.

May I suggest that it would be preferable to avoid this kind of situation (especially for new users like me), by only publishing docs for releases?

One approach would be to only publish docs from the _main_ branch, and do development on a separate _development_ branch which only merges to main for releases. Another approach would be to keep development on main, but only build docs from tagged commits (releases). There are various similar approaches and tradeoffs to consider, but I think it's worth thinking about.

Also, it looks like this project is quite active, but the last release was more than 6 months ago. I suspect getting features out more regularly would be valuable to users, and this is another a reason to consider improving the branching/automation model.

I may be able to help implement a solution if there's agreement on the direction :)

erezsh:
Hello,

Thanks for the compliments!

First of all, you're right that it's been too long without a release! I just released 1.1.6.

I also agree that ideally the docs should be synced to the right version. I'm okay with moving development to a "dev" branch, and merging it only on release. Especially if the dev branch will build to its own branch of documentation, so we can point early adopters there for the latest docs.

PR #1306, opened 2023-07-20
Version bump (1.1.7)

erezsh:
Trying the new workflow of making releases as PRs, to address issue #1301, and also give other maintainers and users the opportunity to comment about new releases as they come.

I plan to push this release today, since it's just a bugfix, but next time I will also give a timeline of a week or so ahead.

Open to thoughts and comments about this.

@MegaIng @nat-n

### DIFF
--- a/lark/__init__.py
+++ b/lark/__init__.py
@@ -14,7 +14,7 @@ from .tree import ParseTree, Tree
 from .utils import logger
 from .visitors import Discard, Transformer, Transformer_NonRecursive, Visitor, v_args
 
-__version__: str = "1.1.6"
+__version__: str = "1.1.7"
 
 __all__ = (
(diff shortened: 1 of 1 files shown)


erezsh:
Also, should we be working from a `dev` branch, and create a PR when we're ready for release, or work from a `v1.x.x` branch, each time moving to a new one?

MegaIng:
I think I prefer a dev branch that get's merged into master. `v1.x.x` style branches would make sense IMO when we are potentially releasing new versions for older subversions (i.e. if we release a bug fix for `1.1` after `1.2` has been released). I don't think this will be that necessary for lark at the moment, so there is no reason to have the infra structure for that.

nat-n (approved):

PR #1308, opened 2023-07-20
Include .lark files in package data

### LINKED ISSUE #1307: Wheels built using the new pyproject file don't include .lark files
**Describe the bug**
When building lark 1.1.6 or later using a pep517-compatible tool, the resulting wheel does not include the `.lark` files in `lark/grammars/`

**To Reproduce**
Build the package with, for example, [build][1]; then, install the resulting package and check `lark/grammars/` ( or try using something that depends on `common.lark` )

Simple Docker reproduction example:
```
docker run --rm -it alpine:edge sh -c 'apk add py3-build py3-wheel py3-setuptools git unzip; git clone -b 1.1.7 https://github.com/lark-parser/lark; cd lark; python3 -m build --wheel --no-isolation; unzip -l ./dist/lark-1.1.7-py3-none-any.whl | grep lark/grammars'
```

[1]: https://pypi.org/project/build/

ptrcnull:
Fixes #1307

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ packages = [
 include-package-data = true
 
 [tool.setuptools.package-data]
+"*" = ["*.lark"]
 lark = ["py.typed"]
 
 [tool.setuptools.dynamic]


MegaIng:
In comparison to the old `setup.py` file, it looks like `*.md` is a pattern that was included back then but isn't any more. Do we still need that.

https://github.com/lark-parser/lark/blob/7d9cfa6ed09a2b125e97e3a1810d9c6443f1718e/setup.py#L20

erezsh:
@MegaIng I don't know. It looks okay in my builds, but apparently I can't rely on that.

MegaIng:
If we want them to be included, they need to be added, even if just for documentation.

erezsh:
Readme.md is already mentioned by name, and docs/* should take care of the rest.

Also, right now we don't include examples/README.rst, or any of the examples subfolders.

MegaIng:
> Readme.md is already mentioned by name, and docs/* should take care of the rest.

Are we looking at the same file? I don't see either of those included

erezsh:
They are in MANIFEST.in

MegaIng:
Aha right.

I don't have experience with `setuptools-scm`, so no real opinion on that.

But if we are using `MANIFEST.in`, shouldn't we add the grammars there instead of as package data?

erezsh:
@KOLANICH @henryiii Do you guys have an opinion about this?

KOLANICH:
1. In order to include the lark files `tool.setuptools.include-package-data=true` should be enough. `tool.setuptools.package-data` section is needed when it is `false`. When it is `true`, it captures all files in dirs having `__init__.py` automatically. In my projects I rely on this behaviour, and the files that shouldn't be within wheels are just out of the main package. Since it is already `true`, the section can be just purged.

2. :+1: for using "setuptools_scm>7". `7` because since that version it can work with sdists generated from git repos.
3. `MANIFEST.in` is relevant only to sdists. But when there is no `-s` or `-w`, `build` builds an sdist first, and then builds a wheel from it.
4. about the readme, I suspect it may make sense to use the file as a readme, rather than hardcode a string into `pyproject.toml`.
5. you cannot include files that are not a part of a package into wheels (except the ones that go into metadata automatically). And it shouldn't be done. I think the HTML docs and tests shouldn't be within wheels.
6. I also propose to eliminate the manual enumeration of the subpackages (`tool.setuptools.packages`) and just use

```toml
[tool.setuptools.packages.find]
include = ["lark", "lark.*"]
```

KOLANICH:
2. Fix: `setuptools_scm[toml]>7`

henryiii:
My preference is to use hatching. That includes everything that is not git ignored by default, and can be configured easily in pyproject.toml without a separate MANIFEST.in file. It's also simpler and faster than setuptools.

For setuptools, package-data and MANIFEST.in are not interchangeable. You place things that go into the SDist into MANIFEST.in. You place things that go into the wheel in package-data. For simplicity, it will also be added to your MANIFEST if it's in package-data (since you need it in the SDist to get it in the wheel), but the question is, does it go in the wheel? If so, then use package-data. Otherwise use MANIFEST.in or setuptools_scm. 

Not a huge fan of setuptools_scm's file finder. You can't turn it off if it's installed, so it affects every other package, even if they have no idea it exists. This doesn't affect isolated builds, of course, but it does affect systems like Spack where there's only one build environment. However, it's a very minor worry, and I do use setuptools_scm for versioning lots of places, which technically installs the file finder too (at least for now).

PR #1309, opened 2023-07-21
Populate the `Token.end_*` fields for ignored tokens

jtbraun:
Lexer callbacks can make use of the `Token.end_*` fields too, even for Ignored tokens. Always populate them.

This preserves the behavior where `Token` objects are only created if they are used.


MegaIng (approved):

erezsh:
Thank you for contributing!

Issue #1325, opened 2023-08-19
Problem matching the body of annotations like `@Annotation(title, some longer description)`


jjalowie:
**What is your question?**

I'm trying to write a parser for annotations like `@Annotation(title, some longer description)`. I have trouble achieving consistent parsing of the `description` rule in the below grammar. Given `@Annotation(title, some longer description)`, the `description` rule sometimes matches `some longer description` and sometimes it matches ` some longer description` (note the leading space). I want to always enforce matching without leading whitespaces.

**If you're having trouble with your code or grammar**

I believe it's either a problem with the below grammar or a bug in the parsing process.

Code reproduction:
```
# File: parser.py

import lark

grammar = """
start: "@Annotation" "(" title "," description ")"
title: /[^,]+/
description: /[^)]+/

%import common.WS
%ignore WS
"""

text = "@Annotation(title, some longer description)"

parser = lark.Lark(grammar)
ir = parser.parse(text)
print(ir.children[1])
```

**Explain what you're trying to do, and what is obstructing your progress.**

It seems to me that ignoring the `WS` rule sometimes takes precedence over the `description` rule and sometimes vice versaand it happens randomly. Output of executing the same script a few times gives (again, note the leading space that sometimes gets matched for `some longer description`:
```
user@machine:/dir> python parser.py 
Tree(Token('RULE', 'description'), [Token('__ANON_2', ' some longer description')])
user@machine:/dir> python parser.py 
Tree(Token('RULE', 'description'), [Token('__ANON_2', 'some longer description')])
user@machine:/dir> python parser.py 
Tree(Token('RULE', 'description'), [Token('__ANON_2', ' some longer description')])
user@machine:/dir> python parser.py 
Tree(Token('RULE', 'description'), [Token('__ANON_2', 'some longer description')])
user@machine:/dir> python parser.py 
Tree(Token('RULE', 'description'), [Token('__ANON_2', ' some longer description')])

erezsh:
Should be solved in the latest master.

PR #1327, opened 2023-08-23
Earley now uses OrderedSet for better output stability

erezsh:
Addresses issue #1325, and others

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -62,7 +62,7 @@ class LarkOptions(Serialize):
     regex: bool
     g_regex_flags: int
     keep_all_tokens: bool
-    tree_class: Any
+    tree_class: Callable
     parser: _ParserArgType
     lexer: _LexerArgType
     ambiguity: 'Literal["auto", "resolve", "explicit", "forest"]'
@@ -73,6 +73,7 @@ class LarkOptions(Serialize):
     edit_terminals: Optional[Callable[[TerminalDef], TerminalDef]]
     import_paths: 'List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]'
     source_path: Optional[str]
+    ordered_sets: bool
 
     OPTIONS_DOC = """
     **===  General Options  ===**
@@ -141,6 +142,8 @@ class LarkOptions(Serialize):
             Dictionary of callbacks for the lexer. May alter tokens during lexing. Use with caution.
     use_bytes
             Accept an input of type ``bytes`` instead of ``str``.
+    ordered_sets
+            Should Earley use ordered-sets to achieve stable output (~10%% slower than regular sets. Default: True)
     edit_terminals
             A callback for editing the terminals before parse.
     import_paths
@@ -179,6 +182,7 @@ class LarkOptions(Serialize):
         'edit_terminals': None,
         'g_regex_flags': 0,
         'use_bytes': False,
+        'ordered_sets': True,
         'import_paths': [],
         'source_path': None,
         '_plugins': {},
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -62,7 +62,7 @@ class LarkOptions(Serialize):
     regex: bool
     g_regex_flags: int
     keep_all_tokens: bool
-    tree_class: Optional[Callable[[str, Sequence], Any]]
+    tree_class: Optional[Callable[[str, List], Any]]
     parser: _ParserArgType
     lexer: _LexerArgType
     ambiguity: 'Literal["auto", "resolve", "explicit", "forest"]'
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -62,7 +62,7 @@ class LarkOptions(Serialize):
     regex: bool
     g_regex_flags: int
     keep_all_tokens: bool
-    tree_class: Callable
+    tree_class: Optional[Callable[[str, Sequence], Any]]
     parser: _ParserArgType
     lexer: _LexerArgType
     ambiguity: 'Literal["auto", "resolve", "explicit", "forest"]'
@@ -70,12 +70,12 @@ class LarkOptions(Serialize):
     priority: 'Optional[Literal["auto", "normal", "invert"]]'
     lexer_callbacks: Dict[str, Callable[[Token], Token]]
     use_bytes: bool
+    ordered_sets: bool
     edit_terminals: Optional[Callable[[TerminalDef], TerminalDef]]
     import_paths: 'List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]'
     source_path: Optional[str]
-    ordered_sets: bool
 
-    OPTIONS_DOC = """
+    OPTIONS_DOC = r"""
     **===  General Options  ===**
 
     start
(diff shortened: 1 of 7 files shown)


MegaIng (approved):

erezsh:
@MegaIng @chanicpanic This change makes Earley approx 10% slower. (on a very simple benchmark)

Do you think the stability it provides is worth it?  (let's assume it gives us 100% output stability)

MegaIng:
Maybe have an option that can change it, but default to the slower, stable version?

Also, I didn't check that this actually hits all locations, I assume you just searched for references to `set`.

chanicpanic:
I like the idea of having an option to enable/disable it. 

Defaulting to the stable version would be more user-friendly. The slowdown is unfortunate, but it is consistent with Lark positioning Earley as the parser for power and LALR as the parser for speed.

On the other hand, in addition to performance, it could be argued that defaulting to the unstable version has the benefit of exposing undesirable ambiguity and forcing grammar authors to remove/resolve it.

lark/parsers/earley_forest.py
```diff
@@ -280,10 +280,10 @@ def visit(self, root):
         # of a symbol/intermediate so that we can process both up and down. Also,
         # since the SPPF can have cycles it allows us to detect if we're trying
         # to recurse into a node that's already on the stack (infinite recursion).
-        visiting = set()
+        visiting = OrderedSet()
```
chanicpanic, line 283: `visiting` and `visited` probably don't need to be ordered because they are never iterated over.

erezsh:
@MegaIng @chanicpanic take 2

lark/parsers/earley.py
```diff
@@ -9,34 +9,36 @@
...
     from ..common import LexerConf, ParserConf
 
 class Parser:
     lexer_conf: 'LexerConf'
     parser_conf: 'ParserConf'
     debug: bool
 
-    def __init__(self, lexer_conf: 'LexerConf', parser_conf: 'ParserConf', term_matcher, resolve_ambiguity=True, debug=False, tree_class=Tree):
+    def __init__(self, lexer_conf: 'LexerConf', parser_conf: 'ParserConf', term_matcher: Callable,
+                 resolve_ambiguity: bool=True, debug: bool=False, tree_class: type=Tree, ordered_sets: bool=True):
```
chanicpanic, line 33: Note: In `LarkOptions` you defined `tree_class` as `Callable`. 
A more specific type might be `Type[Tree]` if we require that tree classes be subclasses of `Tree`. Also, `tree_class` here is `Optional[]`.

lark/lark.py
```diff
@@ -73,6 +73,7 @@ class LarkOptions(Serialize):
     edit_terminals: Optional[Callable[[TerminalDef], TerminalDef]]
     import_paths: 'List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]'
     source_path: Optional[str]
+    ordered_sets: bool
```
chanicpanic, line 76: If we want to be consistent with the order of the options, `ordered_sets` could be placed right after `use_bytes`.

lark/parsers/xearley.py
```diff
@@ -14,19 +14,25 @@
...
 
+if TYPE_CHECKING:
+    from ..common import LexerConf, ParserConf
 
 class Parser(BaseParser):
-    def __init__(self, lexer_conf, parser_conf, term_matcher, resolve_ambiguity=True, complete_lex = False, debug=False, tree_class=Tree):
-        BaseParser.__init__(self, lexer_conf, parser_conf, term_matcher, resolve_ambiguity, debug, tree_class)
+    def __init__(self, lexer_conf: 'LexerConf', parser_conf: 'ParserConf', term_matcher: Callable,
+                 resolve_ambiguity: bool=True, complete_lex: bool=False, debug: bool=False,
+                 tree_class: type=Tree, ordered_sets: bool=True):
```
chanicpanic, line 33: `tree_class` is Optional.

lark/parsers/earley.py
MegaIng: Suggestion: Don't require it to be a subclass of `Tree`.

And I don't think `tree_class` is `Optional`? It can't be `None`, that would cause an error, no?

lark/parsers/xearley.py
MegaIng: See above, I don't think this is the case.

erezsh: I agree with MegaIng on this one. tree_class can't be None (except in `LarkOptions`, and even that might be a bit of a relic).

lark/parsers/earley.py
erezsh: Isn't `tree_class` defined as Any in `LarkOptions`?

It could be Callable, I agree with that. Demanding that it's an instance of Tree might be a bit too limiting. Ideally, imho, we would use Protocol, once we drop Python 3.7.

Anyway, I think Callable is good enough for now, since this PR doesn't pretend to fix *all* the types.

lark/parser_frontends.py
```diff
@@ -1,7 +1,7 @@
 from typing import Any, Callable, Dict, Optional, Collection
 
 from .exceptions import ConfigurationError, GrammarError, assert_config
-from .utils import get_regexp_width, Serialize
+from .utils import get_regexp_width, Serialize, OrderedSet
```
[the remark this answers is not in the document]
erezsh: Fixed.

lark/lark.py
```diff
@@ -141,6 +142,8 @@ class LarkOptions(Serialize):
             Dictionary of callbacks for the lexer. May alter tokens during lexing. Use with caution.
     use_bytes
             Accept an input of type ``bytes`` instead of ``str``.
+    ordered_sets
+            Should Earley use ordered-sets to achieve stable output (~10%% slower than regular sets. Default: True)
```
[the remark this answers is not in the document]
erezsh: Fixed.

lark/parsers/earley.py
erezsh: Changed

chanicpanic: > And I don't think `tree_class` is `Optional`? It can't be `None`, that would cause an error, no?

`tree_class` is `None` for `ambiguity == 'forest'`:

https://github.com/lark-parser/lark/blob/c2e44ea7f823b76695b58304056a4178fadbc873/lark/parser_frontends.py#L200

https://github.com/lark-parser/lark/blob/c2e44ea7f823b76695b58304056a4178fadbc873/lark/parsers/earley.py#L299-L305

mypy may not be reporting an error because `tree_class` gets passed to the `Parser` constructor via dictionary unpacking. i.e.

https://github.com/lark-parser/lark/blob/c2e44ea7f823b76695b58304056a4178fadbc873/lark/parser_frontends.py#L194-L195

> We can turn it into `Callable[[str, Sequence], Any]` for now. Does that make sense?

That works

erezsh: When you're right, you're right!

lark/parsers/earley.py
```diff
@@ -30,7 +30,8 @@ class Parser:
     debug: bool
 
     def __init__(self, lexer_conf: 'LexerConf', parser_conf: 'ParserConf', term_matcher: Callable,
-                 resolve_ambiguity: bool=True, debug: bool=False, tree_class: type=Tree, ordered_sets: bool=True):
+                 resolve_ambiguity: bool=True, debug: bool=False,
+                 tree_class: Optional[type]=Tree, ordered_sets: bool=True):
```
chanicpanic, line 34: Did you mean to make it `Optional[Callable[[str, Sequence], Any]]`?
Same question for `xearley.py`

chanicpanic (approved):

PR #1334, opened 2023-09-05
Fixes invalid document link

cocolato:
Fixes invalid document link for https://github.com/lark-parser/lark/issues/1335.

### DIFF
--- a/examples/composition/README.rst
+++ b/examples/composition/README.rst
@@ -7,4 +7,6 @@ file format that allows both CSV and JSON to co-exist.
 We show how, by using namespaces, Lark grammars and their transformers can be fully reused -
 they don't need to care if their grammar is used directly, or being imported, or who is doing the importing.
 
-See `main.py <main.py>`_ for more details.
+See `main.py`_ for more details.
+
+.. _main.py: https://github.com/lark-parser/lark/blob/master/examples/composition/main.py
--- a/examples/grammars/README.rst
+++ b/examples/grammars/README.rst
@@ -3,4 +3,6 @@ Example Grammars
 
 This directory is a collection of lark grammars, taken from real world projects.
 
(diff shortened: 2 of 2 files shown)


erezsh:
It's written the way it is so that it works through the git repo too. Example: https://github.com/lark-parser/lark/blob/master/examples/grammars/README.rst

The right fix is to make sure sphinx generates the correct link.

cocolato:
The current implementation is a bit cumbersome, perhaps there is a better alternative.  @erezsh  Do you have any suggestions?

erezsh:
Well, maybe https://www.mkdocs.org/ with https://github.com/mkdocstrings/mkdocstrings can do the trick.

erezsh:
@cocolato I will try to find the time and look into it. I agree it shouldn't be this complicated.

erezsh:
Actually, maybe the easiest solution would be to use absolute links to the repository, i.e. https://github.com/lark-parser/lark/blob/master/examples/composition/main.py and https://github.com/lark-parser/lark/blob/master/examples/grammars/verilog.lark

What do you think?

MegaIng:
That wouldn't play well when exploring versions of the repo where those locations or files have changed and has no chance of working locally. But those are both slight edge cases.

cocolato:
Agree, at least the scope of influence will be relatively small.

erezsh:
@cocolato Use the local links please. I wrote them down in my last message.

erezsh:
@MegaIng I agree it's not ideal, but I don't have the time to fix sphinx. Better to have global links than broken links.

cocolato:
> @cocolato Use the local links please. I wrote them down in my last message.
The link has been updated.

erezsh:
Merged.

Issue #1329, opened 2023-08-25
Req: add "basiclexer=" option to Lark, to instruct ContextualLexer to use a custom BasicLexer


davidmcnabnz:
**Suggestion**
In `Lark.open()` (and Lark constructor), add an option such as `basiclexer=...` to allow users to specify an alternate `BasicLexer` class (which must be a subclass of `BasicLexer`) if they want to use the standard `ContextualLexer` with their own customised underlying lexer.

**Describe alternatives you've considered**
At present, there are only 2 ways to do this, both of which are problematic:

1. Create a subclass of `ContextualLexer`, replacing its constructor with one which is identical, save for changing the hardwiring from using BasicLexer, to using one's preferred BasicLexer subclass instead, or
2.  "monkey-patch" the `lexer` module to overwrite `BasicLexer` with one's preferred `BasicLexer` subclass, which will fool `ContextualLexer`'s constructor into using the replaced `BasicLexer` subclass.

Both these approaches are very messy and subject to breakage as Lark evolves. It would be better to have an orderly way to accomplish this.

**Additional context**
My legacy syntax parser project has hit some walls, due to the syntax allowing certain character sequences to delimit areas inside the code where the semantics change radically. Not only that, but these areas can overlap, meaning you can go from "not in an altered area", to "inside area A", then "inside area A _and_ area B", to "inside area B", to "not in an area".

I did consider switching to a custom lexer, based on one of the advanced examples, but I don't want to give up the power of Lark's contextual lexing capability. Also, this would rabbit-hole me into a maze of edge cases which don't allow robust solutions without deep introspection into parser states.

To me, the ability to maintain some global state flags, and control how each context-specific lexer behaves according to these flags, is a dealbreaker necessity for this project to continue.

If there's a way to accomplish this without requiring an added `basiclexer=...` option, or either of the 2 above 'hack' solutions, I'd be grateful to know.

erezsh:
What if we just refactor ContextualLexer to use self.basic_lexer, as official API, and you could inherit and override it easily?

For convenience, we can even provide a utility factory function, so calling Lark will look like `Lark(..., lexer=custom_contextual_lexer(my_basic_lexer))`

davidmcnabnz:
Thanks for your understanding, quick response and thoughts. The pattern you've suggested would be great, passing my subclass into that factory function.

Also seems like an approach that will age well for the Lark codebase.

Once that's in main, I'll transition away from my temporary monkey patch. 😊

[...]

PR #1342, opened 2023-09-23
ContextualLexer now uses self.basic_lexer for easy extensibility (iss…

erezsh:
…ue #1329)

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -626,7 +626,7 @@ class ContextualLexer(Lexer):
     lexers: Dict[str, AbstractBasicLexer]
     root_lexer: AbstractBasicLexer
 
-    basic_lexer: Type[AbstractBasicLexer] = BasicLexer
+    BasicLexer: Type[AbstractBasicLexer] = BasicLexer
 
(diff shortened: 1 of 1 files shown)


erezsh:
@davidmcnabnz Can you confirm this PR solves your issue?

lark/lexer.py
```diff
@@ -611,8 +622,10 @@ def next_token(self, lex_state: LexerState, parser_state: Any = None) -> Token:
 
 class ContextualLexer(Lexer):
 
-    lexers: Dict[str, BasicLexer]
-    root_lexer: BasicLexer
+    lexers: Dict[str, AbstractBasicLexer]
+    root_lexer: AbstractBasicLexer
+
+    basic_lexer: Type[AbstractBasicLexer] = BasicLexer
```
MegaIng, line 628: Why not `BasicLexer` as the name, following the pattern you using in #1327 ?

erezsh: Good point.

davidmcnabnz:
Thanks for that @erezsh . Commits look good, and plug up the configurability gap well. I look forward to migrating my lexer/parser over to the new version.

PR #1346, opened 2023-09-28
Fix #1345 by not calling the callbacks from InteractiveParser.accepts

### LINKED ISSUE #1345: UnexpectedToken.__str__ fails with Lark(transformer=...) when transformer throws exception
I use lark.Lark with the transformer= argument to run a transformer during parsing. I fould that this sometimes produced the following type of error message when an unexpected token was encountered: `lark.exceptions.UnexpectedToken: <exception str() failed>`. This seems to mean that an exception is raised in `UnexpectedToken.__str__`. When I ran the parser without the transformer, this did not happen.

**Minimal example to reproduce:**

```python
import lark

grammar = """
start: node_lhs OP "(" AMP_REF ")" node_rhs

node_lhs: (PERCENT_REF ("," PERCENT_REF)* "=")?
node_rhs: (PERCENT_REF ("," PERCENT_REF)*)?

OP: "add" | "sub"

AMP_REF: "&" DIGIT+
PERCENT_REF: "%" DIGIT+

%import common.DIGIT
%import common.WS_INLINE
%ignore WS_INLINE
"""

class MyTransformer(lark.visitors.Transformer):
    def OP(self, token):
        return {"add":1, "sub":2}[str(token)]

def case1():
    parser = lark.Lark(grammar, parser="lalr",  lexer="basic")

    # Success:
    print(MyTransformer().transform(parser.parse("%17 = add(&1) %13, %16")))

    # Raises graceful lark.exceptions.UnexpectedToken:
    MyTransformer().transform(parser.parse("%17 = add(&1) %13, %16"))
    parser.parse("&17 = add %13, %16")

def case2():
    parser = lark.Lark(grammar, parser="lalr",  lexer="basic", transformer=MyTransformer())

    # Success:
    print(parser.parse("%17 = add(&1) %13, %16"))

    # Raises broken lark.exceptions.UnexpectedToken: <exception str() failed>
    parser.parse("&17 = add %13, %16")

#case1()
case2()
```

`case1()` runs fine, the problem is in `case2()`.

**Analysis:** `UnexpectedToken.__str__` calls `InteractiveParser.accepts` through the property `UnexpectedToken.accepts`. `InteractiveParser.accepts` somehow calls `MyTransformer.OP` with an empty token as argument. `MyTransformer.OP` raises a KeyError when it gets that empty ('') token.

**Fix:** I find that it is acceptable behaviour for the transformer rule to raise an exception when it gets a token that violating the grammar. I guess it should even be acceptable for the transformer to raise an exception in other cases. Possible ways to fix this problem could be:

- Do not attach InteractiveParsers to UnexpectedInput exceptions (`_Parser.parse_from_state`, lalr_parser.py:190) when a transformer is used directly together with a parser. Then, `UnexpectedToken.__str__` would produce a error message using `self.expected` instead of `self.accepts`. I do not know what the drawbacks of using `self.expected` are.
- Catch + handle any type of exception in `UnexpectedToken.accepts`. On error, resort to self.expected or return a meaningful string to prevent `<exception str() failed>`.

Versions: lark-1.1.7, Python 3.11.2

MegaIng:
This also has the nice benefit of making `accepts` faster since it does less work. Technically this is an option that an end user can access on their instance of `InteractiveParser`, so I added a bit of documentation.

### DIFF
--- a/lark/parsers/lalr_interactive_parser.py
+++ b/lark/parsers/lalr_interactive_parser.py
@@ -24,12 +24,16 @@ class InteractiveParser:
         warnings.warn("lexer_state will be removed in subsequent releases. Use lexer_thread instead.", DeprecationWarning)
         return self.lexer_thread
 
-    def feed_token(self, token: Token):
+    def feed_token(self, token: Token, no_callbacks=False):
         """Feed the parser with a token, and advance it to the next state, as if it received it from the lexer.
 
         Note that ``token`` has to be an instance of ``Token``.
+
+        When ``no_callbacks`` is used, no callbacks are applied to the values in the stack, which means
+        that no Tree instances are created. This leads to potentially weird results. Should primarily be used
+        if you don't want to use the parsing result and only care if it matches at all.
         """
-        return self.parser_state.feed_token(token, token.type == '$END')
+        return self.parser_state.feed_token(token, token.type == '$END', no_callbacks=no_callbacks)
 
     def iter_parse(self) -> Iterator[Token]:
         """Step through the different stages of the parse, by reading tokens from the lexer
@@ -106,7 +110,7 @@ class InteractiveParser:
             if t.isupper(): # is terminal?
                 new_cursor = copy(self)
                 try:
-                    new_cursor.feed_token(self.lexer_thread._Token(t, ''))
+                    new_cursor.feed_token(self.lexer_thread._Token(t, ''), no_callbacks=True)
                 except UnexpectedToken:
                     pass
                 else:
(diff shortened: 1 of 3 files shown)


erezsh:
I wonder if instead of having to add a new parameter on all these methods, we can instead clear the callbacks on the copy method which `accepts()` uses. Like `new_cursor = self.copy(callbacks = {})`.

Then all we need is to change in the parser is `value = callbacks[rule](s) if callbacks else s`

It's a bit more implicit, but it feels cleaner to me. What do you think?

MegaIng:
We are currently avoiding creating a copy of the ParserConf object since it's semantically immutable. So we have to either:

- create a deepcopy in `accepts` (very expensive)
- always copying `ParserConf` in the normal `__copy__` (unnessary, although it wouldn't be *that* expensive)
- create an extra set of methods on `InteractiveParser`, `ParserState` (and maybe `ParserConf`) just for this usecase that specifcally doesn't copy callbacks 
- specifically make a copy of `ParserConf` after doing a normal copy of `InteractiveParser`. (Not sure if I like this)
  - Although, I guess we could create a single copy of `ParserConf` in `accepts` and use that for all Tokens...

erezsh:
An alternative implementation has been merged.

PR #1347, opened 2023-09-28
Fix 1345 attempt two

MegaIng:
Alternative to #1346

### DIFF
--- a/lark/parsers/lalr_interactive_parser.py
+++ b/lark/parsers/lalr_interactive_parser.py
@@ -102,9 +102,14 @@ class InteractiveParser:
     def accepts(self):
         """Returns the set of possible tokens that will advance the parser into a new valid state."""
         accepts = set()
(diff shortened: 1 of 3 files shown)


erezsh:
I like this one better!

erezsh:
I tend to prefer this version because -

- Using the no_callbacks option can have "potentially weird results", as you mention in the docstring. But once we add that API, it might be hard to remove it.
- It adds complication to a function that is already fairly complicated. Without understanding the context, it's not clear why it exists.
- Right now overriding the parser_conf is a bit awkward, but if we want to expose it to the end-user, I believe we can find a way to add an API for it. Maybe even provide a way to override the callbacks with arbitrary dictionaries, which will enable the more fearless users to do all kinds of hackable shenanigans.

Anyway, thanks for making both versions! I'll sleep on it, and if there's no new information or arguments, I'll probably merge this one.

Issue #1295, opened 2023-07-03
lark.exceptions.VisitError: Error trying to process rule "expansion":


zdebel:
I've got an exception like in the title, I can't find where I went wrong, attached is the grammar file, I'm running lark using a simple parse call:

`tree = verilog_parser.parse(open("test_data/verilog/inverter.v").read())`

```
//used https://www.verilog.com/VerilogBNF.html as reference, this is Verilog 1995

//1. Source Text

//<source_text>
start: description*

//<description>
description: module

//<module>
module: "module" name_of_module list_of_ports? ";" module_item* "endmodule"

//<name of module>
name_of_module: IDENTIFIER

//<list_of_ports>
list_of_ports: "(" port ("," port)* ")"

//<port> simplified for the moment, not according to ref
port: IDENTIFIER

//<module_item>
module_item: input_declaration
    | output_declaration
    | inout_declaration
    | continuous_assign

//2. Declarations

//<input_declaration>
input_declaration: "input" range? list_of_variables ";"

//<output_declaration>
output_declaration: "output" range? list_of_variables ";"

//<inout_declaration>
inout_declaration: "inout" range? list_of_variables ";"

//<list_of_variables>
list_of_variables: NETTYPE drive_strength? expandrange? delay? list_of_assignments ";"

//<NETTYPE>
NETTYPE: "wire"
    | "tri"
    | "tri1"
    | "supply0"
    | "wand"
    | "triand"
    | "tri0"
    | "supply1"
    | "wor"
    | "trior"
    | "trireg"

//<expandrange>
expandrange: range
    | "scalared" range
    | "vectored" range

//<continuous_assign>
continuous_assign: "assign" drive_strength? delay? list_of_assignments ";"

//<drive_strength>
drive_strength: "(" STRENGTH0 "," STRENGTH1 ")"
    | "(" STRENGTH1 "," STRENGTH0 ")"

//<STRENGTH>
STRENGTH0: "supply0"
    | "strong0"
    | "pull0"
    | "weak0"
    | "highz0"
    |
//<STRENGTH1>
STRENGTH1: "supply1"
    | "strong1"
    | "pull1"
    | "weak1"
    | "highz1"

//<range>
range: "[" constant_expression ":" constant_expression "]"

//<list_of_assignments>
list_of_assignments: assignment ("," assignment)*

//5. Behavioral statements

//<assignment>
assignment: lvalue "=" expression

//7. Expressions

//<lvalue>
lvalue: identifier

[...]

PR #1354, opened 2023-10-20
Typing fixes. Mypy now produces 0 type errors

erezsh:
Also adding some typing info


MegaIng:
This also fixes #1295 

I am surprised mypy doesn't complain about partially typed functions. This looks fine otherwise.

erezsh:
Oh, yeah. I forgot to mention it. Thanks!

Issue #1362, opened 2023-11-12
Black formatter breaks Lark standalone parser generation


marsninja:
**Describe the bug**

The functionality of Lark standalone parser generator breaks if lark codebase is run through the black formatter tool. 

This error is produced:

```
jaclang/vendor/lark/tools/standalone.py:217: in main
    gen_standalone(lark_inst, out=out, compress=ns.compress)
jaclang/vendor/lark/tools/standalone.py:170: in gen_standalone
    code = extract_sections(f)["standalone"]
E   KeyError: 'standalone'
```

The culpret comes from the extract_sections function on line 96 (post black formatting):

```python
def extract_sections(lines):
    section = None
    text = []
    sections = defaultdict(list)
    for line in lines:
        if line.startswith("###"):
            if line[3] == "{":
                section = line[4:].strip()
            elif line[3] == "}":
                sections[section] += text
                section = None
                text = []
            else:
                raise ValueError(line)
        elif section:
            text.append(line)
```

A fixed version is 

```python
def extract_sections(lines):
    section = None
    text = []
    sections = defaultdict(list)
    for line in lines:
        clean_line = line.strip()
        if clean_line.startswith("###"):
            if clean_line[3] == "{":
                section = clean_line[4:].strip()
            elif clean_line[3] == "}":
                sections[section] += text
                section = None
                text = []
            else:
                raise ValueError(line)
        elif section:
            text.append(line)
```

Will make a pr. 

**To Reproduce**

Run lark code base through black formatter, then try to create a standalone parser.

PR #1363, opened 2023-11-12
A fix for black formatter breaking standalone parser generation.

marsninja:
Here lies a fix for Lark's standalone parser generation breaking when Lark is run through the black formatter. The fix simply entails stripping lines as they are iterated through in the `extract_sections` function in standlone.py.

### DIFF
--- a/lark/tools/standalone.py
+++ b/lark/tools/standalone.py
@@ -77,10 +77,11 @@ def extract_sections(lines):
     text = []
     sections = defaultdict(list)
     for line in lines:
-        if line.startswith('###'):
-            if line[3] == '{':
-                section = line[4:].strip()
-            elif line[3] == '}':
+        clean_line = line.strip()
+        if clean_line.startswith("###"):
+            if clean_line[3] == "{":
+                section = clean_line[4:].strip()
+            elif clean_line[3] == "}":
                 sections[section] += text
                 section = None
                 text = []


marsninja:
Fixes #1362

MegaIng:
Or... You don't run `black` on this codebase, which in this case destroys information. I presume the problem is that black indents the comments, which is just wrong behavior since the comments do act on a global scope.

marsninja:
@MegaIng Thats correct, the comments get tabbed in with black, however the `strip` fix appears to correct the issue for both Lark proper and standalone.  Doesn't hurt to make the implementation more black friendly.

erezsh:
@marsninja Why not just run black after the standalone has been generated?

MegaIng:
> @MegaIng Thats correct, the comments get tabbed in with black, however the `strip` fix appears to correct the issue for both Lark proper and standalone. Doesn't hurt to make the implementation more black friendly.

Sure, it technically doesn't hurt. But it is extra code, and it allows something that currently isn't allowed, with good reasons. Having those comments be indented is confusing for human parsers, so it shouldn't be allowed. Considering the alternative solution to this "issue" is "just don't do the thing that breaks it, there is no reason to do it", I don't see why we should add this.

erezsh:
@marsninja Please address the concerns of the maintainers, or I will close this PR. Thanks.

m42e:
Here is something

Issue #1384, opened 2024-01-25
Ability to inline branches not documented in the grammar reference


kpet:
The JSON tutorial states:

>The question-mark prefixing value (”?value”) tells the tree-builder to inline this branch if it has only one member. In this case, value will always have only one member, and will always be inlined.

Unless I've missed something, this feature is not documented in the grammar reference.

erezsh:
No, it's mentioned here: https://lark-parser.readthedocs.io/en/latest/tree_construction.html

But you might be right, that the grammar reference should mention it too.

kpet:
Indeed it is, I'd missed that. I remembered reading about the feature in the tutorial but when I found a need for it in my grammar and wanted to look at the reference, I naturally found myself going to the grammar reference. Either way, I found the info in the tutorial in the end.

(BTW, I didn't open with this but: great project, thank you very much, it's made something really easy for me :))

nchammas:
I submitted #1381 for a similar reason, though that was just to make the features stand out more in the Tree Construction Reference, rather than add them to the Grammar Reference.

Issue #1382, opened 2024-01-08
AssertionError when using templates


subnut:
**Describe the bug**

AssertionError
```
$ python parser.py
Traceback (most recent call last):
  File "/home/user/Projects/project/parser.py", line 5, in <module>
    parser = Lark.open("grammar.lark")
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.config/nvim/venv/lib/python3.12/site-packages/lark/lark.py", line 580, in open
    return cls(f, **options)
           ^^^^^^^^^^^^^^^^^
  File "/home/user/.config/nvim/venv/lib/python3.12/site-packages/lark/lark.py", line 410, in __init__
    self.terminals, self.rules, self.ignore_tokens = self.grammar.compile(self.options.start, terminals_to_keep)
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.config/nvim/venv/lib/python3.12/site-packages/lark/load_grammar.py", line 710, in compile
    terminals = [TerminalDef(name, transformer.transform(term_tree), priority)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.config/nvim/venv/lib/python3.12/site-packages/lark/lexer.py", line 126, in __init__
    assert isinstance(pattern, Pattern), pattern
AssertionError: Tree('template_usage', [NonTerminal('_string'), "'"])
```

**To Reproduce**

This grammar fails with `AssertionError`
```python3
start: _strconst+
_strconst: STRCONST_SQ
         | STRCONST_BQ
         | STRCONST_DQ
STRCONST_SQ: _string{"'"}
STRCONST_BQ: _string{"`"}
STRCONST_DQ: _string{"\""}
_string{quot}: quot /.*?/s /(?<!\\)(\\\\)*?/ quot
```

whereas the same grammar, when expanded, works perfectly fine!
```python3
start: _strconst+
_strconst: STRCONST_SQ
         | STRCONST_BQ
         | STRCONST_DQ
STRCONST_SQ: "'"   /.*?/s /(?<!\\)(\\\\)*?/   "'"
STRCONST_BQ: "`"   /.*?/s /(?<!\\)(\\\\)*?/   "`"
[...]

Issue #1355, opened 2023-10-22
GrammarError: Rule 'anycase' used but not defined (in rule pipesyn)


RossPatterson:
When attempting to load a small grammar, Lark reports that a rule alias name is "used but not defined".  It's rather obviously defined, as the only usage of the name is the alias.  The specific alias that gets flagged varies - running the example below flags `anycase`, `anyof`, and `mask`, seemingly randomly.

The grammar is intended to recognize simple operand strings like:

```
AnYcA mIXEd

ONEs

zero ANYOF
```

To reproduce, run the following:

```python
import lark

grammar = '''
start: pipesyn
pipesyn: ["any"i ("c"i ("a"i ("s"i ("e"i)? )? )? )? -> anycase]
 | [("mixed"i
 | "one"i ("s"i)?
 | "zero"i ("s"i)?) -> mask]
 | ["anyo"i ("f"i)? -> anyof]
'''

parser = lark.Lark(grammar, parser="lalr")
```

You'll get an exception like the following:
```
(.venv) C:\Ross\Source\Pype>py src\pype\issue.py
Traceback (most recent call last):
  File "C:\Ross\Source\Pype\src\pype\issue.py", line 12, in <module>
    parser = lark.Lark(grammar, parser="lalr")
  File "C:\Ross\Source\Pype\.venv\lib\site-packages\lark\lark.py", line 356, in __init__
    self.grammar, used_files = load_grammar(grammar, self.source_path, self.options.import_paths, self.options.keep_all_tokens)
  File "C:\Ross\Source\Pype\.venv\lib\site-packages\lark\load_grammar.py", line 1411, in load_grammar
    return builder.build(), builder.used_files
  File "C:\Ross\Source\Pype\.venv\lib\site-packages\lark\load_grammar.py", line 1370, in build
    self.validate()
  File "C:\Ross\Source\Pype\.venv\lib\site-packages\lark\load_grammar.py", line 1364, in validate
    self._grammar_error(d.is_term, "{Type} '{name}' used but not defined (in {type2} {name2})", sym, name)
  File "C:\Ross\Source\Pype\.venv\lib\site-packages\lark\load_grammar.py", line 1104, in _grammar_error
    raise GrammarError(msg.format(**args))
lark.exceptions.GrammarError: Rule 'anycase' used but not defined (in rule pipesyn)
```

MegaIng:
[...]

Issue #555, opened 2020-04-06
Token templates


erezsh:
Rule templates are already here. Token templates are next?

@MegaIng Do you think this will be ready in the near future? I want to make a release in a few weeks and it would be nice if it made it in. I also want to refactor `load_grammar.py`, and I don't want to cause any collisions for you. But no pressure :)

PR #1388, opened 2024-02-01
Make lark.lark parse the same grammar as load_grammar.py, and make grammar.md document it more fully.

### LINKED ISSUE #391: Lark grammar?
Are there any plans to create a Lark grammar for Lark?  The only file currently in lark/grammars is common.lark, which defines a few terminals.

The authoritative definition of Lark would logically be expressed as a grammar (as opposed to "what the grammar reference says" or "what the latest version of the code parses").  Note that RFC 5234 defines the ABNF grammar that authoritatively validates ABNF specifications.

RossPatterson:
Make `lark.lark` parse the same grammar as `load_grammar.py`, and make `grammar.md` document it more fully.

Fixes #391, #1382, #1384.

- All scenarios have tests in new file `tests/test_lark_lark.py`.  Each test passed `load_grammar.py` and failed `lark.lark` before these changes, and passed both after.  All existing un-skipped unit tests continue to pass after these changes.  The skipped Earley tests have not been checked. 

- Updates to the formal Lark grammar (`lark/grammars/lark.lark`) to make it match the actual parser (`lark/load_grammar.py`)
 1. Token aliases (_e.g._, `NAME_1 -> NAME_2`) are no longer allowed.
    - `load_grammar.py` appears to accept them, but forces the new name to be a rule-name.
    - `tests/test_grammar.py.test_alias_in_terminal` says they should raise a `GrammarError`.
    - Making this change necessitated splitting the shared `expansions` rule, and all the rules it calls, into `rule_expansions` and `token_expansions` (and similarly named rules for them to call).
 1. Rule aliases (`name_1 -> name_2`) are now only valid for alternates at the top-level of a rule.
    - load_grammar.py accepts them at all levels, but as Issue #1355 points out, they are not supported below the top level.
 1. The item in a `%ignore` directive must now be a single token.
    - `load_grammar.py` accepts multiple tokens, but only processes the first.
  1. The `%extend` directive is now defined for rules.
  1. The `%extend` directive is now defined for tokens.
  1. Token templates are no longer allowed (at least until Issue #555 is resolved).
  1. Tokens can no longer include rules.
  1. The `%override` directive now supports overriding token definitions.
  1. Rule modifiers are no longer allowed in rule references (_e.g._, `rule1: !?rule2`)
  1. Rule modifiers are now allowed to be `?!`, in addition to the existing `!`, `?`, and `!?`.

* Several tests in `tests/tests/test_grammar.py` have been copied to new file ` tests/test_grammar_formal.py` and updated to parse using `lark.lark` instead of `load_grammar.py`.  Only the tests that verify how the Lark grammar is parsed could be implemented.  Those that verify how a Lark-generated parser operates could not, because there's no (easy?) way to create a parser from the parse tree.

* Oddities not changed:
  * Literal ranges are defined in both `lark.lark` and `load_grammar.py` as `STRING`s, but must actually be single characters.
    * `load_grammar.py` initially accepts multi-character strings, but later rejects them.
	* Attempting to replace `STRING ".." STRING` in `lark.lark` with `/./ ".." /./` failed, as Lark accepts Unicode escapes as strings (_e.g._ `test_unicode_literal_range_escape2` uses `"\U0000FFFF".."\U00010002"`).
	* Attempting to replace `STRING ".." STRING` in `lark.lark` with `/.|"\\U[0-9a-fA-F]+"/ ".." /.|"\\U[0-9a-fA-F]+"/` failed, causing lots of unrelated tests to fail.
  * `load_grammar.py` does not allow `STRING` tokens to be empty (_i.e._, `""`).  There is no (easy?) way to prevent that in lark.lark.  The obvious answer, adding `/(?<!"")/`, broke `tests/test_parser.py._TestParser.test_backslash2`.

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -59,26 +59,37 @@ Terminals are used to match text into symbols. They can be defined as a combinat
 **Syntax:**
 
 ```html
-<NAME> [. <priority>] : <literals-and-or-terminals>
+<NAME> [. <priority>] : <items-to-match>
 ```
 
-Terminal names must be uppercase.
+Terminal names must be uppercase.  They must start with an underscore (`_`) or a letter (`A` through `Z`), and may be composed of letters, underscores, and digits (`0` through `9`).  Terminal names that start with "_" will not be included in the parse tree, unless the `keep_all_tokens` option is specified.
 
 Literals can be one of:
 
-* `"string"`
-* `/regular expression+/`
-* `"case-insensitive string"i`
-* `/re with flags/imulx`
-* Literal range: `"a".."z"`, `"1".."9"`, etc.
+* Literal range: `"a".."z"`, `"1".."9"`, etc. - Each literal must be a single character, and the range represends all values between the two literals, inclusively.
 
-Terminals also support grammar operators, such as `|`, `+`, `*` and `?`.
+Each item is one of:
+
+* `TERMINAL` - Another terminal, which cannot be defined in terms of this terminal.
+* `"string literal"` - Literal, to be matched as-is.
+* `"string literal"i` - Literal, to be matched case-insensitively.
+* `/regexp literal/` - Regular expression literal.  Can inclde flags.
+* `"character".."character"` - Literal range.  The range represends all values between the two literals, inclusively.
+* `(item item ..)` - Group items
+* `(item | item | ..)` - Alternate items.
+* `[item item ..]` - Maybe. Same as `(item item ..)?`, but when `maybe_placeholders=True`, generates `None` if there is no match.
+* `[item | item | ..]` - Maybe with alternates. Same as `(item | item | ..)?`, but when `maybe_placeholders=True`, generates `None` if there is no match.
+* `item?` - Zero or one instances of item (a "maybe")
+* `item*` - Zero or more instances of item
+* `item+` - One or more instances of item
+* `item ~ n` - Exactly *n* instances of item
+* `item ~ n..m` - Between *n* to *m* instances of item (not recommended for wide ranges, due to performance issues)
 
 Terminals are a linear construct, and therefore may not contain themselves (recursion isn't allowed).
 
 ### Templates
 
-Templates are expanded when preprocessing the grammar.
+Templates are expanded when preprocessing rules in the grammar.  Templates are not allowed with terminals.
 
 Definition syntax:
 
@@ -122,7 +133,7 @@ SIGNED_INTEGER: /
  /x
 ```
 
-Supported flags are one of: `imslux`. See Python's regex documentation for more details on each one.
+Supported flags are one of: `imslux`. See Python's [regex documentation](https://docs.python.org/3/library/re.html#regular-expression-syntax) for more details on each one.
 
 Regexps/strings of different flags can only be concatenated in Python 3.6+
 
@@ -196,25 +207,32 @@ _ambig
 
 **Syntax:**
 ```html
-<name> : <items-to-match>  [-> <alias> ]
+<modifiers><name> : <items-to-match>  [-> <alias> ]
        | ...
 ```
 
-Names of rules and aliases are always in lowercase.
+Names of rules and aliases are always in lowercase.  They must start with an underscore (`_`) or a letter (`a` through `z`), and may be composed of letters, underscores, and digits (`0` through `9`).  Rule names that start with "_" will be inlined into their containing rule.
 
 Rule definitions can be extended to the next line by using the OR operator (signified by a pipe: `|` ).
 
-An alias is a name for the specific rule alternative. It affects tree construction.
+An alias is a name for the specific rule alternative. It affects tree construction (see [Shaping the tree](tree_construction#shaping_the_tree).
 
+The affect of a rule on the parse tree can be specified by modifiers.  The `!` modifier causes the rule to keep all its tokens, regardless of whether they are named or not.  The `?` modifier causes the rule to be inlined if it only has a single child.  The `?` modifier cannot be used on rules that are named starting with an underscore.
 
 Each item is one of:
 
 * `rule`
 * `TERMINAL`
-* `"string literal"` or `/regexp literal/`
+* `"string literal"` - Literal, to be matched as-is.
+* `"string literal"i` - Literal, to be matched case-insensitively.
+* `/regexp literal/` - Regular expression literal.  Can inclde flags.
+* `"character".."character"` - Literal range.  The range represends all values between the two literals, inclusively.
+* template(parameter1, parameter2, ..) - A template to be  expanded with the specified parameters.
 * `(item item ..)` - Group items
+* `(item | item | ..)` - Alternate items.  Note that the items cannot have aliases.
 * `[item item ..]` - Maybe. Same as `(item item ..)?`, but when `maybe_placeholders=True`, generates `None` if there is no match.
-* `item?` - Zero or one instances of item ("maybe")
+* `[item | item | ..]` - Maybe with alternates. Same as `(item | item | ..)?`, but when `maybe_placeholders=True`, generates `None` if there is no match.  Note that the items cannot have aliases.
+* `item?` - Zero or one instances of item (a "maybe")
 * `item*` - Zero or more instances of item
 * `item+` - One or more instances of item
 * `item ~ n` - Exactly *n* instances of item
@@ -297,11 +315,23 @@ Note that `%ignore` directives cannot be imported. Imported rules will abide by 
 
 Declare a terminal without defining it. Useful for plugins.
 
+**Syntax:**
+```html
+%declare <TERMINAL>
+%declare <rule>
+```
+
 ### %override
 
 Override a rule or terminals, affecting all references to it, even in imported grammars.
 
 Useful for implementing an inheritance pattern when importing grammars.
+
+**Syntax:**
+```html
+%override <TERMINAL> ... terminal definition ...
+%override <rule> ... rule definition ...
+```
 
 **Example:**
 ```perl
@@ -318,6 +348,12 @@ Extend the definition of a rule or terminal, e.g. add a new option on what it ca
 Useful for splitting up a definition of a complex rule with many different options over multiple files.
 
 Can also be used to implement a plugin system where a core grammar is extended by others.
+
+**Syntax:**
+```html
+%extend <TERMINAL> ... additional terminal alternate ...
+%extend <rule> ... additional rule alternate ...
+```
 
 
 **Example:**
--- a/docs/tree_construction.md
+++ b/docs/tree_construction.md
@@ -74,6 +74,7 @@ Lark will parse "((hello world))" as:
 The brackets do not appear in the tree by design. The words appear because they are matched by a named terminal.
 
 
+<a name="shaping_the_tree"></a>
 ## Shaping the tree
 
 Users can alter the automatic construction of the tree using a collection of grammar features.
--- a/lark/grammars/lark.lark
+++ b/lark/grammars/lark.lark
@@ -7,46 +7,66 @@ _item: rule
      | token
      | statement
 
-rule: RULE rule_params priority? ":" expansions
-token: TOKEN token_params priority? ":" expansions
+rule: RULE_MODIFIERS? RULE rule_params priority? ":" rule_expansions
+token: TOKEN priority? ":" token_expansions
 
 rule_params: ["{" RULE ("," RULE)* "}"]
-token_params: ["{" TOKEN ("," TOKEN)* "}"]
 
 priority: "." NUMBER
 
-statement: "%ignore" expansions                    -> ignore
+statement: "%ignore" ignore_token                  -> ignore
          | "%import" import_path ["->" name]       -> import
          | "%import" import_path name_list         -> multi_import
          | "%override" rule                        -> override_rule
+         | "%override" token                       -> override_token
          | "%declare" name+                        -> declare
+         | "%extend" rule                          -> extend_rule
+         | "%extend" token                         -> extend_token
+
+ignore_token: ignore_item [ OP | "~" NUMBER [".." NUMBER]]
+ignore_item: STRING | TOKEN | REGEXP
 
 !import_path: "."? name ("." name)*
 name_list: "(" name ("," name)* ")"
 
-?expansions: alias (_VBAR alias)*
+?rule_expansions: rule_alias (_VBAR rule_alias)*
 
-?alias: expansion ["->" RULE]
+?rule_inner_expansions: rule_expansion (_VBAR rule_expansion)*
 
-?expansion: expr*
+?rule_alias: rule_expansion ["->" RULE]
 
-?expr: atom [OP | "~" NUMBER [".." NUMBER]]
+?rule_expansion: rule_expr*
 
-?atom: "(" expansions ")"
-     | "[" expansions "]" -> maybe
-     | value
+?rule_expr: rule_atom [OP | "~" NUMBER [".." NUMBER]]
+?rule_atom: "(" rule_inner_expansions ")"
+          | "[" rule_inner_expansions "]" -> rule_maybe
+          | rule_value
 
-?value: STRING ".." STRING -> literal_range
-      | name
-      | (REGEXP | STRING) -> literal
-      | name "{" value ("," value)* "}" -> template_usage
+?rule_value: RULE "{" rule_value ("," rule_value)* "}" -> rule_template_usage
+           | RULE
+           | token_value
+
+?token_expansions: token_expansion (_VBAR token_expansion)*
+
+?token_expansion: token_expr*
+
+?token_expr: token_atom [OP | "~" NUMBER [".." NUMBER]]
+
+?token_atom: "(" token_expansions ")"
+           | "[" token_expansions "]" -> token_maybe
+           | token_value
+
+?token_value: STRING ".." STRING -> literal_range
+            | TOKEN
+            | (REGEXP | STRING) -> literal
 
 name: RULE
     | TOKEN
 
 _VBAR: _NL? "|"
 OP: /[+*]|[?](?![a-z])/
-RULE: /!?[_?]?[a-z][_a-z0-9]*/
+RULE: /_?[a-z][_a-z0-9]*/
+RULE_MODIFIERS: /!|![?](?=[a-z])|[?]!?(?=[a-z])/
 TOKEN: /_?[A-Z][_A-Z0-9]*/
 STRING: _STRING "i"?
 REGEXP: /\/(?!\/)(\\\/|\\\\|[^\/])*?\/[imslux]*/
--- a/tests/test_grammar_formal.py
+++ b/tests/test_grammar_formal.py
@@ -1,0 +1,169 @@
+from __future__ import absolute_import
+
+import os
+from unittest import TestCase, main
+
+from lark import lark, Lark, UnexpectedToken
+from lark.load_grammar import GrammarError
+
+
+# Based on TestGrammar, with lots of tests that can't be run elided.
+class TestGrammarFormal(TestCase):
+    def setUp(self):
+        lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark')
+        # lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark-ORIG')
+        with open(lark_path, 'r') as f:
+            self.lark_grammar = "\n".join(f.readlines())
+
+    def test_errors(self):
+        # raise NotImplementedError("Doesn't work yet.")
+        l = Lark(self.lark_grammar, parser="lalr")
+
+        # This is an unrolled form of the test_grammar.py:GRAMMAR_ERRORS tests, because the lark.lark messages vary.
+
+        # 'Incorrect type of value', 'a: 1\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..NUMBER., .1..', l.parse, 'a: 1\n')
+        # 'Unclosed parenthesis', 'a: (\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token.._NL.,', l.parse, 'a: (\n')
+        # 'Unmatched closing parenthesis', 'a: )\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..RPAR.', l.parse, 'a: )\n')
+        # 'Unmatched closing parenthesis', 'a: )\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..RPAR.,', l.parse, 'a: )\n')
+        # 'Unmatched closing parenthesis', 'a: (\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token.._NL.,', l.parse, 'a: (\n')
+        # 'Expecting rule or terminal definition (missing colon)', 'a\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token.._NL.,', l.parse, 'a\n')
+        # 'Expecting rule or terminal definition (missing colon)', 'A\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token.._NL.,', l.parse, 'A\n')
+        # 'Expecting rule or terminal definition (missing colon)', 'a->\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..__ANON_0., .->', l.parse, 'a->\n')
+        # 'Expecting rule or terminal definition (missing colon)', 'A->\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..__ANON_0., .->', l.parse, 'A->\n')
+        # 'Expecting rule or terminal definition (missing colon)', 'a A\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..TOKEN., .A..', l.parse, 'a A\n')
+        # 'Illegal name for rules or terminals', 'Aa:\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..RULE., .a..', l.parse, 'Aa:\n')
+        # 'Alias expects lowercase name', 'a: -> "a"\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..STRING., ."a"..', l.parse, 'a: -> "a"\n')
+        # 'Unexpected colon', 'a::\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..COLON.,', l.parse, 'a::\n')
+        # 'Unexpected colon', 'a: b:\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..COLON.,', l.parse, 'a: b:\n')
+        # 'Unexpected colon', 'a: B:\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..COLON.,', l.parse, 'a: B:\n')
+        # 'Unexpected colon', 'a: "a":\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..COLON.,', l.parse, 'a: "a":\n')
+        # 'Misplaced operator', 'a: b??'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..OP., .\?..', l.parse, 'a: b??')
+        # 'Misplaced operator', 'a: b(?)'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..OP., .\?..', l.parse, 'a: b(?)')
+        # 'Misplaced operator', 'a:+\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..OP., .\+..', l.parse, 'a:+\n')
+        # 'Misplaced operator', 'a:?\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..OP., .\?..', l.parse, 'a:?\n')
+        # 'Misplaced operator', 'a:*\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..OP., .\*..', l.parse, 'a:*\n')
+        # 'Misplaced operator', 'a:|*\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..OP., .\*..', l.parse, 'a:|*\n')
+        # 'Expecting option ("|") or a new rule or terminal definition', 'a:a\n()\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..LPAR.,', l.parse, 'a:a\n()\n')
+        # 'Terminal names cannot contain dots', 'A.B\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..TOKEN., .B..', l.parse, 'A.B\n')
+        # 'Expecting rule or terminal definition', '"a"\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..STRING., ."a"..', l.parse, '"a"\n')
+        # '%import expects a name', '%import "a"\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..STRING., ."a"..', l.parse, '%import "a"\n')
+        # '%ignore expects a value', '%ignore %import\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..__ANON_2., .%import..', l.parse, '%ignore %import\n')
+
+    # def test_empty_literal(self):
+        # raise NotImplementedError("Breaks tests/test_parser.py:_TestParser:test_backslash2().")
+
+    # def test_ignore_name(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_override_rule_1(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_override_rule_2(self):
+        # raise NotImplementedError("Can't test semantics of grammar, only syntax.")
+
+    # def test_override_rule_3(self):
+        # raise NotImplementedError("Can't test semantics of grammar, only syntax.")
+
+    # def test_override_terminal(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_extend_rule_1(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_extend_rule_2(self):
+        # raise NotImplementedError("Can't test semantics of grammar, only syntax.")
+
+    # def test_extend_term(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_extend_twice(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_undefined_ignore(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    def test_alias_in_terminal(self):
+        l = Lark(self.lark_grammar, parser="lalr")
+        g = """start: TERM
+            TERM: "a" -> alias
+            """
+        # self.assertRaisesRegex( GrammarError, "Aliasing not allowed in terminals", Lark, g)
+        self.assertRaisesRegex( UnexpectedToken, "Unexpected token Token.'__ANON_0', '->'.", l.parse, g)
+
+    # def test_undefined_rule(self):
+        # raise NotImplementedError("Can't test semantics of grammar, only syntax.")
+
+    # def test_undefined_term(self):
+        # raise NotImplementedError("Can't test semantics of grammar, only syntax.")
+
+    # def test_token_multiline_only_works_with_x_flag(self):
+        # raise NotImplementedError("Can't test regex flags in Lark grammar.")
+
+    # def test_import_custom_sources(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_import_custom_sources2(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_import_custom_sources3(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_my_find_grammar_errors(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_ranged_repeat_terms(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_ranged_repeat_large(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_large_terminal(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+    # def test_list_grammar_imports(self):
+        # raise NotImplementedError("Can't test semantics of grammar, only syntax.")
+
+    def test_inline_with_expand_single(self):
+        l = Lark(self.lark_grammar, parser="lalr")
+        grammar = r"""
+        start: _a
+        !?_a: "A"
+        """
+        # self.assertRaisesRegex(GrammarError, "Inlined rules (_rule) cannot use the ?rule modifier.", l.parse, grammar)
+        # TODO Is this really catching the right problem?
+        self.assertRaisesRegex(UnexpectedToken, "Unexpected token Token.'OP', '?'.", l.parse, grammar)
+
+
+    # def test_line_breaks(self):
+        # raise NotImplementedError("Can't parse using parsed grammar.")
+
+
+if __name__ == '__main__':
+    main()
--- a/tests/test_lark_lark.py
+++ b/tests/test_lark_lark.py
@@ -1,0 +1,165 @@
+from __future__ import absolute_import
+
+import os
+from unittest import TestCase, main
+
+from lark import lark, Lark, UnexpectedToken
+from lark.load_grammar import GrammarError
+
+
+# Test that certain previous differences between load_grammar.py and 
+# grammars/lark.lark have been resolved.
+class TestLarkLark(TestCase):
+    def setUp(self):
+        lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark')
+        # lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark-ORIG')
+        self.lark_parser = Lark.open(lark_path, parser="lalr")
+
+    def test_01_no_alias_in_terminal_lg(self):
+        g = """start: TERM
+            TERM: "a" -> alias
+            """
+        self.assertRaisesRegex( GrammarError, "Aliasing not allowed in terminals", Lark, g)
+
+    def test_01_no_alias_in_terminal_ll(self):
+        g = """start: TERM
+            TERM: "a" -> alias
+            """
+        self.assertRaisesRegex( UnexpectedToken, "Unexpected token Token.'__ANON_0', '->'.", self.lark_parser.parse, g)
+
+    def test_02_no_rule_aliases_below_top_level_lg(self):
+        g = """start: rule
+            rule: ("a" -> alias
+                 | "b")
+            """
+        self.assertRaisesRegex( GrammarError, "Rule 'alias' used but not defined", Lark, g)
+
+    def test_02_no_rule_aliases_below_top_level_ll(self):
+        g = """start: rule
+            rule: ("a" -> alias
+                 | "b")
+            """
+        self.assertRaisesRegex( UnexpectedToken, "Unexpected token Token.'__ANON_0', '->'.", self.lark_parser.parse, g)
+
+    def test_03_ignore_single_token_lg(self):
+        g = """start: TERM
+        %ignore "a" "b" /c/
+        TERM: "d"
+        """
+        # This SHOULD raise some sort of error, but silently discards the extra tokens instead.
+        # self.assertRaises( UnexpectedToken, Lark, g)
+        Lark(g)
+
+    def test_03_ignore_single_token_ll(self):
+        g = """start: TERM
+        %ignore "a" "b" /c/
+        TERM: "d"
+        """
+        self.assertRaisesRegex( UnexpectedToken, "Unexpected token Token.'STRING', '.b.'.", self.lark_parser.parse, g)
+
+    def test_04_extend_rule_lg(self):
+        g = """
+            %import .grammars.ab (startab, A, B, expr)
+
+            %extend expr: B A
+        """
+        Lark(g, start='startab', source_path=__file__)
+
+    def test_04_extend_rule_ll(self):
+        g = """
+            %import .grammars.ab (startab, A, B, expr)
+
+            %extend expr: B A
+        """
+        self.lark_parser.parse(g)
+
+    def test_05_extend_term_lg(self):
+        g = """
+            %import .grammars.ab (startab, A, B, expr)
+
+            %extend A: "c"
+        """
+        Lark(g, start='startab', source_path=__file__)
+
+    def test_05_extend_term_ll(self):
+        g = """
+            %import .grammars.ab (startab, A, B, expr)
+
+            %extend A: "c"
+        """
+        self.lark_parser.parse(g)
+
+    def test_06_no_term_templates_lg(self):
+        g = """start: TERM
+        separated{x, sep}: x (sep x)* 
+        TERM: separated{"A", " "}
+        """
+        self.assertRaises( AssertionError, Lark, g)
+
+    def test_06_no_term_templates_ll(self):
+        g = """start: TERM
+        separated{x, sep}: x (sep x)* 
+        TERM: separated{"A", " "}
+        """
+        self.assertRaisesRegex( UnexpectedToken, "Unexpected token Token.'RULE', 'separated'.", self.lark_parser.parse, g)
+
+    def test_07_term_no_call_rule_lg(self):
+        g = """start: TERM
+        TERM: rule
+        rule: "a"
+        """
+        self.assertRaisesRegex( GrammarError, "Rules aren't allowed inside terminals", Lark, g)
+
+    def test_07_term_no_call_rule_ll(self):
+        g = """start: TERM
+        TERM: rule
+        rule: "a"
+        """
+        self.assertRaisesRegex( UnexpectedToken, "Unexpected token Token.'RULE', 'rule'.", self.lark_parser.parse, g)
+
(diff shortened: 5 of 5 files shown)


docs/grammar.md
```diff
@@ -59,26 +59,37 @@ Terminals are used to match text into symbols. They can be defined as a combinat
 **Syntax:**
 
 ```html
-<NAME> [. <priority>] : <literals-and-or-terminals>
+<NAME> [. <priority>] : <items-to-match>
```
erezsh, line 62: "items" may be more correct, but to a newcomer it sounds ambiguous. What kind of items?

Perhaps below should be an enumeration of the kinds of items possible (i.e. rules, terminals, templates)

docs/grammar.md
```diff
@@ -59,26 +59,37 @@ Terminals are used to match text into symbols. They can be defined as a combinat
...
 ```
 
-Terminal names must be uppercase.
+Terminal names must be uppercase.  They must start with an underscore (`_`) or a letter (`A` through `Z`), and may be composed of letters, underscores, and digits (`0` through `9`).  Terminal names that start with "_" will not be included in the parse tree, unless the `keep_all_tokens` option is specified.
```
erezsh, line 65: Since you add a lot of detail, worth mentioning that terminals that appear in other terminals (i.e. terminal fragments) will be included even if they start with `_`.

docs/grammar.md
```diff
@@ -59,26 +59,37 @@ Terminals are used to match text into symbols. They can be defined as a combinat
...
 
 Literals can be one of:
 
-* `"string"`
-* `/regular expression+/`
-* `"case-insensitive string"i`
-* `/re with flags/imulx`
-* Literal range: `"a".."z"`, `"1".."9"`, etc.
+* Literal range: `"a".."z"`, `"1".."9"`, etc. - Each literal must be a single character, and the range represends all values between the two literals, inclusively.
```
erezsh, line 69: If  you're going to remove or move the list of literals elsewhere, might as well remove/move all of it. This one single item separately is just confusing.

docs/grammar.md
```diff
@@ -59,26 +59,37 @@ Terminals are used to match text into symbols. They can be defined as a combinat
...
 
-Terminals also support grammar operators, such as `|`, `+`, `*` and `?`.
+Each item is one of:
```
erezsh, line 71: This list already appears below. Why duplicate it?

docs/grammar.md
```diff
@@ -297,12 +315,24 @@ Note that `%ignore` directives cannot be imported. Imported rules will abide by
...
+
 ### %override
 
 Override a rule or terminals, affecting all references to it, even in imported grammars.
 
 Useful for implementing an inheritance pattern when importing grammars.
 
+**Syntax:**
+```html
+%override <TERMINAL> ... terminal definition ...
```
erezsh, line 332: Not accurate..

docs/tree_construction.md
```diff
@@ -74,6 +74,7 @@ Lark will parse "((hello world))" as:
 The brackets do not appear in the tree by design. The words appear because they are matched by a named terminal.
 
 
+<a name="shaping_the_tree"></a>
```
erezsh, line 77: This should be automatic, no?

tests/test_grammar_formal.py
```diff
@@ -0,0 +1,169 @@
...
+from lark.load_grammar import GrammarError
+
+
+# Based on TestGrammar, with lots of tests that can't be run elided.
+class TestGrammarFormal(TestCase):
+    def setUp(self):
+        lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark')
+        # lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark-ORIG')
+        with open(lark_path, 'r') as f:
+            self.lark_grammar = "\n".join(f.readlines())
```
erezsh, line 16: Why not just f.read() ?

tests/test_grammar_formal.py
```diff
@@ -0,0 +1,169 @@
...
+        # 'Terminal names cannot contain dots', 'A.B\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..TOKEN., .B..', l.parse, 'A.B\n')
+        # 'Expecting rule or terminal definition', '"a"\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..STRING., ."a"..', l.parse, '"a"\n')
+        # '%import expects a name', '%import "a"\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..STRING., ."a"..', l.parse, '%import "a"\n')
+        # '%ignore expects a value', '%ignore %import\n'
+        self.assertRaisesRegex(UnexpectedToken, 'Unexpected token Token..__ANON_2., .%import..', l.parse, '%ignore %import\n')
+
+    # def test_empty_literal(self):
```
erezsh, line 79: Why all the commented out parts?

lark/grammars/lark.lark
```diff
@@ -7,46 +7,66 @@ _item: rule
...
+            | (REGEXP | STRING) -> literal
 
 name: RULE
     | TOKEN
 
 _VBAR: _NL? "|"
 OP: /[+*]|[?](?![a-z])/
-RULE: /!?[_?]?[a-z][_a-z0-9]*/
+RULE: /_?[a-z][_a-z0-9]*/
+RULE_MODIFIERS: /!|![?](?=[a-z])|[?]!?(?=[a-z])/
```
erezsh, line 69: Why do you need the `?=`s here? I don't recall that ? or ! are valid anywhere else in the grammar..?

lark/grammars/lark.lark
```diff
@@ -7,46 +7,66 @@ _item: rule
...
-?expansion: expr*
+?token_expansion: token_expr*
 
-?expr: atom [OP | "~" NUMBER [".." NUMBER]]
+?token_expr: token_atom [OP | "~" NUMBER [".." NUMBER]]
 
-?atom: "(" expansions ")"
-     | "[" expansions "]" -> maybe
-     | value
+?token_atom: "(" token_expansions ")"
```
erezsh, line 55: token_atom and rule_atom are the same, why not merge them using templates? Same with rule_expr and atom_expr, rule_expansion etc.

erezsh (changes requested):
Here are some suggestions and questions I have regarding your PR. I tried to get everything down but I might think of more after you address them.

docs/grammar.md
RossPatterson: Good point, will do.

docs/grammar.md
RossPatterson: Whoops, that was a late-evening editting error.  I'll fix it up.

docs/grammar.md
RossPatterson: The item list for terminals is different from the list later for rules, in that a terminal's items cannot include rules.  But yeah, it's easy to overlook that in the big wall of text.

Honestly, if I was writing this document from scratch, I'd have written the full grammar in BNF and avoided the simplifying statements.  But I assume your audience here is people without a CS background who may never have written a parser by hand, and who might find BNF opaque.

docs/grammar.md
RossPatterson: https://github.com/lark-parser/lark/blob/master/lark/load_grammar.py#L162-L165 and https://github.com/lark-parser/lark/blob/master/lark/load_grammar.py#L1260-L1262 make it look like you can override a terminal, by supplying the full definition, preceded by `%override`.  What did I misunderstand?

Again, if the description was less words and more productions, I'd have just written `<override> ::= "%override" (rule|term)`.

docs/tree_construction.md
RossPatterson: I'd hope so, but I saw you used a similar construct at https://github.com/lark-parser/lark/blob/master/docs/grammar.md?plain=1#L242, and Sphinx ain't one of my strong suits, so I assumed an HTML anchor was necessary.

tests/test_grammar_formal.py
RossPatterson: A bad idiom that I use all over the place :-)  I'll fix it up.

tests/test_grammar_formal.py
RossPatterson: I wanted it to be clear which `test_grammar` tests weren't implementable for `lark.lark`.  Sort of a "_show your work_" thing.  But I can remove them.

MegaIng:
My general opinion is that `lark.lark` should reflect the syntax as defined by the BNF inside of `load_grammar.py` and not the semantic checks implemented on top of that (and therefore most of this PR shouldn't be done in it's current form). But if @erezsh is of a different mind, I am not going to argue too much. I also wouldn't be opposed to making the syntax specfified in `load_grammar.py` more restrictive and therefore removing semantic checks in other parts of the code. My "dream" would be that the BNF hard coded in `load_grammar.py` could be generated from `lark.lark` and we only have one source of truth. Further diverging the two grammars is IMO not a good idea.

RossPatterson:
> My "dream" would be that the BNF hard coded in `load_grammar.py` could be generated from `lark.lark` and we only have one source of truth. 

Mine as well.  I was actually quite surprised at how different they were.  It was your comments in Issue #1388 that pointed me down this part.

> Further diverging the two grammars is IMO not a good idea.

I guess I see it more as specifying the same grammar in two forms, not two grammars.  But you're right about semantics - there are some things (_e.g._, my comments about `literal_range`) that Lark can't easily express today, and that require post-processing.

RossPatterson:
I missed two lines that I should have deleted in the test files, which I used for before/after testing (`... 'grammars/lark.lark-ORIG' ...`).  I'll remove them.

docs/grammar.md
RossPatterson: Fixed in 9493f81e9eab6dbe42096938bbcfc1f0cf5a2135.

tests/test_grammar_formal.py
RossPatterson: Fixed in 9493f81e9eab6dbe42096938bbcfc1f0cf5a2135.

tests/test_grammar_formal.py
RossPatterson: Fixed in 9493f81e9eab6dbe42096938bbcfc1f0cf5a2135.

docs/grammar.md
erezsh: I'm not following the reasoning. Why not have one list, and in rules write: "an item can be everything in that list, plus a rule". Or even put rule as part of the list, and mention that it can't appear in terminals.

docs/grammar.md
erezsh: Sorry, I should have been more specific.

The syntax is `%override <terminal definition>`, or `%override <TERMINAL> : <items>`

Either way, but the way you write it is confusing, and would make the reader think the syntax is different.

docs/tree_construction.md
erezsh: Funny, I'm really not sure why it's there. But as you can see, it isn't needed: https://lark-parser.readthedocs.io/en/stable/grammar.html#directives

tests/test_grammar_formal.py
erezsh: Very bad idiom :)

tests/test_grammar_formal.py
erezsh: I appreciate your aim for completeness, but I think it was more confusing than anything.

If you think it's important information, perhaps just write a list of the names, without Python syntax. like

```python
# List of non-testable constructs: empty literal, kraken, foobar, etc.
```

lark/grammars/lark.lark
erezsh: Oh I see. But wouldn't users be able to bypass it by adding whitespace between them?

lark/grammars/lark.lark
erezsh: Before we dig into that, let's come to an agreement in the main conversation (regarding restrictive-validating grammar vs relaxed grammar + validating visitor).

Afterwards if it's still relevant, I will try to craft a solution to this problem.

RossPatterson:
> As for what to do with lark.lark, in my mind it's still an open question, and depends heavily on what are its main uses. Is it to parse and manipulate lark grammars programatically?

That's part of what drove me down this path.  I've got a use case where I'm playing with @MegaIng's `lark2railroad`, combined with my own recent contributions to `railroad-diagrams`, to generate documentation for small programs that use Lark to parse their input.

> Is it to validate correct grammars?

That's another part of my motivation.  As noted in Issue #1355, I attempted to use aliases on rule-alternates inside a rule-group, and received a weird result (from `load_grammar.py`, which as you implied, has great contextual error messages).  It turns out, although neither `lark.lark` nor the documetation says so, that's invalid syntax.  Now, I can simply avoid doing that again, but I try to contribute incremental improvements to open source projects that I use, so here we are :-)

> I think the options that we are facing are:
> ...
> What do you guys think?

I think having an inaccurate reference grammar for human purposes is almost worse than not having one at all.  I think having a partial grammar for parsing purposes is somewhat useful, but will be surprising to anyone who tries to use it reliably.  And I think better documentation might obviate the need for a perfect reference grammar, except for the cases like `lark2railroad` that are actually trying to parse Lark grammars from random authors.

I'm not going to be offended by however the decision turns out.  If it's something I can help with, fine, if not, also fine.  This isn't my project, I'm just a (so far very happy) user who has the skills to contribute a bit.

docs/grammar.md
RossPatterson: Fixed by 83a374f5d00cd1f6bf3b43d5bf799cf7ef804cb3.

docs/tree_construction.md
RossPatterson: Fixed by 83a374f5d00cd1f6bf3b43d5bf799cf7ef804cb3.

erezsh:
> I think having an inaccurate reference grammar for human purposes

@RossPatterson Can you explain which option this refers to? I don't think I advocated at any point for an inaccurate grammar, only a non-validating one

> which as you implied, has great contextual error messages

The load_grammar error messages are definitely better than the generic lark error messages. If I overlooked or missed something, it's not proof that the approach is bad

Do you have any objections to option 2 that I suggested? Do you think it doesn't serve your efforts as well as option 3? Do you think it's more confusing?

Frankly, I don't accept or rejects PR based on who will be offended. I'm just trying to make Lark as good as it can be. (with the limited time that I can spare for it)

lark/grammars/lark.lark
RossPatterson: Drat!  I missed a test case.  But no, this test fails similarly in `load_grammar.py`, the original `lark.lark`, and my modified version:

```        g = """start: rule1
            rule1: rule2
            ?! rule2: "a"
        """
```

`load_grammar.py`:
```
  File "C:\Ross\Source\lark\lark\parsers\lalr_parser_state.py", line 77, in feed_token
    action, arg = states[state][token.type]
KeyError: 'OP'

During handling of the above exception, another exception occurred:

  File "C:\Ross\Source\lark\tests\test_lark_lark.py", line 169, in test_11_rule_modifier_query_bang_space_lg
    Lark(g)
...
  File "C:\Ross\Source\lark\lark\lexer.py", line 674, in lex
    raise UnexpectedToken(token, e.allowed, state=parser_state, token_history=[last_token], terminals_by_name=self.root_lexer.terminals_by_name)
lark.exceptions.UnexpectedToken: Unexpected token Token('OP', '?') at line 3, column 13.
Expected one of:
        * "%declare"
        * "%import"
        * "%extend"
        * RULE
        * "%override"
        * _NL
        * $END
        * "%ignore"
        * TOKEN
        * RULE_MODIFIERS
Previous tokens: [Token('_NL', '\n            ')]
```

Original `lark.lark`:
```
  File "C:\Ross\Source\lark\lark\lexer.py", line 665, in lex
    yield lexer.next_token(lexer_state, parser_state)
  File "C:\Ross\Source\lark\lark\lexer.py", line 598, in next_token
    raise UnexpectedCharacters(lex_state.text, line_ctr.char_pos, line_ctr.line, line_ctr.column,
lark.exceptions.UnexpectedCharacters: No terminal matches '?' in the current parser context, at line 3 col 13

            ?! rule2: "a"
            ^
...
During handling of the above exception, another exception occurred:

  File "C:\Ross\Source\lark\tests\test_lark_lark.py", line 176, in test_11_rule_modifier_query_bang_space_ll
    self.lark_parser.parse(g)
...
  File "C:\Ross\Source\lark\lark\lexer.py", line 674, in lex
    raise UnexpectedToken(token, e.allowed, state=parser_state, token_history=[last_token], terminals_by_name=self.root_lexer.terminals_by_name)
lark.exceptions.UnexpectedToken: Unexpected token Token('OP', '?') at line 3, column 13.
Expected one of:
        * _NL
        * "%import"
        * "%ignore"
        * "%override"
        * TOKEN
        * $END
        * "%declare"
        * RULE
Previous tokens: [Token('_NL', '\n            ')]
```

Modified `lark.lark`:
```
  File "C:\Ross\Source\lark\lark\lexer.py", line 665, in lex
    yield lexer.next_token(lexer_state, parser_state)
  File "C:\Ross\Source\lark\lark\lexer.py", line 598, in next_token
    raise UnexpectedCharacters(lex_state.text, line_ctr.char_pos, line_ctr.line, line_ctr.column,
lark.exceptions.UnexpectedCharacters: No terminal matches '?' in the current parser context, at line 3 col 13

            ?! rule2: "a"
            ^
...
During handling of the above exception, another exception occurred:

  File "C:\Ross\Source\lark\tests\test_lark_lark.py", line 176, in test_11_rule_modifier_query_bang_space_ll
    self.lark_parser.parse(g)
...
  File "C:\Ross\Source\lark\lark\lexer.py", line 674, in lex
    raise UnexpectedToken(token, e.allowed, state=parser_state, token_history=[last_token], terminals_by_name=self.root_lexer.terminals_by_name)
lark.exceptions.UnexpectedToken: Unexpected token Token('OP', '?') at line 3, column 13.
Expected one of:
        * "%declare"
        * RULE
        * RULE_MODIFIERS
        * $END
        * "%ignore"
        * "%extend"
        * "%import"
        * TOKEN
        * _NL
        * "%override"
Previous tokens: [Token('_NL', '\n            ')]
```

RossPatterson:
> > I think having an inaccurate reference grammar for human purposes
> 
> @RossPatterson Can you explain which option this refers to? I don't think I advocated at any point for an inaccurate grammar, only a non-validating one

Sorry, I misunderstood.

> The load_grammar error messages are definitely better than the generic lark error messages. If I overlooked or missed something, it's not proof that the approach is bad

No, in fact, like I said, it's great.  The `load_grammar.py` code is hard to understand, but the error messages are great.

> Do you have any objections to option 2 that I suggested? Do you think it doesn't serve your efforts as well as option 3? Do you think it's more confusing?

No, I have no objections to option 2.  My efforts will be served as long as `lark.lark` can parse legitimate Lark grammars, and as long as I know the things I shouldn't do.

> Frankly, I don't accept or reject PR based on who will be offended. I'm just trying to make Lark as good as it can be. (with the limited time that I can spare for it)

I understand completely.  I was trying to say that I'm OK with whatever you decide to do.

erezsh:
@MegaIng Any objections to option 2? i.e. keep lark.lark non-validating and pair it with a visitor that performs the same extra validation as load_grammar?

MegaIng:
Currently no, I think that is the ideal solution. Preferably the same visitor is used both internally and externally and is the only source of GrammerErrors in load_grammer  to make sure that everything is in sync.

MegaIng:
> That means getting lark.lark as close as possible to load_grammar

To note: lark.lark should probably still use ebnf and not restrict itself to bnf like the hardcoded grammar does, and IMO it's ok to change the grammar in load_grammar a bit if it makes stuff easier to sync up (if it doesn't have a performance hit and ofcourse doesn't break anything)

RossPatterson:
> @RossPatterson Would you be okay adapting your PR to solution 2?
>
> That means getting lark.lark as close as possible to load_grammar, and adding a validating visitor for the rest (or somehow using the one in load_grammar)

Yup, I can do that.  I've already had to do the analysis to find the differences, and to fix them.  So it looks like I'll just be backing out some of the fixes, and creating their equivalent in a visitor.

> If you'd like, we can split this PR to two, one for the docs, which I think we all agree can be merged already as is (or with tiny changes), and a separate one for lark.lark. But it's up to you.

I'm fine with keeping them together.

RossPatterson:
> To note: lark.lark should probably still use ebnf and not restrict itself to bnf like the hardcoded grammar does, 

I'm personally biased in favor of BNF, because it's so well defined, whereas "EBNF" means any of a variety of possibilities.  But I agree, `lark.lark` should continue to be in the Lark variant of EBNF.

> IMO it's ok to change the grammar in load_grammar a bit if it makes stuff easier to sync up (if it doesn't have a performance hit and ofcourse doesn't break anything)

I don't anticipate that being necessary.  But I've been wrong before :-)

RossPatterson:
I sat down tonight to start implementing option 2, and quickly found myself categorizing most of my changes (at the top of this PR) as "_making `lark.lark` accept the same syntax as `load_grammar`_", and therefore I was planning not to revert them.  In fact, when I was done, the only changes I felt weren't either adding omitted directives or removing illegal syntax were three of the last four: 

- 7 Tokens can no longer include rules.
- 9 Rule modifiers are no longer allowed in rule references (_e.g._, `rule1: !?rule2`)
- 10 Rule modifiers are now allowed to be `?!`, in addition to the existing `!`, `?`, and `!?`.

I somehow doubt this is what either of you expect.

I think the changes that correct the `%extend` and `%override` syntax (4, 5, and 8) should obviously be retained.  The changes that correct the token rule syntax (1 and 6) seem appropriate to keep as well.  The rule aliases change (2), `%ignore` syntax (3) and tokens-can't-include-rules (7) are on the fence, and my personal feeling is that 2 and 3 should be kept and 7 should be reverted and moved to the new visitor.  I'm conflicted on rule references (9) - it seems like it's a syntax issue, but it's only implied by the `RULE` terminal definition, it isn't explicitly stated in any rule.  So I'm inclined to revert it an move it to the visitor.  I think the last one, the query-bang rule modifier (10), is a real nit, and ought to be in the visitor instead.

So, before I go off and start messing things up, can you tell me if I'm reading things right or wrong, and if wrong, how so?

MegaIng:
@RossPatterson I think you misunderstood what we (at least I) meant with option 2. At the top of `load_grammar.py`, there is a BNF + terminals description of the exact grammar (the syntax) that is parsed by the file. The idea is that `lark.lark` should *exactly* match that, ignoring every other post processing step (the semantics) done in `load_grammar.py`. Then we also provide an extra visitor somewhere that does all the other validation (i.e. the semantic checks). If you think it's reasonable to move stuff from semantics to syntax, then you can do that, but please that also do the move in the BNF grammar defined in `load_grammar.py`.

RossPatterson:
I reverted all my changes to `lark.lark`, and did what @MegaIng suggested - made it do exactly what the BNF in `load_grammar.py` does.

 1. In `rule`, split out the modifiers to `rule_modifiers`.
 1. In `token`, removed `token_params`.
 1. In `statement`, added `%extend`.
 1. In `statement`, added `%override token`.
 1. In `alias`, the `RULE` usage now does not accept `!` and `?` modifiers (see `RULE_MODIFIERS` below).
 1. In `value` > `template_usage`, changed `name` to `RULE`, to prevent accepting `TOKEN`s.
 1. In `RULE`, split out the rule modifiers to `RULE_MODIFIERS`.


I built the validating visitor, `lark_validator_vistor.py::LarkValidatorVisitor`.  It is invoked by a class-level `validate(tree)` method.  It checks:
 1. That `token`s don't have aliases.
 1. That `%ignore` has only one token specified.  Note that `load_grammar` doesn't do this, you get unexpected results from `%ignore A B`.
 1. That `rule`s don't have `alias`s on "inner `expansions`".  Note that `load_grammar` doesn't do this, you get unexpected results instead.

There wasn't any existing doc on using the `lark.lark` grammar, so I put some instructions for the validator at the top of the grammar.

I also moved the doc for templates in `grammar.md` from the "Terminals" section to the "Rules" section, since you can't use templates with terminals, and left behind the "_Templates are not allowed with terminals._" part

Some notes:
 1. `lark.lark` inlines several rules that `load_grammar` does not.  I haven't changed this, because it will change trees produced by `lark.lark`, but I think I should.
    - `expansions`
    - `expansion`
    - `value`
 1. `load_grammar` inlines one rule that `lark.lark` does not.  I haven't changed this, because it will change trees produced by `lark.lark`, but I think I should.
    - `name`
 1. `load_grammar` defines `OP` as `[+*]|[?](?![a-z_])`, while `lark.lark` omits the underscore in the not-followed-by set.  I assume `load_grammar` does this to help in recognizing the difference between `atom?` and `?rulename:`, but it's a little too arcane for me to be sure.
 1. `load_grammar` defines `RULE_MODIFIERS` as `(!|![?]?|[?]!?)(?=[_a-z])`.  Why does it use both `!` and `![?]?`?  Why not just `![?]?`?
 1. `lark.lark` imports `ESCAPED_STRING` from `common.lark` as `_STRING`, while `load_grammar`, of course, defines it itself.  They're expressed differently, but I believe they accept the same set of values.  I have left them both as they were originally.
 1. `load_grammar` defines `STRING` as `"(\\"|\\\\|[^"\n])*?"i?`.  Why use `*?`?  Isn't that the same as `*`?
 1. `lark.lark` imports `SIGNED_INT` from `common.lark` as `NUMBER`, and `WS_INLINE` from `common.lark`, while `load_grammar`, of course, defines them itself.  They're expressed differently, but they accept the same set of values.  I have left them both as they were originally.
 1. `lark.lark` and `load_grammar` define `COMMENT` slightly differently, but the resulting regexp is identical.  I have left them both as they were originally.

MegaIng:
Very nice, this is about what I imagined :-) Thank you for going through this effort.

> I haven't changed this, because it will change trees produced by `lark.lark`, but I think I should. 

I think that's fine. I don't think there are too many people out there relying on the details of that grammar, and I think we changed it before without warning.

> `lark.lark` inlines several rules that load_grammar does not.

`alias` and `expr` are also never inlined by `lark.lark` despite having a `?` because the `[]` makes it so that we always have a `None` as a second child. I think this should both be changed to use `()?` instead and be inlined. IMO that is more inline with that would be expected from the parsing, matches `load_grammar` better and is probably? easier to work with. But it would require changes in the validator you wrote.

It might make sense to write an `Interpreter` instead of a `Visitor` which allows slighter better control about what happens when.

> There wasn't any existing doc on using the lark.lark grammar, so I put some instructions for the validator at the top of the grammar.

We should have a doc page with a reference list for our "stdlib" grammars, but that can be a different PR. An example using the validator might be nice.

docs/grammar.md
```diff
@@ -51,6 +51,29 @@ Lark begins the parse with the rule 'start', unless specified otherwise in the o
...
 
+## EBNF Expressions
+
+The EBNF expression in a Lark termminal definition is a sequence of items to be matched.
+Each item is one of:
+
+* `TERMINAL` - Another terminal, which cannot be defined in terms of this terminal.
+* `"string literal"` - Literal, to be matched as-is.
+* `"string literal"i` - Literal, to be matched case-insensitively.
+* `/regexp literal/` - Regular expression literal.  Can inclde flags.
```
MegaIng, line 62: I think mentioned here where those flags can be placed is a good idea, not everyone is familar with that syntax:

```
* `/regexp literal/[imslux]` - Regular expression literal.  Can include the stdlib's `re` [flags `imslux`](https://docs.python.org/3/library/re.html#contents-of-module-re)
```

(also typo inclde -> include)

docs/grammar.md
```diff
@@ -51,6 +51,29 @@ Lark begins the parse with the rule 'start', unless specified otherwise in the o
...
+* `"character".."character"` - Literal range.  The range represends all values between the two literals, inclusively.
+* `(item item ..)` - Group items
+* `(item | item | ..)` - Alternate items.
+* `[item item ..]` - Maybe. Same as `(item item ..)?`, but when `maybe_placeholders=True`, generates `None` if there is no match.
+* `[item | item | ..]` - Maybe with alternates. Same as `(item | item | ..)?`, but when `maybe_placeholders=True`, generates `None` if there is no match.
+* `item?` - Zero or one instances of item (a "maybe")
+* `item*` - Zero or more instances of item
+* `item+` - One or more instances of item
+* `item ~ n` - Exactly *n* instances of item
+* `item ~ n..m` - Between *n* to *m* instances of item (not recommended for wide ranges, due to performance issues)
```
MegaIng, line 72: I believe this warning can be removed :-) I added a smarter system some time ago that should allow this with quite decent performance.

docs/grammar.md
```diff
@@ -297,12 +304,24 @@ Note that `%ignore` directives cannot be imported. Imported rules will abide by
 
 Declare a terminal without defining it. Useful for plugins.
 
+**Syntax:**
+```html
+%declare <TERMINAL>
+%declare <rule>
```
MegaIng, line 310: `%declare <rule>` is not actually useable and should also be rejected by `lark.lark` and `load_grammar` (currently it causes an ugly `AttributeError: 'NoneType'`) I think changing the syntax is fine here, unless @erezsh knows of some way to actually use this.

tests/test_lark_lark.py
```diff
@@ -0,0 +1,211 @@
...
+            %extend A: "c"
+        """
+        self.lark_parser.parse(g)
+
+    def test_06_no_term_templates_lg(self):
+        g = """start: TERM
+        separated{x, sep}: x (sep x)*
+        TERM: separated{"A", " "}
+        """
+        self.assertRaisesRegex( AssertionError, "Tree.'template_usage', .NonTerminal.'separated'.", Lark, g)
```
MegaIng, line 83: `AssertionErrors` are not something we want to have in the public interface, this should be changed to a `GrammarError`. (although you don't need to do it now in this PR)

lark/grammars/lark.lark
```diff
@@ -1,25 +1,39 @@
...
+# This grammar matches that one, but does not enfore some rules that it does.
+# If you want to enforce those, you can pass the "LarkValidatorVisitor" over
+# the parse tree, like this:
+
+# import os
+# import lark
+# from lark.lark_validator_visitor import LarkValidatorVisitor
+#
+# lark_path = os.path.join(os.path.dirname(lark.__file__), 'grammars/lark.lark')
+# lark_parser = Lark.open(lark_path, parser="lalr")
```
MegaIng, line 12: For exactly this kind of usecase we have `Lark.open_from_package`:

```
lark_parser = Lark.open_from_package("lark", "grammars/lark.lark")
```

This should also be used in the tests below.

MegaIng (changes requested):
With regard to tests, also see that we are already testing `lark.lark`: https://github.com/lark-parser/lark/blob/7646fb31609c6c83e0998b121eeffb908f5ec5ba/tests/test_parser.py#L1047

It might make sense to extend the tests there to include `LarkValidatorVisitor` and make sure that both also raise the same error messages  although it might be hard to check that we don't expect grammar analysis errors. I don't think `earley` should be raising any? But I might be wrong with that.

docs/grammar.md
RossPatterson: Fixed in 200d6b5f004425174d2364cf701e964c3ca2e2a0.

docs/grammar.md
RossPatterson: Fixed in 200d6b5f004425174d2364cf701e964c3ca2e2a0.

lark/grammars/lark.lark
RossPatterson: Fixed in 200d6b5f004425174d2364cf701e964c3ca2e2a0, including in the test modules.

tests/test_lark_lark.py
RossPatterson: Fixed in 0fb28f9ad16b461821227f2dbfa71a01d84cf363.  I already had an error for this in `LarkValidatorVisitor`: "_Templates not allowed in terminals_".  Thanks to your hint elsewhere about rules not being inlined because they have multiple children, I realized `template_usage` can never be inlined - the minimum is two children: a `RULE` and a `value`.  Which made it easy to add `TerminalTreeToPattern.template_usage()` to catch the bad use.

docs/grammar.md
RossPatterson: Since `%declare` was in both `load_grammar.py` and `lark.lark` before I started, how about we leave that for a new Issue?  I expect someone more familiar with `load_grammar.py` would be better at cutting `%declare rule` out of it than I will.

RossPatterson:
> > I haven't changed this, because it will change trees produced by `lark.lark`, but I think I should.
> 
> I think that's fine. I don't think there are too many people out there relying on the details of that grammar, and I think we changed it before without warning.

OK, I'll remove inlining from `expansions`, `expansion`, and `value` in `lark.lark`.

RossPatterson:
> It might make sense to write an `Interpreter` instead of a `Visitor` which allows slighter better control about what happens when.

The `Visitor` model seems to fit well for the validator.  By its nature, it finds all the `token`s, `alias`es, and `ignore`s, which I'd have to write code to hunt down if I switched to `Interpreter`.  Is there a subtlty I'm missing here?

RossPatterson:
> > There wasn't any existing doc on using the lark.lark grammar, so I put some instructions for the validator at the top of the grammar.
 > 
> We should have a doc page with a reference list for our "stdlib" grammars, but that can be a different PR. An example using the validator might be nice.

If someone wants to write that page, I'll certainly write of an example for the validator.

docs/grammar.md
erezsh: @MegaIng I agree, I can't see a way in which declaring a rule is useful, or even makes sense.

Personally instead of removing it from the grammar, I would just add validation code.  @RossPatterson  I can write the validation code for load_grammar myself, if you prefer. Although it's not exactly rocket science :)

erezsh:
@RossPatterson Nice work. Let me know once you've cleaned it up a bit, and I'll give it another review.

I'll see if I find the time to write that doc page. I imagine it should be pretty short and straight-forward.

RossPatterson:
> `alias` and `expr` are also never inlined by `lark.lark` despite having a `?` because the `[]` makes it so that we always have a `None` as a second child. I think this should both be changed to use `()?` instead and be inlined.  IMO that is more inline with that would be expected from the parsing, matches `load_grammar` better and is probably? easier to work with. But it would require changes in the validator you wrote.

Fixing the validator was pretty easy.  Fixed in e9c026eecdc69a0c030a10184bfd61f1d9e7e04c.

docs/grammar.md
RossPatterson: > @MegaIng I agree, I can't see a way in which declaring a rule is useful, or even makes sense.
> 
> Personally instead of removing it from the grammar, I would just add validation code. @RossPatterson I can write the validation code for load_grammar myself, if you prefer. Although it's not exactly rocket science :)

It proved easier than I expected.  Fixed in 9bf7ddf466979c2f61d6e67d9c4b1e19b3f65e0d.

RossPatterson:
> With regard to tests, also see that we are already testing `lark.lark`:
> 
> https://github.com/lark-parser/lark/blob/7646fb31609c6c83e0998b121eeffb908f5ec5ba/tests/test_parser.py#L1047
> 
> It might make sense to extend the tests there to include `LarkValidatorVisitor` and make sure that both also raise the same error messages although it might be hard to check that we don't expect grammar analysis errors. I don't think `earley` should be raising any? But I might be wrong with that.

I'm working on blending my `test_lark_lark.py` and `test_grammar_formal.py` into `test_parser.py` and `test_grammar.py`.   I'm using the technique used in `test_parser.py` for running a set of tests against multiple parsers.

RossPatterson:
> I'm working on blending my `test_lark_lark.py` and `test_grammar_formal.py` into `test_parser.py` and `test_grammar.py`. I'm using the technique used in `test_parser.py` for running a set of tests against multiple parsers.

While doing that, I found that neither the original `lark.lark` nor my version pass `test_grammar.py` testcase `test_line_breaks`.  Apparently `load_grammar.py` accepts escaping of newlines (_i.e._, backslash-newline) to continue a statement.  If I'm following things correctly, that's because `load_grammar` includes `BACKSLASH` in its ignore-set, while `lark.lark` does not.  

Fixed in 7f02bd130b40b46fcd2709986e9e2a855c82cc9d.

RossPatterson:
I'll be offline for the next few weeks, but I'll wrap this up after I return in mid-March.

RossPatterson:
While testing and verifying my changes, I realized there are a few more differences between `load_grammar.py` and `lark.lark`, which result in differently-shaped parse trees.  Specifically:

1. The `load_grammar.py` version of `rule` pushes the optionality of `rule_modifiers` and `priority`down into `rule_modifiers` and `priority`,  while `lark.lark` does not.  That means the `load_grammar.py` tree always contains nodes for `rule_modifiers` and `priority` node, while the `lark.lark` tree may not.  I'm going to change `lark.lark` to do what `load_grammar.py` does, because the validator is affected by this difference.  FYI, `term` does not do the same thing with `priority` as `rule` does.  Both `load_grammar.py` and `lark.lark` agree that the `priority` is optional in `term`.
2. `load_grammar.py` calls token names `TERMINAL`, and calls token definitions `term`, while `lark.lark` uses `TOKEN` and `token`.  I'm not going to change this, because it seems like a big change.
3. The `load_grammar.py` version of `value` splits the recognition of `RULE` and `TERMINAL` into `nonterminal` and `terminal`, while  `lark.lark` uses the combined `name` (which exists in both grammars).  I'm not going to change this, because  leaving 2 above intact means this will be different anyway.
4.  In `load_grammar.py`, `name` is inlined, while in `lark.lark`, it is not.  I don't know what to do about this, if anything.

RossPatterson:
> 1. The `load_grammar.py` version of `rule` pushes the optionality of `rule_modifiers` and `priority`down into `rule_modifiers` and `priority`,  while `lark.lark` does not.

FIxed by 4f7a5ebacaf3afe679ba552aba76a7a6a722b68a.

RossPatterson:
@erezsh @MegaIng
Can one of you take a look at `test_grammar.TestGrammar.test_undefined_term`?  It expects a `GrammarError`, but the error that is actually raised seems wrong.  It says `Rule 'A' used but not defined (in rule start)"`, but the entire grammar is just `start: A`.  Shouldn't it raise "**Terminal** 'A' ..." instead?

MegaIng:
> @erezsh @MegaIng Can one of you take a look at `test_grammar.TestGrammar.test_undefined_term`? It expects a `GrammarError`, but the error that is actually raised seems wrong. It says `Rule 'A' used but not defined (in rule start)"`, but the entire grammar is just `start: A`. Shouldn't it raise "**Terminal** 'A' ..." instead?

That's a bug that was introducded in #1018 that we both missed, specfically see here:

https://github.com/lark-parser/lark/blob/706190849ee4529cfc852bc1adb86f1aab11c560/lark/load_grammar.py#L1367-L1369

That shouldn't be `d.is_term`, that is referring to whether we are in the definition of a symbol/rule, but `_grammar_error` uses it to refer to the name itself.

RossPatterson:
> That's a bug that was introducded in #1018 that we both missed, specfically see here:
> ...
> That shouldn't be `d.is_term`, that is referring to whether we are in the definition of a symbol/rule, but `_grammar_error` uses it to refer to the name itself.

Fixed in 40576d2383f57b58a25d303b599bfcc54cbe23dc.

RossPatterson:
As of daac65d84b1d61cf4ff3e5eb00ce99f1b5a5ae02, everything I intended to do is done.  @erezsh @MegaIng, I think this meets your expectations, and is ready for code review.

RossPatterson:
Is this work still of interest?  I think it's ready for final review, and probably for merging.

lark/lark_validator.py
```diff
@@ -0,0 +1,265 @@
...
+    def _load_grammar(self, tree: Tree) -> None:
+        for stmt in tree.children:
+            if stmt.data == 'declare':
+                self._declare(stmt)
+            elif stmt.data == 'extend':
+                self._extend(stmt)
+            elif stmt.data == 'ignore':
+                self._ignore(stmt)
+            elif stmt.data in ['import', 'multi_import']:
+                # TODO How can we process imports in the validator?
```
MegaIng, line 170: imports always explicitly list the names being imported, so we can treat them like `%declare`, which should be decent enough, right?

Or maybe that produces problems with `%extend`, hm...

lark/lark_validator.py
```diff
@@ -0,0 +1,265 @@
...
+class Definition:
+    def __init__(self, is_term, tree, params=(), options=None):
+        self.is_term = is_term
+        self.tree = tree
+        self.params = tuple(params)
+
+class LarkValidator:
+
+    @classmethod
+    def validate(cls, tree: Tree, options: Dict[str, Any] = {}):
```
MegaIng, line 18: Why is there an `options` argument? I don't see it being used anywhere. Was the idea that it selects which checks to perform? I wouldn't be opposed to that being possible.

lark/lark_validator.py
```diff
@@ -0,0 +1,265 @@
...
+from .load_grammar import eval_escaping
+from .tree import Tree
+
+class Definition:
+    def __init__(self, is_term, tree, params=(), options=None):
+        self.is_term = is_term
+        self.tree = tree
+        self.params = tuple(params)
+
+class LarkValidator:
```
MegaIng, line 15: Please add a docstring here and in `validate`, something like:

```
Checks a grammar parsed by `lark.lark` for validated using a variety of checks similar to what `load_grammar.py` does on parser creation. The only stable public entry point is `LarkValidator.validate(tree)`

Checks:
- Illegal constructs not prevented by the grammar
  - `alias` not in the top expansions of a rule
  - Incorrect `%ignore` lines
  - Invalid literals (like newlines inside of regex without the `x` flag)
  - Rules used inside of Terminals
- Undefined symbols
- Incorrectly used templates
```

MegaIng: Also we should make sure that this is exposed in the docs, @erezsh can probably say what needs to be done for that.

MegaIng (changes requested):
Thank you again for all your effort, sorry for not responding for so longer, I had other things on my mind.

Looks pretty good for the most part! Ofcourse, there is now a lot of duplication between `load_grammar.py` and `lark_validator.py`, but I think that can be reduced in a future PR and is fine for now.

lark/grammars/lark.lark
```diff
@@ -1,56 +1,70 @@
...
 start: (_item? _NL)* _item?
 
 _item: rule
      | token
      | statement
 
-rule: RULE rule_params priority? ":" expansions
-token: TOKEN token_params priority? ":" expansions
+rule: rule_modifiers RULE rule_params priority ":" expansions
+token: TOKEN priority? ":" expansions
```
erezsh, line 21: `priority` is already optional

tests/test_grammar.py
```diff
@@ -1,304 +1,437 @@
...
+        options.update(kwargs)
+        if "start" in options and options["start"] != "start":
+            # We're not going to parse with the parser, so just override it.
+            options["start"] = "start"
+        options["propagate_positions"] = True
+        lark_parser = Lark.open_from_package("lark", "grammars/lark.lark", **options)
+        tree = lark_parser.parse(grammar)
+        LarkValidator.validate(tree)
+
+    def parse(self, text: str, start=None, on_error=None):
```
erezsh, line 27: Looks to me like it would be better if the tests that try to call `parse()` wouldn't be inside `_make_tests`, since their dot-lark variation will be skipped anyway.

tests/test_grammar.py
```diff
@@ -1,304 +1,437 @@
...
+
+class LarkDotLark:
+    def __init__(self, grammar, **kwargs):
+        options = {}
+        options.update(kwargs)
+        if "start" in options and options["start"] != "start":
+            # We're not going to parse with the parser, so just override it.
+            options["start"] = "start"
+        options["propagate_positions"] = True
+        lark_parser = Lark.open_from_package("lark", "grammars/lark.lark", **options)
```
erezsh, line 23: Why are you passing `options` when you parse the `lark.lark` grammar? I think the right thing here is to ignore them, because they shouldn't affect parsing the grammar.

tests/test_grammar.py
```diff
@@ -1,304 +1,437 @@
...
+                %override sep{item, delim}: item (delim item)* delim?
+                %ignore " "
+            """, source_path=__file__)
+
+            a = p.parse('[1, 2, 3]')
+            b = p.parse('[1, 2, 3, ]')
+            assert a == b
+
+        def test_override_rule2(self):
+            if parser == LarkDotLark:
```
erezsh, line 72: If you always skip for LarkDotLark, why not take it out of `_make_tests`?

Same question applies to the other tests with the same pattern.

lark/grammars/lark.lark
```diff
@@ -1,56 +1,70 @@
 # Lark grammar of Lark's syntax
 # Note: Lark is not bootstrapped, its parser is implemented in load_grammar.py
+# This grammar matches that one, but does not enfore some rules that it does.
```
erezsh, line 3: Enforce

erezsh (changes requested):
Adding to @MegaIng 's review.

tests/test_grammar.py
```diff
@@ -1,304 +1,437 @@
...
+    def parse(self, text: str, start=None, on_error=None):
+        raise SkipTest("Cannot test cases with lark.lark that try to parse using the tested grammar.")
+
+
+def _make_tests(parser):
+    class _TestGrammar(TestCase):
+        def test_errors(self):
+            # TODO test_errors needs a lot of work for lark.lark.
+            if parser == LarkDotLark:
+                self.skipTest("test_errors needs a lot of work for lark.lark.")
```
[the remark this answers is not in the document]
RossPatterson: Yeah, that was really vague.  What I meant is that `test_errors()` needs some rebuilding to use it for `lark.lark` as well as `load_grammar()`.  It's not infeasible, but I didn't attempt it.

tests/test_grammar.py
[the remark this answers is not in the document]
RossPatterson: Will do, especially given then `test_override_rule1()` example.

tests/test_lark_validator.py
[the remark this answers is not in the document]
RossPatterson: Agreed.  Will delete.

lark/lark_validator.py
RossPatterson: Right, there's `%extend` and `%override`, at least.  I didn't feel up to recreating the logic of %import.

lark/lark_validator.py
RossPatterson: That was left over from an earlier version of the validator.  I honestly don't remember what I did with it.  But you're right, it's dead code.  I think it ought to be removed, rather than leave it as a place to hang future functionality.  OK?

lark/lark_validator.py
RossPatterson: Will do.

lark/grammars/lark.lark
RossPatterson: Yes, but `load_grammar.py` says [`priority` is a required element of `rule`](https://github.com/lark-parser/lark/blob/master/lark/load_grammar.py#L114), and that [`priority` is `_DOT NUMER` or null](https://github.com/lark-parser/lark/blob/master/lark/load_grammar.py#L117).  I wanted `lark.lark` to produce the same parse tree as `load_grammar.py`.

It's different for `token` (`term` in `load_grammar.py`) - there, `load_grammar.py` [says `priority` is optional[(https://github.com/lark-parser/lark/blob/master/lark/load_grammar.py#L162-L163).

tests/test_grammar.py
RossPatterson: I had some hope when I wrote that stuff that I could solve the "can't call parse()" problem.  I made several attempts, all failed.  If you want me to move them out, I will.

tests/test_grammar.py
RossPatterson: `lark.lark` requires the starting rule to be named "start".  Some of the tests don't have one, or have another starting rule.  For example, `test_override_terminal()`.  Without overriding `options['start']`, `lark.lark` won't load those test grammars.

The `propagate_positions` is a debugging leftover.  I'll remove it.

tests/test_grammar.py
RossPatterson: It seemed easier to not break the file up that way.  If you want me to do that, I will.

tests/test_grammar.py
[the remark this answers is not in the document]
RossPatterson: Fixed by commit 5f37365b542402ab66d2616d929601be34eadcd4.

tests/test_lark_validator.py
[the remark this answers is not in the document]
RossPatterson: Fixed by commit 697841b7da28aac8ffc214cfb5c3b029f57fdfa4.

lark/lark_validator.py
RossPatterson: Fixed by commit 654e102300faeeba277841c3fe0af75ad66f41e5.

lark/grammars/lark.lark
RossPatterson: Fixed by commit 33d7088c59cb704b235b8427666ebf4296e52686.

tests/test_grammar.py
RossPatterson: `propagate_positions` fixed by commit 0d01fe224fa05fd3575064247c3e8d749ea2e84f.

tests/test_grammar.py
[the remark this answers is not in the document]
RossPatterson: @MegaIng If my comment of 2024-06-20 is acceptable, let's resolve this point.

lark/lark_validator.py
RossPatterson: If my comment of 2024-06-20 is acceptable, let's resolve this point.

lark/lark_validator.py
RossPatterson: Fixed by commit 20302ca65239bf952c3a10e4acd26b2f625b2a43.

lark/grammars/lark.lark
RossPatterson: @erezsh If my comment of 2024-06-20 is acceptable, let's resolve this point.

tests/test_grammar.py
RossPatterson: @erezsh Would you like me to do that?

erezsh: If I said so, then I guess so, yes :)

Unless there is a reason not to.

tests/test_grammar.py
erezsh: If it's not too much trouble, yes.

tests/test_grammar_formal.py
RossPatterson: I don't think it's important information.  Resolving this section of the PR.

tests/test_grammar.py
RossPatterson: Fixed by ff01d9650a72605cbdb12c399c13d91c58ac75cf.

tests/test_grammar.py
RossPatterson: Fixed by ff01d9650a72605cbdb12c399c13d91c58ac75cf.

RossPatterson:
All outstanding requests have been implemented.  There are conflicts, but I don't know how to fix them in GitHub.

PR #1392, opened 2024-02-14
codespell: move and adjust configuration to pyproject.toml, fix few new typos

yarikoptic:
Best to have configuration(s) in central place so people could just use `codespell` without pre-commit

TODOs

- [ ] remove TMP commit with typo after having verified that pre-commit does catch it. If does -- also remove that added workflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,3 +194,4 @@ Questions about code are best asked on [gitter](https://gitter.im/lark-parser/Lo
 For anything else, I can be reached by email at erezshin at gmail com.
 
  -- [Erez](https://github.com/erezsh)
+Installating
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,23 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,9 @@ repos:
       exclude: '(^tests/.*\.lark|\.svg)$'
 
 - repo: https://github.com/codespell-project/codespell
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -194,3 +194,4 @@ Questions about code are best asked on [gitter](https://gitter.im/lark-parser/Lo
 For anything else, I can be reached by email at erezshin at gmail com.
 
  -- [Erez](https://github.com/erezsh)
+Installating
```
erezsh, line 197: Huh?

yarikoptic: See the TODOs above ;) it was added explicitly in a commit to see if you do pre-commit checking some how (e.g. via github integration or workflow). Apparently you do, but in a Python type check and not a dedicated workflow.

So in principle I can remove dedicated workflow here OR separate our "Python type check" into two -- 2nd one would just run pre-commit and have a clear/different name. WDYT?

erezsh: What would you say are the pros and cons of each approach?

yarikoptic: Having dedicated codespell action makes it clear/easy to find what failed the CI since name of it (codespell) says it all.  When it is bundled in another action (`pre-commit`) - it not immediately clear. If it is bundled somehow in `Python type check` then it becomes even misleading.

I personally would have
- removed codespell action (duplicates what pre-commit does anyways)
- separated pre-commit into a dedicated workflow: it is expected for it to never fail since it is expected that it would be ran by contributors locally first. And even if it fails -- it is "low priority" since typically only formatting etc aspects, and thus even lower than "Type Checking"

yarikoptic: And I am not too familiar with the https://pre-commit.ci/ service  to advice on that

erezsh: Well, one big disadvantage with your proposal is that now instead of only having to run pre-commit, which can run automatically before each commit, contributors would have to run two things, the second one manually. They would only see the error after they open a pull request, instead of seeing it locally when they try to commit. 

@MegaIng What do you think?

yarikoptic: We might be misunderstanding... I added pre-commit for codespell, nothing to be ran manually

yarikoptic: Try to add new popular typo!

yarikoptic:
ended up mentioning this project, so ran back into my PR here, rebased/refreshed. I think I am done with it , consider adopting or closing. Cheers!

erezsh:
Honestly I don't remember what it was about, maybe I misunderstood.

Overall I'm leaning towards merging it. I'm not sure about adding the `tomli` dependency, so I might remove that part.

Anyway, I hope to get to it soon.

Issue #1394, opened 2024-02-25
Transformer raises AttributeError when a tree is only a token


The-Arx:
**Describe the bug**

When a token is the outermost layer of a tree, transformer raises `AttributeError: 'Token' object has no attribute 'children'`.

Here is a simple program causing the bug.
```python
from lark import Lark, Transformer

l = Lark("""?start: INT
            
            %import common.INT
        """)

class MyTransformer(Transformer):
    def INT(self, value):
        return int(value)
  
tree = l.parse("123") # Token('INT', '123')

MyTransformer().transform(tree)

erezsh:
Fixed.

PR #1395, opened 2024-02-25
Fix for when providing a transformer with a Token

erezsh:
Fixes issue #1394

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -158,7 +158,11 @@ class Transformer(_Decoratable, ABC, Generic[_Leaf_T, _Return_T]):
 
     def transform(self, tree: Tree[_Leaf_T]) -> _Return_T:
         "Transform the given tree, and return the final result"
-        return self._transform_tree(tree)
+        res = list(self._transform_children([tree]))
+        if not res:
+            return None     # type: ignore[return-value]
+        assert len(res) == 1
+        return res[0]
 
     def __mul__(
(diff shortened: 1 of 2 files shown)


lark/visitors.py
```diff
@@ -158,7 +158,11 @@ def _transform_tree(self, tree):
 
     def transform(self, tree: Tree[_Leaf_T]) -> _Return_T:
         "Transform the given tree, and return the final result"
-        return self._transform_tree(tree)
+        res = list(self._transform_children([tree]))
```
MegaIng, line 161: this has the side effect of also allowing everything that isn't a tree or token to be passed without error. Not sure if that is good behavior.

erezsh: What's bad about it?

MegaIng: It might hide user error where they pass things that aren't the result of a `Lark.parse` call, which is probably never intentional. But this isn't a strong objection, if you think this is fine, go ahead.

erezsh: I see your point, but I think I'll keep it. My reasoning is that:

1) Lark.parse may return an arbitrary object (through another transformer)

2) It would make the root behave the same as the leaf, in which we allow arbitrary objects and just skip them without an error

3) It's not the kind of mistake that is likely to proceed much without being noticed down the line. (though I admit in some edge cases it might happen)

MegaIng (approved):

PR #1423, opened 2024-06-15
Docs: Added Indenter

erezsh:


MegaIng (changes requested):
I think a slightly more detailed explanation of what the `Indenter` does would be a good idea, especially with regard to parenthesis. I don't think that it is obvious if you don't have python indentation rules exactly in mind.

erezsh:
Good idea, thanks.

erezsh:
True, but they are now on latest: https://lark-parser.readthedocs.io/en/latest/classes.html#indenter

Once we release 1.2.0, I'll update stable too.

Issue #1424, opened 2024-06-17
Is it possible to parse parts of the input?


jjalowie:
**What is your question?**

I'm interested in writing a transformer for certain parts of input programs that wouldn't require a full language grammar. So let's say I have the below sample cpp code:
```
void function(int a, int b) {
  // body
}
```

I want to parse just function signatures and transform them e.g. by adding the '_' suffix to all argument names. I.e. the output for the above input program should be:
```
void function(int _a, int _b) {
  // body (unchanged)
}
```

How could I possibly achieve this using Lark?

MegaIng:
See https://github.com/lark-parser/lark/issues/1371

It's not exactly easy and especially for C-like syntax that is already tricky to parse with lark it might be quite annoying.

jjalowie:
Wow, thanks for a quick response.
I'm thinking about moving on with Earley using the `start: /.*/ rule /.*/` approach (my inputs aren't very big so should be fine performance-wise).
I ran into the following problem though when experimenting around:
```
import lark
grammar = f"""
    start: /.*/ expr /.*/
    expr: "qwerty"
    %import common.WS
    %ignore WS
"""
parser = lark.Lark(grammar, start='start', parser='earley', lexer='dynamic_complete')

ast = parser.parse("""asdf qwerty poiuy""")
print(ast.pretty())
```
This gives the following error:
```
[...]
  File "/home/jjalowie/.bin/mlir_parser/venv/lib/python3.10/site-packages/lark/parser_frontends.py", line 215, in create_earley_parser
    return f(lexer_conf, parser_conf, resolve_ambiguity=resolve_ambiguity,
  File "/home/jjalowie/.bin/mlir_parser/venv/lib/python3.10/site-packages/lark/parser_frontends.py", line 192, in create_earley_parser__dynamic
    earley_matcher = EarleyRegexpMatcher(lexer_conf)
  File "/home/jjalowie/.bin/mlir_parser/venv/lib/python3.10/site-packages/lark/parser_frontends.py", line 178, in __init__
    raise GrammarError("Dynamic Earley doesn't allow zero-width regexps", t)
lark.exceptions.GrammarError: ("Dynamic Earley doesn't allow zero-width regexps", TerminalDef('__ANON_0', '.*'))
```

I'm using lark 1.1.9.

MegaIng:
As it says in the error message, the terminal can't be zero width. Use `/.+/?` instead.

jjalowie:
[...]

Issue #536, opened 2020-03-24
not finding ambiguous parse


christolmicrosoft:
Greetings, and thank you for your excellent package.

I'm not sure if I'm doing something wrong here, but the following CFG gives me an unambiguous parse on the input "a b", when I believe it should be ambiguous

```
%import common.WS
%ignore WS
start: a -> start_0
a: "a" b -> a_0
| "a" "b" -> a_1
b: "b" -> b_0
```

this grammar works though, and provides an ambiguous parse for "a b":

```
%import common.WS
%ignore WS
start: a -> start_0
| a2 -> start_1
a: "a" b -> a_0
a2: "a" "b" -> a_1
b: "b" -> b_0
```

erezsh:
I think you're right, and this is a bug. I'll mark it as such.

I hope you can work around it for now.

I'll also add another version of this bug, which is even more innocent.

```python
from lark import Lark

g1 = """
start: a
a: "a" "b" -> a_0
 | a2
a2: "a" b -> a_1
b: "b"
"""

g2 = """
start: a
a: "a" b -> a_0
 | a2
a2: "a" "b" -> a_1
b: "b"
"""

print(Lark(g1, ambiguity='explicit').parse("ab"))
print(Lark(g2, ambiguity='explicit').parse("ab"))

# Tree(start, [Tree(a_0, [])])
# Tree(start, [Tree(_ambig, [Tree(a_0, [Tree(b, [])]), Tree(a, [Tree(a_1, [])])])])
```

erezsh:
This bug has been fixed in #1427 , and will make it into v1.2  🎉

Only took 4 years.

PR #1427, opened 2024-06-18
BUGFIX Earley: Now yielding a previously repressed ambiguity

erezsh:
Contains 3 commits:

1) Test `test_multiple_start_solutions`, that fails before the fix

2) Fixes Earley. Previously, slightly different items were considered equal.

3) A fix to the test_cycle2, which I believe was previously wrong

Issue #1424 was the first lead to finding this bug. But seems that it isn't solved by this change alone.

@chanicpanic Can you please verify my work?

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -15,7 +15,7 @@ from collections import deque
 from ..lexer import Token
 from ..tree import Tree
 from ..exceptions import UnexpectedEOF, UnexpectedToken
-from ..utils import logger, OrderedSet
+from ..utils import logger, OrderedSet, dedup_list
 from .grammar_analysis import GrammarAnalyzer
 from ..grammar import NonTerminal
 from .earley_common import Item
@@ -282,7 +282,7 @@ class Parser:
         # If the parse was successful, the start
         # symbol should have been completed in the last step of the Earley cycle, and will be in
         # this column. Find the item for the start_symbol, which is the root of the SPPF tree.
-        solutions = list(OrderedSet([n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0]))
+        solutions = dedup_list(n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0)
         if not solutions:
             expected_terminals = [t.expect.name for t in to_scan]
             raise UnexpectedEOF(expected_terminals, state=frozenset(i.s for i in to_scan))
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -294,7 +294,8 @@ class Parser:
             except ImportError:
                 logger.warning("Cannot find dependency 'pydot', will not generate sppf debug image")
             else:
-                debug_walker.visit(solutions[0], "sppf.png")
+                for i, s in enumerate(solutions):
+                    debug_walker.visit(s, f"sppf{i}.png")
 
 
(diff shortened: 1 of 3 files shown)


erezsh:
P.S. It's easier to see why there should be ambiguity in test_cycle2, if we rename `_recurse` to `recurse`. The second recurse is now taken, and was previously ignored. (i.e. it's a new path, not an extra cycle)

lark/parsers/earley.py
```diff
@@ -296,13 +297,13 @@ def parse(self, lexer, start):
                 debug_walker.visit(solutions[0], "sppf.png")
```
chanicpanic, line 297: Optional: create debug images for all solutions

lark/parsers/earley.py
```diff
@@ -296,13 +297,13 @@ def parse(self, lexer, start):
...
-        if len(solutions) > 1:
-            assert False, 'Earley should not generate multiple start symbol items!'
-
         if self.Tree is not None:
             # Perform our SPPF -> AST conversion
             transformer = ForestToParseTree(self.Tree, self.callbacks, self.forest_sum_visitor and self.forest_sum_visitor(), self.resolve_ambiguity)
-            return transformer.transform(solutions[0])
+            solutions = [transformer.transform(s) for s in solutions]
+
+            return solutions[0] if len(solutions) == 1 else self.Tree('_ambig', solutions)
```
chanicpanic, line 305: For multiple solutions, it is possible for one or more of the solutions to be `_ambig` nodes themselves, in which case their children could be inlined as children of the `_ambig` node being created here.

This is probably just a matter of tree aesthetics.

chanicpanic (approved):
LGTM with a couple optional comments.

While this doesn't fully solve the issue from #1424, it does appear to fix all the examples in #536.

lark/parsers/earley.py
erezsh: Done.

lark/parsers/earley.py
erezsh: Done. 

Might be a bit slow if there are many solutions, but we deal with that when it happens.

erezsh:
@chanicpanic  Like that?

chanicpanic:
Yup, looks good

erezsh:
@MegaIng Any comments? Or should I just merge it?

lark/parsers/earley.py
```diff
@@ -281,7 +282,7 @@ def parse(self, lexer, start):
         # If the parse was successful, the start
         # symbol should have been completed in the last step of the Earley cycle, and will be in
         # this column. Find the item for the start_symbol, which is the root of the SPPF tree.
-        solutions = [n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0]
+        solutions = list(OrderedSet([n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0]))
```
MegaIng, line 285: We have `lark.utils.dedup_list` which does this exact operation, but it should be slightly faster (and more explicit).

MegaIng (approved):
I can't say much on the actual implementation changes since I don't know the earley algorithm well enough, but looks good in general.

lark/parsers/earley.py
erezsh: Good catch, thanks!

erezsh:
Thanks @chanicpanic  and @MegaIng  !

PR #1428, opened 2024-06-18
Clean up test_parser.py, use xFail instead of skip where appropriate

MegaIng:
This is something I did while working on the `Lark.scan` implementation. I wanted to use `xfail` and decided that it should be quite useful for many other tests as well. I also deleted a few old comments/completely skipped tests suggesting features that will probably never be implemented.

I adjusted a few of the conditions based on the actual behavior of the code, for example `cyk` does actually support the tree less `transformer` argument. It doesn't have quite as big of a benefit, but it does work.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -23,10 +23,6 @@ try:
 except ImportError:
     regex = None
 
-try:
-    import pytest
-except ImportError:
-    pytest = None
 
 import lark
 from lark import logger
@@ -931,9 +927,6 @@ class CustomLexerNew(Lexer):
     def lex(self, lexer_state, parser_state):
         return self.lexer.lex(lexer_state, parser_state)
 
-    def make_lexer_state(self, text):
-        return self.lexer.make_lexer_state(text)
-
     __future_interface__ = True
 
 class CustomLexerOld(Lexer):
@@ -1583,7 +1576,6 @@ def _make_parser_test(LEXER, PARSER):
             g = """!start: "a"i "a"
                 """
             l = _Lark(g)
-            print(l.terminals)
             self.assertEqual(l.parse('aa').children, ['a', 'a'])
             self.assertEqual(l.parse('Aa').children, ['A', 'a'])
             self.assertRaises(UnexpectedInput, l.parse, 'aA')
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -11,11 +11,6 @@ from lark.utils import isascii
 
 from lark import Token, Transformer_NonRecursive, LexError
 
-try:
-    from cStringIO import StringIO as cStringIO
-except ImportError:
-    # Available only in Python 2.x, 3.x only has io.StringIO from below
-    cStringIO = None
 from io import (
         StringIO as uStringIO,
         BytesIO,
@@ -28,6 +23,11 @@ try:
 except ImportError:
     regex = None
 
+try:
+    import pytest
+except ImportError:
+    pytest = None
+
 import lark
 from lark import logger
 from lark.lark import Lark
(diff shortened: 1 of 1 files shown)


erezsh:
Why change the skip to xfail? xfail implies the test should be fixed in the future, i.e. failing for now but should work. (I'm sure you know pytest supports skip too)


Also, I'm wondering if this PR won't cause too many collisions with the lark.lark PR..

MegaIng:
> xfail implies the test should be fixed in the future, i.e. failing for now but should work

I somewhat disagree: While this is a possible meaning, I would in this case interpret it as these should fail, and if they don't the tests aren't actually testing for what they should, without any promises about the future. E.g. a test case that is designed to specifically need the `dynamic` lexer should not function without it. But if you are strongly against this interpretation, that's fine by me.

> Also, I'm wondering if this PR won't cause too many collisions with the lark.lark PR..

I don't think so? That PR hasn't touched `test_parser.py` at all.

erezsh:
Oh right, sorry my bad.

As for xfail, I don't think it means that the test won't function, but rather that the test is correct but the source code isn't. So for example xfailing ambiguity on lalr would mean that we expect lalr to support ambiguity, and it's temporarily broken.

If it's not a bother, could you please change xfail into skip?

MegaIng:
> As for xfail, I don't think it means that the test won't function, but rather that the test is correct but the source code isn't. So for example xfailing ambiguity on lalr would mean that we expect lalr to support ambiguity, and it's temporarily broken.

This is *a* possible interpretation, and probably the more common one. But I don't think the feature itself is all too prescriptive in this regard.

As a concrete example, consider the tests that use the tree less `transform` parameter. These only work for `lalr` (and `cyk`) and should produce an error message for `earley`, meaning that the tests using this should not pass, i.e. they are expected to fail. We currently have no plans to ever implement this feature (AFAIK), but it's still an ok usecase of `xfail` IMO. If you think this is to confusing, I will revert that part of this PR.

erezsh:
Yes, using Earley in that case should produce an error, but usually that would be caught in an AssertRaises and not with the xfail mechanism.

But actually, if you feel strongly about it that's fine by me. I usually just use xfail as a loose TODO marking, so not much loss there.

MegaIng:
> Edit: I take it back. I just ran it in the console and got a huge wall of xfail warnings.. Let's just use skip, okay?

Sure, did that. Now it's a huge wall of skip warnings instead.

tests/test_parser.py
```diff
@@ -946,6 +931,9 @@ def __init__(self, lexer_conf):
     def lex(self, lexer_state, parser_state):
         return self.lexer.lex(lexer_state, parser_state)
 
+    def make_lexer_state(self, text):
```
erezsh, line 934: This doesn't seem necessary

tests/test_parser.py
```diff
@@ -1666,6 +1577,26 @@ def test_token_flags(self):
             tree = l.parse('AB,a')
             self.assertEqual(tree.children, ['AB'])
 
+        @unittest.skipIf(LEXER in ('basic', 'custom_old', 'custom_new'), "Requires context sensitive terminal selection")
+        def test_token_flags_collision(self):
+
+            g = """!start: "a"i "a"
+                """
+            l = _Lark(g)
+            print(l.terminals)
```
erezsh, line 1586: Remove the print

tests/test_parser.py
```diff
@@ -28,6 +23,11 @@
 except ImportError:
     regex = None
 
+try:
```
erezsh, line 26: No longer needed

erezsh (changes requested):
Thanks!

For whatever reason, the wall of skips is 3 times smaller than the wall of xfails was.

I also think to make it even less, by replacing a few of the skips with a simple if stmt. But I can do that later, after this has been merged.

erezsh (approved):

PR #1430, opened 2024-06-20
Update config and drop python < 3.8

MegaIng:
This PR consists of 2 parts:

It removes python code specific to python 3.6/3.7 (and some only relevant for 2.7 that we forgot!)

It updates quite a bit config, although it's little "real" changes. I list them here as checkboxes so that you @erezsh can check them off if you agree with them

- [ ] Drop 3.6 and 3.7 from all relevant locations. Note that there is now the sentence `Since version 1.2, only Python versions 3.8 and up are supported`, so we should increment the version when this is released.
- [ ] Add `3.13-dev` to the tests workflow
- [ ] Change `pypy-3.7` to `pypy-3.10` in the tests workflow.
- [ ] Move `pytest.ini` into `pyproject.toml`, with the goal of reducing the number of config files we have. I didn't do the same for `tox.ini` because it would just be putting the raw `.ini` file into a multiline string, which seems too weird for me.
- [ ] Added `unused-ignore` to the error codes for mypy which tells us when we can drop an `ignore` (I also dropped a few ignores based on this warning)
- [ ] Remove the pin for the `types-regex` version. This is no longer necessary with us dropping python <3.8
- [ ] Pin `mypy` to `1.10` instead of `0.950`. Pinning is probably a good idea in general, but everything works with the current version, so no reason I know to keep an older version.
- [ ] Change `fail_ci_if_error` in `codecov.yml` from `true` to `false`. This is primarily a selfish change, since it's very annoying to always get failed CI on my repo since I don't have the secret set. If you don't like this version, there is also a way to conditionally execute the workflow which would be an alternative solution.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -35,6 +35,6 @@ jobs:
         flags: unittests
         env_vars: OS,PYTHON
         name: codecov-umbrella
-        fail_ci_if_error: true
+        fail_ci_if_error: false
         path_to_write_report: ./coverage/codecov_report.txt
         verbose: true
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,11 +3,10 @@ on: [push, pull_request]
 
 jobs:
   build:
-    # runs-on: ubuntu-latest
-    runs-on: ubuntu-20.04   # See https://github.com/actions/setup-python/issues/544
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "pypy-3.7"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev", "pypy-3.10"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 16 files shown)


erezsh:
Looks good.

As for codecov, ideally I would like it to work for everyone. I have the `CODECOV_TOKEN` env var set for the repo. Do you have any idea why it doesn't work for you?

Also, I think the version bump commit should be part of this PR. Feel free to do it, or I can create and push it here if you prefer.

MegaIng:
> I have the CODECOV_TOKEN env var set for the repo. Do you have any idea why it doesn't work for you?

Because quite reasonably, it's a secret so that it's not publicly accessible, and to specifically prevent everyone from pushing stuff to codecov.

erezsh:
Well, I'd still like to fix codecov for all pull requests, but I suppose it doesn't have to be in this PR.

So, is it ready for merge?

MegaIng:
Ok, looks like the `codecov` action is currently a bit buggy: https://github.com/codecov/codecov-action/issues I am going to revert these commits and we can try again in a few weeks.

MegaIng:
Now we are ready to merge.

PR #1432, opened 2024-06-25
Fix SymbolNode.end for completed tokens

### LINKED ISSUE #1431: Dynamic Earley: Incorrect value for SymbolNode.end
Consider the following:

```python
grammar = r"""
start: "ABC"
"""

print(Lark(grammar, ambiguity="forest").parse("ABC"))
```

Output: `(start, 0, 2, -inf)`

The end index is 2 which suggests that the substring "AB" was parsed. However, we know that the full string was parsed, so the end index should be 3.

Note that the issue does not occur if we use a rule instead:

```python
grammar = r"""
start: abc 
abc: "ABC"
"""

print(Lark(grammar, ambiguity="forest").parse("ABC"))
```

Output: `(start, 0, 3, -inf)`

I will open a PR with a fix.

chanicpanic:
Fixes #1431.

I realized that that bug helped #1427 work: multiple solutions had different end values and were considered distinct. After fixing the bug, the end values of solutions are the same, and the solutions get deduped being considered equal. However, we know that the solutions have different children and should be considered distinct. For simplicity, I updated `SymbolNode` to use the default identity-based equality. Let me know if you have any other ideas for its `__eq__` and `__hash__` implementations.

### DIFF
--- a/lark/parsers/earley_forest.py
+++ b/lark/parsers/earley_forest.py
@@ -38,15 +38,15 @@ class SymbolNode(ForestNode):
 
     Parameters:
         s: A Symbol, or a tuple of (rule, ptr) for an intermediate node.
-        start: The index of the start of the substring matched by this symbol (inclusive).
-        end: The index of the end of the substring matched by this symbol (exclusive).
+        start: For dynamic lexers, the index of the start of the substring matched by this symbol (inclusive).
+        end: For dynamic lexers, the index of the end of the substring matched by this symbol (exclusive).
 
     Properties:
         is_intermediate: True if this node is an intermediate node.
         priority: The priority of the node's symbol.
     """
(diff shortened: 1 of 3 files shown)


erezsh:
LGTM, but my understanding of this code is limited. What do you think are the chances that this will create duplicate solutions?

chanicpanic:
I think the chances are low. I would hope that if duplicate solutions could be created, it would have happened for some of the existing tests causing them to fail.

From my (imperfect) understanding of the code, I believe that there cannot be more than two different solution nodes at the end:

1. One for completed start rule alternatives that ended in a non-terminal
2. One for completed start rule alternatives that ended in a terminal

Within those two groups, ambiguities are handled as usual and `SymbolNode` creation is limited by the `node_cache`s.

erezsh:
Thanks!

PR #1433, opened 2024-06-28
Disable ForestToParseTree cache when ambiguity='resolve' 

### LINKED ISSUE #1283: Earley parser produces wrong parse Tree
Over on SO someone asked this question: https://stackoverflow.com/questions/76366280/parsing-formulas-using-lark-ebnf/76381256

As far as I can tell, it shows a bug in the earley parser where it duplicates some Tokens because of enormous amounts of ambiguities. I don't have the expertise with earley to figure out what is happening or the time to create a minimal example right now.

chanicpanic:
Fixes #1283.

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -300,7 +300,7 @@ class Parser:
 
         if self.Tree is not None:
             # Perform our SPPF -> AST conversion
-            transformer = ForestToParseTree(self.Tree, self.callbacks, self.forest_sum_visitor and self.forest_sum_visitor(), self.resolve_ambiguity)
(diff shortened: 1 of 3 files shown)


erezsh:
Lgtm.

I think maybe it's a good idea to add a comment explaining that caching is disabled because of the tree construction, or mention the issue number? Because it's not an obvious connection imho.

chanicpanic:
> I think maybe it's a good idea to add a comment explaining that caching is disabled because of the tree construction, or mention the issue number? Because it's not an obvious connection imho.

Good idea, added.

erezsh:
Thanks!

PR #1435, opened 2024-06-30
Bugfix for issue #1434

### LINKED ISSUE #1434: Inconsistent parse results from simple ambiguous grammar
**Describe the bug**

I have a small grammar that is a starting point to parse regular expressions. When run repeatedly on the same input Lark produces two different parse trees.

There is an ambiguity in my toy grammar because a RE `quantifier` (`*`, `+`, `?`) can also be interpreted as a `printable_char` (ASCII values between space and tilde). However my reading previous issues (#81, #83) indicates inconsistent parse results should be treated as a bug (and it was certainly a surprise).

**To Reproduce**

Lark 1.1.9, Python 3.10.12.

```python
#!/usr/bin/env python

from lark import Lark

parser = Lark('''
start: factor+

?factor: atom quantifier?

quantifier: "*" | "+" | "?"

?atom: DOT | printable_char

printable_char: /[ -~]/

DOT: "."
''')

regex = r'a.?'
tree = parser.parse(regex)
print(tree.pretty())
```

**Sample results**

```
start
  printable_char	a
  .
  printable_char	?
```

```
start
  printable_char	a
  factor
    .
    quantifier
```

erezsh:
Fixes #1434

### DIFF
--- a/lark/parsers/grammar_analysis.py
+++ b/lark/parsers/grammar_analysis.py
@@ -3,7 +3,7 @@
 from collections import Counter, defaultdict
 from typing import List, Dict, Iterator, FrozenSet, Set
 
-from ..utils import bfs, fzset, classify
+from ..utils import bfs, fzset, classify, OrderedSet
 from ..exceptions import GrammarError
 from ..grammar import Rule, Terminal, NonTerminal, Symbol
 from ..common import ParserConf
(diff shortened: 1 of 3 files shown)


erezsh:
Not much to review but I'll follow protocol just in case.

MegaIng:
If  this is a bugfix, we should have tests that fail (sometimes) beforehand.

Although, for that we ideally had infrastructure to start a new process with different, controllable hash seeds which then execute the snippet we want to check out.

erezsh:
Added. Anything else?

MegaIng (approved):

PR #1446, opened 2024-08-14
Initial support for import * (WIP)

erezsh:


erezsh:
@v-ramesh I only abandoned it for lack of public interest.

If someone wants to take this PR and re-submit it with some tests, they are welcome to do it!

PR #1447, opened 2024-08-15
Bugfix Earley: only transform the solutions we yield

erezsh:
Fixes an issue overlooked by PR #1444 

Addresses performance issues mentioned in #1443

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -282,10 +282,15 @@ class Parser:
         # If the parse was successful, the start
         # symbol should have been completed in the last step of the Earley cycle, and will be in
         # this column. Find the item for the start_symbol, which is the root of the SPPF tree.
-        solutions = dedup_list(n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0)
+        solutions = [n.node for n in columns[-1] if n.is_complete and n.node is not None and n.s == start_symbol and n.start == 0]
         if not solutions:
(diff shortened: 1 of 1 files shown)


MegaIng (approved):
I don't have full understanding of this code, but this looks reasonable.

The deleted comment refers to `ambiguity="forest"`, right? Are multiple start solutions implicitly encoded in there? If not, is there a good way to expose them without breaking API?

I think leaving a small doc note somewhere would be a good idea either way. (e.g. "`amibuity="forset"` only returns the root for the first found solution")

erezsh:
Yes, that's a good point. It should return several solutions.

I think the correct thing is to break the API, since the current behavior is broken. But that would break compatibility.

Another option is to distinguish between ambiguity="forest" and ="forests" .. but that sounds a bit silly.

Another option is to return a new SymbolNode that has the solution as its children. That way we don't introduce a new type (or a python list), but then it's not clear to me what its arguments should be (meaning s, start, end), or if it would constitute a valid "forest" object.

Or we can just leave it as-is, because no one has complained so far.

@chanicpanic Maybe you have an opinion about this?

chanicpanic:
I agree that an array is correct given the current behavior. Although, I wouldn't make the change just yet because I believe the earley implementation can be adjusted so that we don't end up with more than one distinct root in the first place. Sharing the `node_cache` between `scan` and `predict_and_complete` seems to do the trick, albeit with a few test failures to look into.

erezsh:
No longer relevant.

Issue #1457, opened 2024-08-29
v_args seems to be ignored for subclass of Visitor class


skepppy:
**Describe the bug**

I want to write a `Visitor` class that uses `inline=True`, but Lark seems to ignore it and I do not understand why.

The documentation writes:

> v_args can modify this behavior. When used on a transformer/visitor class definition, it applies to all the callback methods inside it.

Furthermore, it writes:

> inline (bool, optional) – Children are provided as *args instead of a list argument (not recommended for very long lists).

With this, I would expect subclasses of the `Visitor` class to receive parsed tokens from the abstract syntax tree (AST) through its parameters. However, I am still receiving a single parameter with type `lark.Tree`.

Maybe it's something else that is wrong and causes , if so, please enlighten me.

**To Reproduce**

I tried to create a relatively simple script that shows the problem I'm encountering. It is a script that creates both a `Transformer` and a `Visitor` class, both decorated with `v_args(inline=True)`, but the `Transformer` class actually receiving the words through its parameters, while the `Visitor` class still receives the `lark.Tree` class. According to the docs, it says that it should work for both classes, but the `Visitor` class seems to ignore it.

```python
from lark import Lark, Transformer, Tree, Visitor
from lark.visitors import v_args

LARK_GRAMMAR = r"""
start: numbers
numbers: NUMBER [WS_INLINE numbers]

%import common.WS_INLINE
%import common.NUMBER
%import common.LETTER
"""


@v_args(inline=True)
class TransformNumbers(Transformer):
    def numbers(self, tree):
        assert isinstance(tree, Tree)  # This fails, like I would expect it to.


@v_args(inline=True)
class VisitNumbers(Visitor):
    def numbers(self, tree):
        assert isinstance(tree, Tree)  # This should fail, because inline=True.


number_text = "123 456"

number_parser = Lark(LARK_GRAMMAR)

tree = number_parser.parse(number_text)

VisitNumbers().visit(tree)
TransformNumbers().transform(tree)
```

Furthermore, I'm running the following:
```bash
$ pip list | grep lark
lark                      1.2.2
$ python --version
Python 3.10.12
```
If I run the above script, I get an exception, but only for the `TransformNumbers`, not for `VisitNumbers`:

```bash
$ python lark-visitor.py 
Traceback (most recent call last):
  File "/home/user/.local/share/virtualenvs/test-visitor/lib/python3.10/site-packages/lark/visitors.py", line 122, in _call_userfunc
[...]

PR #1458, opened 2024-08-29
Rephrase v_args() documentation to underline it is related to Transformer classes only

skepppy:
Related to issue #1457, rewording the documentation of `v_args()` to specify it is only applicable to `Transformer` classes.


erezsh:
I think we should avoid using the word visitor at all, when the Visitor class isn't involved, because it is confusing. I think it's best to just refer to them as callbacks, or transformer callbacks.

skepppy:
I reworded it again, removing the references to visitor.

erezsh:
Thanks!

erezsh:
Yes, that sounds like the right thing to do.

Issue #1311, opened 2023-07-25
TypeError: Parser.parse() got an unexpected keyword argument 'on_error'


aspizu:
```
ERROR:root:parser raised exception
Traceback (most recent call last):
  File "/home/aspizu/Projects/goboscript-lsp/goboscript-lsp/diag.py", line 32, in get_diagnostics
    parser.parse(doctext, on_error=on_error)
  File "/usr/lib/python3.11/site-packages/lark/lark.py", line 645, in parse
    return self.parser.parse(text, start=start, on_error=on_error)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/lark/parser_frontends.py", line 96, in parse
    return self.parser.parse(stream, chosen_start, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Parser.parse() got an unexpected keyword argument 'on_error'
```

MegaIng:
`on_error` is only supported for `parser='lalr'`. Can you provide an actual reproducer script please?

aspizu:
yes was using earley

aspizu:
another suggestion is that instead of a string definiing what parser to use
just impor the parser class and pass it to Lark  so you can use typing to convey that earley doesnt support on_error

erezsh:
I agree, a better error is warranted.

If we want to be type-friendly, we can define different interfaces for each parser, and use `@overload` to return the correct one based on the parser name (using Literal etc.).

erezsh:
Error is now fixed.

If anyone wants to suggest a PR to improve the API or make it more type-friendly, they are welcome to do so.

PR #1460, opened 2024-09-01
Better error in Lark.parse when using on_error when parser!=lalr (issue #1311)

erezsh:
Not much to review, but I'll still leave it up for comments for a while


MegaIng (approved):
Yeah, not much to review. LGTM

PR #1467, opened 2024-09-11
Add Tree.find_token() method

### LINKED ISSUE #1466: Syntactic sugar: Tree.find_token()
**Suggestion**
Add new method `Tree.find_type(typ: str) -> Iterator[Token]` as a companion for `Tree.find_data()`:
```python
for term in tree.find_type('TERM'):
    ...
```

Right now this can be achieved by
```python
for term in tree.find_pred(lambda node: isinstance(node, Token) and node.type == 'TERM'):
    ...
```
which doesn't fit Black's line width and will look even uglier in real cases.

**Additional context**
I can provide PR if you find this enhancement reasonable.

The method proposed seems to be a logical extension of existing API:
```
Tree => find_data()
Token => find_type()
```

makukha:
Resolves #1466.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 *.pyo
 /.tox
+/lark.egg-info/**
 /lark_parser.egg-info/**
 tags
 .vscode
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -171,6 +171,10 @@ class Tree(Generic[_Leaf_T]):
 
 ###}
 
+    def find_token(self, typ: str) -> Iterator[_Leaf_T]:
+        """Returns all tokens whose type equals the given typ."""
+        return self.scan_values(lambda v: isinstance(v, Token) and v.type == typ)
+
     def expand_kids_by_data(self, *data_values):
         """Expand (inline) children with any of the given data values. Returns True if anything changed"""
         changed = False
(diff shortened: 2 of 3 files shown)


makukha:
@erezsh ready for review.

erezsh:
Overall looks good. A few requests:

- Add a test
- Rename typ to token_type
- Clarify in the docstring that it is a recursive function, i.e. it will find tokens in all the subtrees.

makukha:
Instead of `isinstance(v, Token)` I switched to `not isinstance(v, Tree)` similar to `_pretty()` to avoid importing `Token`.

erezsh:
Sorry it tooks me this long to reply! It slipped my memory.

Why did you change `isinstance(v, Token)` to `not isinstance(v, Tree)`?

erezsh:
The logic in pretty is a little different.

P.S. if you don't have time right now, let me know, and I'll make the change for you.

makukha:
Thank you so much, @erezsh! This change is pretty simple, added it to PR.

MegaIng:
Looks good, however I would also like to see a test for what happens if there is a leaf node that isn't a `Token` instance. (i.e. it doesn't get listed).

makukha:
@MegaIng Thank you for the hint. I added a non-token item at the end of the tree. As expected, it didn't affect the result.

MegaIng (approved):

erezsh:
Thanks!

Issue #1480, opened 2024-10-25
python 3.14.0a1 - test_partial and test_smart_decorator are failing 


hrnciar:
**Describe the bug**

Fedora is starting to rebuild packages with upcoming python 3.14. Some tests are failing.

~~~~
======================================================================
ERROR: test_partial (tests.test_trees.TestTrees.test_partial)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 122, in _call_userfunc
    return f.visit_wrapper(f, tree.data, children, tree.meta)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 504, in _vargs_inline
    return f(*children)
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 482, in __call__
    return self.base_func(*args, **kwargs)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: TestTrees.test_partial.<locals>.test() got multiple values for argument 'postfix'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/tests/test_trees.py", line 361, in test_partial
    res = T().transform(tree)
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 161, in transform
    res = list(self._transform_children([tree]))
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 146, in _transform_children
    res = self._transform_tree(c)
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 156, in _transform_tree
    children = list(self._transform_children(tree.children))
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 146, in _transform_children
    res = self._transform_tree(c)
  File "/builddir/build/BUILD/python-lark-1.2.2-build/lark-1.2.2/lark/visitors.py", line 157, in _transform_tree
    return self._call_userfunc(tree, children)
[...]

PR #1483, opened 2024-11-06
Wrap functools.partial in staticmethod() to add compatibility with Python 3.14

hrnciar:
Python 3.14

Fixes: #1480


erezsh:
Thanks!

Issue #408, opened 2019-07-21
Transformer (non-inline) and v_args(tree=True) does nothing


Qix-:
Using `@v_args(tree=True)` on a subclass of `Transformer` doesn't have any effect. No errors, nothing - it simply doesn't change any behavior (the children are still being passed as normal).

This is also the case with `@v_args(meta=True)` though this incantation causes an error to be raised - something about not being able to pass meta information to internal visitors (very cryptic error).

How do you get a non-inline `Transformer` subclass to receive `Tree` or `Meta` information?

erezsh:
Your title mentions `Transformer`, but you say nothing about it in the text.

`Visitor` passes the tree by default, which lets you access everything, including the meta.

erezsh:
Ok, nice ninja edit. Please paste the exact code you're using, that shows that it doesn't work.

Qix-:
Yeah, sorry about that - caught it right after I posted.

Here's a repro case. This is using `lark-parser` from pypi:

```python
import sys

from lark import Lark, Transformer, v_args

GRAMMAR = """
?start: foo
foo: BAR?
BAR: "bar"
"""

@v_args(meta=True)
class FooTransformer(Transformer):
    def foo(self, t, m=None):
        print(t, file=sys.stderr)
        print(m, file=sys.stderr)

parser = Lark(
    GRAMMAR,
    parser="lalr",
    transformer=FooTransformer(),
    propagate_positions=True,
)

print(parser.parse("bar"))
```

For `meta=True` I get

```
Traceback (most recent call last):
  File "test.py", line 21, in <module>
    propagate_positions=True,
  File "/private/tmp/test-lark-bug/env3/lib/python3.7/site-packages/lark/lark.py", line 221, in __init__
    self.parser = self._build_parser()
  File "/private/tmp/test-lark-bug/env3/lib/python3.7/site-packages/lark/lark.py", line 239, in _build_parser
    self._prepare_callbacks()
  File "/private/tmp/test-lark-bug/env3/lib/python3.7/site-packages/lark/lark.py", line 236, in _prepare_callbacks
    self._callbacks = self._parse_tree_builder.create_callback(self.options.transformer)
  File "/private/tmp/test-lark-bug/env3/lib/python3.7/site-packages/lark/parse_tree_builder.py", line 230, in create_callback
    assert not getattr(f, 'meta', False), "Meta args not supported for internal transformer"
AssertionError: Meta args not supported for internal transformer
```

[...]

PR #1486, opened 2024-11-12
Mention internal/external transformers in the documentation

odanoburu:
Helps users not fall into a problem similar to #408 — which I recently did.

I'm sure the actual text could be improved; feel free to modify it to taste, but I do think it would be helpful to have this or something similar in the docs. Thanks for the library!

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -523,7 +523,7 @@ def v_args(inline: bool = False, meta: bool = False, tree: bool = False, wrapper
 
     Parameters:
         inline (bool, optional): Children are provided as ``*args`` instead of a list argument (not recommended for very long lists).
-        meta (bool, optional): Provides two arguments: ``meta`` and ``children`` (instead of just the latter); ``meta`` is only available for external transformers (i.e., those not supplied as ``Lark``'s ``transformer`` parameter).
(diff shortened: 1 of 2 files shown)


lark/visitors.py
```diff
@@ -523,7 +523,7 @@ def v_args(inline: bool = False, meta: bool = False, tree: bool = False, wrapper
 
     Parameters:
         inline (bool, optional): Children are provided as ``*args`` instead of a list argument (not recommended for very long lists).
-        meta (bool, optional): Provides two arguments: ``meta`` and ``children`` (instead of just the latter)
+        meta (bool, optional): Provides two arguments: ``meta`` and ``children`` (instead of just the latter); ``meta`` is only available for external transformers (i.e., those not supplied as ``Lark``'s ``transformer`` parameter).
```
erezsh, line 526: What if we phrase it as 

```
``meta`` isn't available for transformers supplied to Lark using the ``transformer`` parameter (aka internal transformers).
```

odanoburu: Certainly sounds better/is more direct. I've changed it!

erezsh:
Thanks for contributing!

Issue #1463, opened 2024-09-04
Lark resulting LALR -nor- Earley parser will pick out the longest match of say "a + b + c + ... + d" example input.  Bug (?)


FruitfulApproach:
**Describe the bug**

Normally when you create a calculator you work merely with binary operations and everything is parenthesized internally like so (unless they transform it like I think I will be doing to solve this):
a + (b + c)

Now, Lark is thinking this is what we always want.  When I try to match a finite number of additions, greater than two:

```
from lark import Lark, Transformer

KATEX_SUBSET_GRAMMAR = r"""
start : (RAW_STR | katex_formula)*
katex_formula : DDOLR content DDOLR | DOLR content DOLR
content : addition | variable | int_const
addition : content ("+" content)+
variable : text_var | atomic_var
atomic_var : GREEK | LATIN
text_var : TEXT_CMD "{" var_name "}"
var_name : (NAME | "-")+
int_const : INT
LATIN : /[a-zA-Z]/
GREEK : /\\alpha|\\beta|\\gamma|\\delta/
RAW_STR : /[^\$]+/
TEXT_CMD : /\\text|\\textbf/
DDOLR : /\$\$/
DOLR : /\$/
%import common.INT
%import python.NAME
%import common.WS
%ignore WS
"""

katex_parser = Lark(grammar=KATEX_SUBSET_GRAMMAR, parser='earley')

TEST_ENUM = False

if not __debug__ or TEST_ENUM:
    Op = 'o'
    Data = '@'
    Var, Concat, KatexBlock, Katex, Add = range(5)
else:
    Op = 'Op'; Data = 'Data'
    Var = 'Var';  Concat = 'Concat'; Add = 'Add'
    KatexBlock = '$$'; Katex = '$'
    
wrapping_ops = { Katex, KatexBlock }
infix_ops = { Concat }
        
class KaTeXtoJson(Transformer):     
    def start(self, tree):
        if len(tree) > 1:
            return self._concat(tree)
        return tree[0]
    
    def _concat(self, tree):
        return [Concat, tree]
    
    def variable(self, tree):
        print(tree)
        return tree
    
    def atomic_var(self, tree):
        print(tree)
        return tree
    
    def RAW_STR(self, tree):
        return tree.strip()
    
    def int_const(self, tree):
        return int(tree.data[0])
    
    def katex_formula(self, tree):
        return ''.join(tree.data)
    
    def atomic_var(self, tree):
        return tree[0]
    
[...]

Issue #370, opened 2019-04-25
Possible bug in 'standard' lexer: the longest token is matched incorrectly


kodo-pp:
When I use the `standard` lexer (default for `lalr` parser) sometimes it determines the longest match incorrectly.

Consider the following code:
```python
import lark

grammar = r'''
start: BINOP | ASSIGNMENT_OP

ASSIGNMENT_OP: "="
    | "+="

BINOP: "*"
    | "/"
    | "%" 
    | "+"
    | "-"
    | "long_operator"

%import common.WS
%ignore WS
'''

parser = lark.Lark(grammar, parser='lalr')
code = '+='
print(list(parser.lex(code)))
```

It should print `[Token(ASSIGNMENT_OP, '+=')]` because the longest match is `+=` and the [documentation](https://lark-parser.readthedocs.io/en/latest/grammar/#notes-for-when-using-a-lexer) says that in this case the longest match should be returned. However, the code prints `[Token(BINOP, '+'), Token(ASSIGNMENT_OP, '=')]`, which is incorrect.

I suppose the lexer might be comparing not the lengths of actually matching terminals (`+` and `+=` in this case) but the lengths of the longest possible options (`long_operator` and `+=`)

P.S. lark version is 0.7.0, installed as a package `python-lark-parser` in ArchLinux

erezsh:
> I suppose the lexer might be comparing not the lengths of actually matching terminals (+ and += in this case) but the lengths of the longest possible options (long_operator and +=)

Exactly. The only real fix would be to implement regexp analysis, which isn't worth the trouble.

kodo-pp:
Then I have two questuons:

1. Why does then the documentation says that the lexer performs in the "correct" way when in fact it isn't true?

2. Won't the following be the fix? For each option we compute the length of the longest matching token (in the case I described `long_operator` would have zero-length match while `+` would have a match with length 1) and then use the longest token

whitten:
kodo-pp said; Then I have two questions:
1.Why does the documentation say that the lexer performs in the "correct" way when in fact it isn't true?
As you wrote your grammar, "+=" and "="  are each one token long. Perhaps if you separated the "+=" into "+" "=" the lexer would see it is two separate tokens?
LALR loves to compile things into a table and the grammar rules are fully distinct from the lexer rules.  I expect that forcing them into separate tokens will allow the parts of "+=" to be seen by the grammar rules.

[...]

PR #1490, opened 2024-11-17
Longest match lexer option to mimic the Unix tool Lex

AiStudent:
Hi. I added a lexer for the lalr parser as a complement, which mimics the behavior of the Unix tool [Lex](https://man7.org/linux/man-pages/man1/lex.1p.html) and libraries such as [flex](https://westes.github.io/flex/manual/Matching.html#Matching). The behavior in question is to: match with the longest match found; if there are multiple longest matches the precedence of the terminals is in the order that they are defined.

This also means that the terminals are **not** sorted according to the priority rules defined in grammar.md:
1. Highest priority first (priority is specified as: TERM.number: ...)
2. Length of match (for regexps, the longest _theoretical_ match is used)
3. Length of literal / pattern definition
4. Name

By not relying on the above rules for precedence, it is possible to use grammars defined for Lex and its derivations in Lark.

Below follows an example using `longest_match`:

```
from lark import Lark

parser = Lark(r"""
        start: AB -> ab
          | AC -> ac
        
        AB: /a(b)*/
        AC: /a(c)*/
        
        """, parser='lalr', lexer='longest_match')

data = 'ac'
tree = parser.parse(data)
print(tree)
```
This is a grammar that neither the basic nor contextual lexer can deal with. Both will use AB, and the contextual will not try AC as AB is a possible token to parse from start. It's not possible for the programmer to set the precedence in this scenario to tokenize "ab" or "ac" correctly. Using basic yields:
```
lark.exceptions.UnexpectedCharacters: No terminal matches 'c' in the current parser context, at line 1 col 2

ac
 ^
Expected one of: 
	* AB
	* AC
```
Using contextual yields:
```
lark.exceptions.UnexpectedCharacters: No terminal matches 'c' in the current parser context, at line 1 col 2

ac
 ^
Expected one of: 
	* <END-OF-FILE>

Previous tokens: Token('AB', 'a')
```
Regarding the implementation it consists of a new lexer and a new scanner: LongestMatchLexer, which inherits from BasicLexer; LongestMatchScanner, which attempts to match against every terminal, yielding the longest match. (Not optimal - but it's an option.)

It seems some other users have attempted to use longest matches (as I tried when I used lark first):
https://github.com/lark-parser/lark/issues/370
https://github.com/lark-parser/lark/issues/1463

Edit:
An issue with using **earley** instead, is that it may not yield the desired result for ambiguous grammars such as in the example below (which is a simplification of a grammar that worked for a lex derivation):
```
from lark import Lark

grammar = r"""
  start: INTEGER "." INTEGER -> eproj
    | DOUBLE -> double

    INTEGER: /\d+/
    DOUBLE: /\d+\.\d+/
"""

parser = Lark(grammar, start='start', parser='earley')
result = parser.parse("1.2")
print(result)
```
Which yields:
```
Tree('eproj', [Token('INTEGER', '1'), Token('INTEGER', '2')])
```

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -135,6 +135,8 @@ When using a lexer (basic or contextual), it is the grammar-author's responsibil
 3. Length of literal / pattern definition
 4. Name
 
+When using the longest_match lexer, for matches that have the same length the literal that is defined first is used.
+
 **Examples:**
 ```perl
 IF: "if"
--- a/docs/how_to_use.md
+++ b/docs/how_to_use.md
@@ -44,7 +44,7 @@ But if it doesn't, feel free to ask us on gitter, or even open an issue. Post a 
 
 ### Regex collisions
 
-A likely source of bugs occurs when two regexes in a grammar can match the same input. If both terminals have the same priority, most lexers would arbitrarily choose the first one that matches, which isn't always the desired one. (a notable exception is the `dynamic_complete` lexer, which always tries all variations. But its users pay for that with performance.)
+A likely source of bugs occurs when two regexes in a grammar can match the same input. If both terminals have the same priority, most lexers would arbitrarily choose the first one that matches, which isn't always the desired one. (a notable exception is the `dynamic_complete` lexer, which always tries all variations. But its users pay for that with performance.) The `longest_match` lexer chooses the terminal with the longest matched length, similiar to the Unix tool "Lex".
 
 These collisions can be hard to notice, and their effects can be difficult to debug, as they are subtle and sometimes hard to reproduce.
 
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -21,7 +21,7 @@ from .load_grammar import load_grammar, FromPackageLoader, Grammar, verify_used_
 from .tree import Tree
 from .common import LexerConf, ParserConf, _ParserArgType, _LexerArgType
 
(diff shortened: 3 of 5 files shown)


MegaIng:
IMO, it is better to just write your grammar correctly which I am pretty sure is always possible. If your grammar is not ambiguous, you will get nice performance guarantees. (i.e. `O(n)` with `lalr`). We cannot give you these guarantees for these ambiguous grammars, and your solution makes the performance worse in all situations.


Note that if you really don't want to adjust your grammar, you can just create a custom lexer and pass it to `Lark` - but I don't think it should be within the library itself.

erezsh:
I share @MegaIng 's concerns regarding the performance of this lexing method.

If we were to include this behavior as an official lexer, I think it makes more sense to specify a subset of "competing" terminals, rather than the entire set of terminals. i.e. that only (AC, AB) will be evaluated for length, and not every terminal in the grammar.

Another thing that is maybe worth pointing out - regexes are technically capable of solving this particular example using a single match:

```
>>> re.match('a(b|c)*', 'ac')
<re.Match object; span=(0, 2), match='ac'>
```

If there was a way to manually (or even automatically?) merge these terminals, and later discern which one was matched, I believe that would address the performance issues, while still supporting this behavior.

erezsh:
Closed for lack of activity

PR #1501, opened 2024-12-25
Tree.data typing - extend type with Token class

Andrej730:
Because it tends to be `Token` also.

See example below:
```python
from lark import Lark, Tree

grammar = """
    start: entity+
    entity: "entity"
"""

parser = Lark(grammar)
tree = parser.parse("entityentityentity")

for subtree in tree.iter_subtrees_topdown():
    # start RULE <class 'lark.lexer.Token'>
    # entity RULE <class 'lark.lexer.Token'>
    # entity RULE <class 'lark.lexer.Token'>
    # entity RULE <class 'lark.lexer.Token'>
    print(subtree.data, subtree.data.type, type(subtree.data))
```

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -51,10 +51,10 @@ class Tree(Generic[_Leaf_T]):
             contain 'a' will consider it in full, including _A and _C for all attributes.
     """
 
-    data: str
+    data: Union[str, "Token"]
     children: 'List[Branch[_Leaf_T]]'
 
-    def __init__(self, data: str, children: 'List[Branch[_Leaf_T]]', meta: Optional[Meta]=None) -> None:
+    def __init__(self, data: Union[str, "Token"], children: 'List[Branch[_Leaf_T]]', meta: Optional[Meta]=None) -> None:
         self.data = data
         self.children = children
         self._meta = meta


Andrej730:
@erezsh

erezsh:
The choice to have data a Token is a choice that I'm now unsure about, and I'm thinking of changing it to only be `str` in future versions.

So, I'm leaning towards not accepting this PR.

Andrej730:
But still, doesn't it makes sense for current typing to just reflect the way it is now?

erezsh:
Token is a subclass of str, so the typing is correct.

Saying it's a Token would mean people could rely on it having a type, line number, etc., which I think isn't a good idea.

Andrej730:
If it's not reliable, what would be a safe to get all tokens for the tree?

Currently I'd use:
```python
from lark import Lark, Tree, Token

grammar = """
    start: entity+
    entity: "entity"
"""

parser = Lark(grammar)
tree = parser.parse("entityentityentity")

# Get all tokens.
for subtree in tree.iter_subtrees_topdown():
    child = subtree.data
    if not isinstance(child, Token):
        continue
    # start RULE <class 'lark.lexer.Token'>
    # entity RULE <class 'lark.lexer.Token'>
    # entity RULE <class 'lark.lexer.Token'>
    # entity RULE <class 'lark.lexer.Token'>
    print(subtree.data, subtree.data.type, type(subtree.data))
```

AdLThinhRose:
_toi can-biet vab hie ra van de = mã nguồn + cấu trúc -why = :toi khong
biet-

Vào Th 4, 25 thg 12, 2024 lúc 17:40 Andrej ***@***.***> đã
viết:

MegaIng:
When getting all tokens of a parse tree, you don't actually want the tokens that are inside of the `data` attributes since they didn't actually belong to that tree but to the grammar (which is itself parsed with lark)


IMO the only result of data attributes being Tokens is that the repr is less readable - I haven't found any actual benefit.


Either way, using `Uniom[str, Token]` instead of `str` is pointless from the static type checker perspective since it's equivalent to just `str`.

Andrej730:
> When getting all tokens of a parse tree, you don't actually want the tokens that are inside of the `data` attributes since they didn't actually belong to that tree but to the grammar (which is itself parsed with lark)

I mean in my case I need to get what grammar tokens were parsed from the query and it's possible using `Tree.data.type`.

AdLThinhRose:
= hien tai can toi da hoa het data?

Vào Th 4, 25 thg 12, 2024 lúc 18:31 Andrej ***@***.***> đã
viết:

MegaIng:
> I mean in my case I need to get what grammar tokens were parsed from the query and it's possible using Tree.data.type.

No? That is always going to be `RULE` as can be seen in your output.

Andrej730:
> No? That is always going to be `RULE` as can be seen in your output.

Oh, sorry, you're right, got confused. 
Then, yeah, `str` will work just fine for this and I have no use case for `type` or other `Token` attributes.

AdLThinhRose:
_data=trung bình +mone/0ne Y. -(ly do ban can nhac la gi)

Vào Th 4, 25 thg 12, 2024 lúc 19:11 Andrej ***@***.***> đã
viết:

Issue #863, opened 2021-03-24
Indenter doesn't work intuitively with ignored tokens


ColonelThirtyTwo:
**Describe the bug**

Indenter is cited as the way to parse whitespace-sensitive languages, but it has unintiutive and obtrusive behavior when there is an ignored token (ex a comment) in the middle of a newline sequence.

**To Reproduce**

Example:
```
GRAMMAR = r"""
	%import common.WS_INLINE
	%import common(C_COMMENT, CPP_COMMENT, SIGNED_NUMBER)

	%declare _INDENT _DEDENT
	%ignore WS_INLINE
	%ignore CPP_COMMENT
	%ignore C_COMMENT
	_NL: /(\r?\n[\t ]*)+/

	VARNAME: /[_a-zA-Z][_a-zA-Z0-9\/]*/

	start: [_NL] stmt*
	stmt: VARNAME _NL [_INDENT stmt+ _DEDENT]
"""

from lark import Lark
from lark.indenter import Indenter

class TreeIndenter(Indenter):
	NL_type = '_NL'
	OPEN_PAREN_types = []
	CLOSE_PAREN_types = []
	INDENT_type = '_INDENT'
	DEDENT_type = '_DEDENT'
	tab_len = 4

parser = Lark(GRAMMAR, postlex=TreeIndenter(), parser="lalr", start="start")
data = """
foo
bar
	/* */
	baz
"""
for tok in parser.lex(data):
	print(repr(tok))
print(parser.parse(data).pretty())
```

Remove the `/* */` from the `data` variable and it parses ok.

Happens because the ignored comment splits up the _NL tokens, and the indenter does not coalesce them.

erezsh:
The recommended solution is to capture comments in the newline.

Something like:

```lark
_NL: (NEWLINE | COMMENT)+
```

If you have a better way, we'll be happy to hear it.

ColonelThirtyTwo:
@erezsh Thanks for the quick response. Looks like that works.

Might be worth putting that in the docs somewhere - seems like a gotcha. I think that the Indenter could be altered to coalesce adjacent _NL tokens too, so that the alteration isn't needed.

erezsh:
I think you're right, it might be possible to do so. It would prevent languages in which a blank space is a dedent, but perhaps those aren't very common.

Anyway, in terms of performance, the current solution works best.

We can keep this issue open, while I consider the best option. If the code remains the same, I agree we should mention this somewhere in the docs.

[...]

PR #1502, opened 2025-01-02
Add guidance on handling comments in languages with significant indentation

nchammas:
- Expand the indented tree example to show how to correctly handle comments.
- Add a new section to Recipes that references this example, as it's perhaps the most accessible starting point for those looking to write a grammar with significant indentation.
- Add some guidances on how to build the docs.
- Fix a warning coming from Sphinx by configuring the language explicitly.

Related to (though perhaps not exactly a fix for) #863.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ tags
 /build
 docs/_build
 docs/examples
+docs/sg_execution_times.rst
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -202,5 +202,5 @@ If your grammar needs to support significant whitespace, you will need to use th
 Take a look at the [indented tree example][indent] as well as the [Python grammar][python] for
 inspiration.
 
-[indent]: /examples/indented_tree.py
+[indent]: examples/indented_tree.html
 [python]: https://github.com/lark-parser/lark/blob/master/lark/grammars/python.lark
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -79,6 +79,7 @@ Prints out:
 
 *Note: We don't have to return a token, because comments are ignored*
 
+
 ## CollapseAmbiguities
 
 Parsing ambiguous texts with earley and `ambiguity='explicit'` produces a single tree with `_ambig` nodes to mark where the ambiguity occurred.
@@ -193,3 +194,13 @@ def parse_with_progress(parser: Lark, text: str, start=None):
 ```
 
 Keep in mind that this implementation relies on the `InteractiveParser` and, therefore, only works with the `LALR(1)` parser, and not `Earley`.
+
+
+## Parsing a Language with Significant Whitespace
+
+If your grammar needs to support significant whitespace, you will need to use the `Indenter` class.
+Take a look at the [indented tree example][indent] as well as the [Python grammar][python] for
+inspiration.
+
+[indent]: /examples/indented_tree.py
+[python]: https://github.com/lark-parser/lark/blob/master/lark/grammars/python.lark
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -196,11 +196,11 @@ def parse_with_progress(parser: Lark, text: str, start=None):
 Keep in mind that this implementation relies on the `InteractiveParser` and, therefore, only works with the `LALR(1)` parser, and not `Earley`.
 
 
-## Parsing a Language with Significant Whitespace
+## Parsing a Language with Significant Indentation
 
-If your grammar needs to support significant whitespace, you will need to use the `Indenter` class.
-Take a look at the [indented tree example][indent] as well as the [Python grammar][python] for
-inspiration.
+If your grammar needs to support significant indentation (e.g. Python, YAML), you will need to use
+the `Indenter` class. Take a look at the [indented tree example][indent] as well as the
+[Python grammar][python] for inspiration.
 
 [indent]: examples/indented_tree.html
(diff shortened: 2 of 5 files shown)


examples/indented_tree.py
```diff
@@ -3,28 +3,34 @@
...
+    %ignore SH_COMMENT
+    %declare _INDENT _DEDENT
 
-    _NL: /(\r?\n[\t ]*)+/
+    ?start: _NL* tree
+    tree: NAME _NL [_INDENT tree+ _DEDENT]
+    _NL: (/\r?\n[\t ]*/ | SH_COMMENT)+
 """
 
 class TreeIndenter(Indenter):
```
nchammas, line 36: Btw, this is tangential to the purpose of this PR, but I'm curious to know:
1. Why is `tab_len` set to 8? Perhaps I misunderstand its purpose, but the example seems to demonstrate a tab length of 4. And yet, it parses correctly. So I'm a bit confused.
2. These attributes of `Indenter` are decorated with `@property`, so my IDE complains about how they are overridden:
    > "OPEN_PAREN_types" incorrectly overrides property of same name in class "Indenter"
    > Pylance [reportIncompatibleMethodOverride](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportIncompatibleMethodOverride)

    That's perhaps something to improve in the future.

examples/indented_tree.py
```diff
@@ -3,28 +3,34 @@
...
+manufacture `INDENT`/`DEDENT` tokens.
 
-It is crucial for the indenter that the NL_type matches
-the spaces (and tabs) after the newline.
+It is crucial for the indenter that the `NL_type` matches the spaces (and
+tabs) after the newline.
+
+If your whitespace-significant grammar supports comments, then `NL_type`
+must match those comments too. Otherwise, comments that appear in the middle
+of a line will [confuse Lark][1].
```
nchammas, line 16: Hmm, I wonder if this is supposed to use RST syntax instead of Markdown. If I can get some guidance on how to build and review the docs locally, I can confirm the correct syntax to use.

erezsh: The docs are fairly easy to build imho. Just `cd docs && make` (but make sure to install all the dependencies through `docs/requirements.txt`)

docs/recipes.md
```diff
@@ -193,3 +194,13 @@ def parse_with_progress(parser: Lark, text: str, start=None):
 ```
 
 Keep in mind that this implementation relies on the `InteractiveParser` and, therefore, only works with the `LALR(1)` parser, and not `Earley`.
+
+
+## Parsing a Language with Significant Whitespace
```
erezsh, line 199: Indentation is a specific kind of significant whitespace. Better to call it significant indentation, or such.

examples/indented_tree.py
nchammas: Just to be clear, when we say "tab" in this context we mean a literal tab character, correct? i.e. `\t`. So this has nothing to do with specifying how many spaces you want to be part of a single "indent".

Does that mean that when Lark detects indentation (assuming we are only using spaces and not tabs), it doesn't care how many spaces are used? It just looks for the common indentation level.

``` 
a
   b
   c
```

I'm using an indent of 3 spaces here. This will parse fine as an indent and has nothing to do with `tab_len`. 

Am I understanding correctly?

.gitignore
```diff
@@ -12,3 +12,4 @@ tags
 /build
 docs/_build
 docs/examples
+docs/sg_execution_times.rst
```
nchammas, line 15: Dunno if this shows up for you too after running `make html`, but I assume it's meant to be ignored.

nchammas (commented):
I added some instructions on how to build the docs and, since I am now able to review the built docs, I also fixed the markup.

examples/indented_tree.py
erezsh: Yes, that's right.

.gitignore
nchammas: Could it be the version of Sphinx you are running? This is what I have:

```
Sphinx                        7.3.7
sphinx-gallery                0.18.0
sphinx-markdown-tables        0.0.17
sphinx-rtd-theme              3.0.2
sphinxcontrib-applehelp       2.0.0
sphinxcontrib-devhelp         2.0.0
sphinxcontrib-htmlhelp        2.1.0
sphinxcontrib-jquery          4.1
sphinxcontrib-jsmath          1.0.1
sphinxcontrib-qthelp          2.0.0
sphinxcontrib-serializinghtml 2.0.0
```

nchammas: The dev requirements are open ended, so if you haven't updated them in a while you are likely running a much older version of Sphinx than I am (since I just installed them today).

If you're courageous, I would try `pip install -U -r docs/requirements.txt` and then see if you get the same behavior that I am seeing.

erezsh: Yes, you're right, it gets generated by sphinx-gallery. I think it's fine to just ignore it.

erezsh:
Thanks!

nchammas:
@erezsh - I am not seeing these updates on [readthedocs/latest][1], even though that version of the docs should in theory be tracking `master`. Is something about the documentation build broken, perhaps?

[1]: https://lark-parser.readthedocs.io/en/latest/recipes.html

nchammas:
Oh, [I see it now][1]!

[1]: https://lark-parser.readthedocs.io/en/latest/recipes.html#parsing-a-language-with-significant-indentation

PR #1503, opened 2025-01-03
Type `Terminal.__init__`

tybug:
Ran into this in my quest to strictly type [Hypothesis](https://github.com/hypothesisWorks/hypothesis/) 🙂


erezsh:
Thanks!

Issue #1472, opened 2024-09-23
`cache` option does not work with `Reconstructor`


weaversam8:
**Describe the bug**

If an `lalr` grammar is loaded from a cached file, and then the parser is passed to initialize a `Reconstructor(...)` class, initialization fails with the following error:

```
AttributeError: 'Lark' object has no attribute 'grammar'
```

**To Reproduce**

1. Open a parser with `Lark.open(...)` while setting the `cache` argument to a file path.
2. Try to initialize a `Reconstructor(...)` with that parser.

erezsh:
I think this can now be closed as fixed. Feel free to reopen if you think it's still an issue.

PR #1506, opened 2025-01-17
Serialize Lark.grammar (fixes issue #1472)

NasalDaemon:
Fixes https://github.com/lark-parser/lark/issues/1472

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -278,7 +278,7 @@ class Lark(Serialize):
     def __init__(self, grammar: 'Union[Grammar, str, IO[str]]', **options) -> None:
         self.options = LarkOptions(options)
         if self.options.cache_grammar:
-            self.__serialize_fields__ += 'grammar'
+            self.__serialize_fields__.append('grammar')
         re_module: types.ModuleType
 
         # Set regex or re module
@@ -341,7 +341,7 @@ class Lark(Serialize):
                         username = "unknown"
 
 
-                    cache_fn = tempfile.gettempdir() + "/.lark_%s_%s_%s_%s_%s_%s.tmp" % (
+                    cache_fn = tempfile.gettempdir() + "/.lark_%s_%s_%s_%s_%s.tmp" % (
                         "cache_grammar" if self.options.cache_grammar else "cache", username, cache_sha256, *sys.version_info[:2])
 
                 old_options = self.options
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 from .exceptions import ConfigurationError, assert_config, UnexpectedInput
 from .utils import Serialize, SerializeMemoizer, FS, logger
-from .load_grammar import load_grammar, FromPackageLoader, Grammar, verify_used_files, PackageResource, sha256_digest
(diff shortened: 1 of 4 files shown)


erezsh:
How much does this change add to the cache size? (in rough terms)

NasalDaemon:
> How much does this change add to the cache size? (in rough terms)

In my scenario with a file.lark of 4kb:

cache without grammar: 63,994 bytes
cache with grammar: 115,970 bytes

making an increase of ~52kb

As a quick and dirty performance test, I ran my script which uses lark to parse a file and produce a cpp (with jinja) multiple times under different regimes.

Without caching: 440ms
With caching, but using `Lark.grammer = load_grammar(...)`: 370ms
With caching+serializing grammar: 320ms

This shows that there is indeed a noticeable speed benefit even if the cache has become larger, despite the overhead of jinja and the rest of my script.

erezsh:
How does serializing the grammar make it run faster? It only adds operations, and doesn't remove any.

NasalDaemon:
> How does serializing the grammar make it run faster? It only adds operations, and doesn't remove any.

According to the above tests, deserialising the grammar from the cache is faster than reproducing it from the source .lark file each time.

erezsh:
I think you're mistaken.

NasalDaemon:
> I think you're mistaken.

If you would like to avoid the cost of de/serialising the grammar by default, I can make its serialisation optional instead, so you must specify cacheWithGrammar=True.

erezsh:
I prefer `cache_grammar=False`.

Also consider adding text about this in the reconstructor error.

NasalDaemon:
Sorry, was away on holiday after your last comment. I have committed further changes taking it into account. Any further feedback?

erezsh:
@NasalDaemon Can you explain the code in tree_matcher? Specfically when postlex is None?

NasalDaemon:
> @NasalDaemon Can you explain the code in tree_matcher? Specfically when postlex is None?

It looks like the grammar never actually needs to be compiled from the parser, as the rules and terminals of a parser are always created during its construction (including from a cache), so I've completely removed the call to compile and just copy the stored rules and tokens from parser.

https://github.com/lark-parser/lark/blob/87bb8efe0d425187159b39fd788640da33d1878e/lark/lark.py#L407
https://github.com/lark-parser/lark/blob/87bb8efe0d425187159b39fd788640da33d1878e/lark/lark.py#L540-L544

erezsh:
Hmm no, I get it now. You had it right before. When there is a postlex, the `always_accepts` value can affect the compilation, which will disrupt the parser. So when there is a postlex, we need to recompile the grammar. But otherwise, we probably don't.

NasalDaemon:
> Hmm no, I get it now. You had it right before. When there is a postlex, the `always_accepts` value can affect the compilation, which will disrupt the reconstructor. So when there is a postlex, we need to recompile the grammar. But otherwise, you're probably right that we don't need to.

Ah, indeed! I've reverted to the old behaviour again

erezsh:
@NasalDaemon I made some fixes in a new PR. Feel free to review. The only thing that's missing is tests. I'll do it some day, unless you want to do it.

erezsh:
Thanks for the contribution!

PR #1507, opened 2025-01-21
Add note on binding power of numeric quantifiers

414owen:
I found this part of the syntax non-obvious, because `~` could conceivably have a lower binding power than sequences, so I added an example to the docs.


erezsh:
Thanks!

PR #1518, opened 2025-03-09
Fix generic type of Transformer_InPlaceRecursive

### LINKED ISSUE #1517: Missing Generic annotation for Transformer_InPlaceRecursive
**Describe the bug**

Different from the other Transformer classes, `Transformer_InPlaceRecursive` is not annotated with the generic types `[_Leaf_T, _Return_T]`. It is also not exported in `__init__.py` like the other subclasses.

Is there a reason for it?

**To Reproduce**

Create a custom subclass like

```python
class MyTransformer(Transformer_InPlaceRecursive[Token, ParseTree]):
    pass
```

add generic annotations to it as you would with other Transformers, and observe the type checker errors. I reproduces this with the current pyright release.

lbhm:
This fixes #1517

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -333,7 +333,7 @@ class Transformer_NonRecursive(Transformer[_Leaf_T, _Return_T]):
         return cast(_Return_T, result)
 
 
-class Transformer_InPlaceRecursive(Transformer):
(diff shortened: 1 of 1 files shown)


erezsh:
LGTM.

PR #1524, opened 2025-04-07
add function `replace_nodes` to Tree

zzc0430:
This can be used to replace tokens in a `Tree` that satisfy a specified rule.

### DIFF
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -195,7 +195,6 @@ class Tree(Generic[_Leaf_T]):
                 changed = True
         return changed
 
-
     def scan_values(self, pred: 'Callable[[Branch[_Leaf_T]], bool]') -> Iterator[_Leaf_T]:
         """Return all values in the tree that evaluate pred(value) as true.
 
@@ -211,7 +210,21 @@ class Tree(Generic[_Leaf_T]):
             else:
                 if pred(c):
                     yield c
+                    
+    def replace_nodes(self, pred: 'Callable[[Token], Optional[Token]]') -> None:
+        """replace tokens in the tree using the result of pred(token) when it is not none.
 
+        Example:
+            >>> tree.replace_nodes(lambda v: v.update(value=v.rstrip("\n")) if v.endswith("\n") else None)
+        """
+        for index, child in enumerate(self.children):
+            if isinstance(child, Tree):
+                self.replace_nodes(pred)
+            elif isinstance(child, Token):
+                result = pred(child)
+                if isinstance(result, Token):
+                    self.children[index] = result
+        
     def __deepcopy__(self, memo):
         return type(self)(self.data, deepcopy(self.children, memo), meta=self._meta)
 
--- a/lark/tree.py
+++ b/lark/tree.py
@@ -210,7 +210,7 @@ class Tree(Generic[_Leaf_T]):
             else:
                 if pred(c):
                     yield c
-                    
+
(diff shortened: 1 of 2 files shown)


erezsh:
Add tests.

lark/tree.py
```diff
@@ -211,7 +210,21 @@ def scan_values(self, pred: 'Callable[[Branch[_Leaf_T]], bool]') -> Iterator[_Le
             else:
                 if pred(c):
                     yield c
+                    
+    def replace_nodes(self, pred: 'Callable[[Token], Optional[Token]]') -> None:
+        """replace tokens in the tree using the result of pred(token) when it is not none.
```
MegaIng, line 215: The description here `it is not none` does not match the implemented behavior `isinstance(node, Token)`. I think you should stick with `is not none` and change the implementation, sometimes there is a reason to use none-Token nodes. (Not sure what to do about the type hints).

MegaIng (changes requested):

zzc0430:
> Add tests.

added 388fd80af56469302e2dac907cf13d973582df93

lark/tree.py
zzc0430: fixed 6194e05b6175022de664c9d2d1c3e92f21a88713

erezsh:
Let's call it `replace_tokens`.

Also in the tests, you are changing `tree2` in-place, which could break other tests if run in a different order.

zzc0430:
> Let's call it `replace_tokens`.
> 
> Also in the tests, you are changing `tree2` in-place, which could break other tests if run in a different order.

done 545b8087ff521e536100a080f9b663c2175f6e7c, caea7808abdc4dfff34d22b411ff736deb025279

erezsh:
*deepcopy

zzc0430:
> *deepcopy

5ab7313081a8d7cc004796c7038591d7937ad17c

MegaIng (approved):

MegaIng:
True, probably not needed.

To get the type checking to work:

- replace `Token` in the signature with the Leaf generic parameter.
- instead of `isinstance(..., Token)` just use an else statement.

This slightly changes the semantics, but I think it's more useful.

zzc0430:
> True, probably not needed.
> 
> To get the type checking to work:
> 
> * replace `Token` in the signature with the Leaf generic parameter.
> * instead of `isinstance(..., Token)` just use an else statement.
> 
> This slightly changes the semantics, but I think it's more useful.

c21139c95a074225f72486f85cf2b9e701830103

erezsh:
Tbh I'm not convinced this is a useful method. I can't imagine many situations where it's desirable.

But here's how it should look, if it was ever accepted: (more or less)

```python
    def replace_values(self, replacer: 'Callable[[_Leaf_T], _Leaf_T]') -> None:
        """Replace all leaf values in the tree using the result of ``replacer(value)``.

        The callback receives each leaf and should return its replacement,
        or the same value unchanged if no replacement is desired.

        Example:
            >>> tree.replace_values(lambda v: v.update(value="y") if v == "x" else v)
        """
        for index, child in enumerate(self.children):
            if isinstance(child, Tree):
                child.replace_values(replacer)
            else:
                result = replacer(child)
                if result is not child:
                    self.children[index] = result
```

erezsh:
Closed due to inactivity.

Sorry for closing this after all this back and forth. I'll take it as a reminder to address my core doubts first in the future, instead of getting caught in the more superficial details.

PR #1531, opened 2025-05-08
better error messages for shift/reduct conflicts

skogsbaer:
In case of strict=True, include the offending rule for shift/reduce conflicts


erezsh:
Did you test it, and made sure it displays the rule?

skogsbaer:
Now it does. I tested it manually.

PR #1535, opened 2025-05-15
made the interpreter code example correct using @visit_children_decor

sakshamwithweb:


erezsh:
Why is that correct?

sakshamwithweb:
Apologies, I forgot that num was for a leaf node so there is no need to use visit or visit children.
Let me close it.

PR #1540, opened 2025-07-06
Fixes for PR #1506, which adds the option to cache the grammar definition

erezsh:
Fixes for PR #1506


NasalDaemon (approved):

MegaIng (commented):
I would like to see tests before approving this - the logic isn't quite trivial enough to not have edge cases I can't think of or to not be broken in a future "unrelated" code change. Otherwise looks good.

erezsh:
@MegaIng Added tests

MegaIng (approved):

PR #1551, opened 2025-09-05
Fix standalone parser import error generated with --compress

mikebohdan:
### Problem
When generating and importing a standalone parser using the `--compress` flag, importing it later raises:
```
ModuleNotFoundError: No module named 'lark'
```
This happens because the generated module references `lark` during unpickling, even though its contents are embedded—so the original `lark` module isn’t installed in that context.

### Solution
Added a **custom unpickler** in the generated module that intercepts references to the original `lark` module and redirects them to the embedded content in the standalone parser.

### Demo / Verification
See a working demonstration here:  
https://github.com/mikebohdan/lark-compress-demo


mikebohdan:
Hello @erezsh, thank you for responding. Here is the [issue](https://github.com/lark-parser/lark/issues/1552) you requested.

erezsh:
Fixed in PR #1553

Issue #1560, opened 2025-10-20
Backward compatibility broken for custom lexers which parse a non-textual stream of data


jmfernandez:
**Describe the bug**

Lark 1.2.2 supported custom lexers focused on parsing non-textual stream of data . Lark 1.3.0 has broken that support, as you can see running stable documentation's example: 

```python-traceback
python parse_to_dict.py 
['alice', 1, 27, 3, 'bob', 4, 'carrie', 'dan', 8, 6]
Traceback (most recent call last):
  File "/home/jmfernandez/projects/python-groovy-parser/parse_to_dict.py", line 46, in <module>
    test()
    ~~~~^^
  File "/home/jmfernandez/projects/python-groovy-parser/parse_to_dict.py", line 38, in test
    tree = parser.parse(data)
  File "/home/jmfernandez/projects/python-groovy-parser/.full13/lib/python3.13/site-packages/lark/lark.py", line 676, in parse
    return self.parser.parse(text, start=start, on_error=on_error)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jmfernandez/projects/python-groovy-parser/.full13/lib/python3.13/site-packages/lark/parser_frontends.py", line 122, in parse
    stream = self._make_lexer_thread(text)
  File "/home/jmfernandez/projects/python-groovy-parser/.full13/lib/python3.13/site-packages/lark/parser_frontends.py", line 113, in _make_lexer_thread
    return text if self.skip_lexer else cls(self.lexer, None) if text is None else cls.from_text(self.lexer, text)
                                                                                   ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/jmfernandez/projects/python-groovy-parser/.full13/lib/python3.13/site-packages/lark/lexer.py", line 457, in from_text
    text = TextSlice.cast_from(text_or_slice)
  File "/home/jmfernandez/projects/python-groovy-parser/.full13/lib/python3.13/site-packages/lark/utils.py", line 213, in cast_from
    return cls(text, 0, len(text))
  File "<string>", line 6, in __init__
  File "/home/jmfernandez/projects/python-groovy-parser/.full13/lib/python3.13/site-packages/lark/utils.py", line 196, in __post_init__
    raise TypeError("text must be str or bytes")
TypeError: text must be str or bytes
```

**To Reproduce**

[...]

PR #1562, opened 2025-10-21
Bugfix: Restore support for custom input, alongside text and TextSlice

erezsh:
Fixes issue #1560

### DIFF
--- a/lark/parser_frontends.py
+++ b/lark/parser_frontends.py
@@ -1,7 +1,7 @@
 from typing import Any, Callable, Dict, Optional, Collection, Union, TYPE_CHECKING
 
 from .exceptions import ConfigurationError, GrammarError, assert_config
-from .utils import get_regexp_width, Serialize, TextOrSlice, TextSlice
+from .utils import get_regexp_width, Serialize, TextOrSlice, TextSlice, LarkInput
 from .lexer import LexerThread, BasicLexer, ContextualLexer, Lexer
 from .parsers import earley, xearley, cyk
 from .parsers.lalr_parser import LALR_Parser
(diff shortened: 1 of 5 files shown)


lark/parser_frontends.py
```diff
@@ -108,11 +110,15 @@ def _verify_start(self, start=None):
...
 
-    def _make_lexer_thread(self, text: Optional[TextOrSlice]) -> Union[TextOrSlice, LexerThread, None]:
+    def _make_lexer_thread(self, text: Optional[LarkInput]) -> Union[LarkInput, LexerThread, None]:
         cls = (self.options and self.options._plugins.get('LexerThread')) or LexerThread
-        return text if self.skip_lexer else cls(self.lexer, None) if text is None else cls.from_text(self.lexer, text)
+        if self.skip_lexer:
+            return text
+        if text is None:
+            return cls(self.lexer, None)
+        return cls.from_text(self.lexer, text) if isinstance(text, (str, bytes, TextSlice)) else cls.from_custom_input(self.lexer, text)
```
MegaIng: At this point I would split this line up completely for future maintainability and ease of reading.

lark/utils.py
```diff
@@ -226,6 +226,7 @@ def rindex(self, substr: AnyStr):
 
 
 TextOrSlice = Union[AnyStr, 'TextSlice[AnyStr]']
+LarkInput = Union[AnyStr, TextSlice[AnyStr], Any]
```
MegaIng, line 229: I would write this as `LarkInput = Union[TextOrSlice, Any]` to make it clearer what the logic here is. Functionally they are equivalent.

MegaIng (approved):
Excluding those two minor style changes, this looks good.

Sorry, I also forgot about this usecase when reviewing the original change set.

lark/utils.py
erezsh: I guess `Any` overrides any other type in the union, but I want there to be an explicit list of the types that could be accepted in the normal case.

MegaIng: IMO the explicit lists is also just above the line anyway, but sure, if you want to keep it like this that's fine.

(And yes, from a type-checking perspective having Any in there removes all checks - to fix this we would need essentially depended types which don't (yet) exists in python's typing system)

erezsh: I have a mild preference to keeping it this way.

If we really cared about typing, we could have a different function like `parse_custom()`. But I'm not sure if it's worth the trouble.

MegaIng: It's not IMO and still a breaking change.

MegaIng (approved):

PR #1572, opened 2026-01-09
Replace dead wikiwand links

johan456789:
Replaces dead wikiwand links with direct Wikipedia URLs in docs.


erezsh:
Thank you but it seems like it was only temporarily down.

PR #1575, opened 2026-03-02
exceptions.py: sort set of expected things to get deterministic result

denballakh:
Currently expected tokens are stored in a set, and because of that the order they appear in error message is unstable - different runs may produce different orderings.

Here I sort the set of expected tokens, so that the order is guaranteed to be stable between runs.


erezsh:
Looks good to me. Thanks for contributing.

PR #1577, opened 2026-03-07
Bugfix Earley: handle ambiguity from ignored tokens

### LINKED ISSUE #1576: Ambiguity from trailing whitespace causes a RuntimeError in Earley
```py
from lark import Lark

parser = Lark(
    r'''
    %import common.WS
    %ignore WS

    start: "foobar" | "foobar "
    '''
)

print(parser.parse("foobar "))
```

```
RuntimeError: Earley should not generate multiple start symbol items! Please report this bug.
```

I got this in a much larger grammar with a complex regex and managed to simplify it down this far. I don't know what the expected output is, but the `RuntimeError` told me to report it, so here we are.

chanicpanic:
Fixes #1576: an edge case in which start symbol nodes were not shared between start symbol items that completed at different parts of the string due to ignores.

Thanks @kwilinsi for the great MCVE!

### DIFF
--- a/lark/parsers/xearley.py
+++ b/lark/parsers/xearley.py
@@ -22,6 +22,7 @@ from ..exceptions import UnexpectedCharacters
 from ..lexer import Token
 from ..grammar import Terminal
 from .earley import Parser as BaseParser
+from .earley_common import Item
 from .earley_forest import TokenNode
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


chanicpanic:
I realized that the issue also applies to ambiguity involving ignores anywhere in the string. The parser now produces previously unsurfaced derivations.

erezsh:
Nice, thank you!

PR #1586, opened 2026-04-07
Fix type hint of @v_args() decorator

### LINKED ISSUE #1579: `v_args` decorator retval type hint isn't correct, might as well be removed
**Describe the bug**

The return value type hint for `v_args()` isn't fully correct

https://github.com/lark-parser/lark/blob/52ef09dfe19b1bfb18a8d0a553b964363a0332d0/lark/visitors.py#L513

When ` -> Callable[[_DECORATED], _DECORATED]` is removed entirely, i.e. the line changed to:

```py
def v_args(inline: bool = False, meta: bool = False, tree: bool = False, wrapper: Optional[Callable] = None):
```

my typechecker (pyright/basedpyright, i.e. pylance from VSCode) properly infers the return type of `transform()`

**To Reproduce**

```python
from lark import Transformer, v_args, Token, Tree

@v_args(meta=True)
class Foo(Transformer[Token, str]):
    def __init__(self):
        super().__init__()

    def foo(self, meta, args):
        return 'baz'


bar = Foo().transform(Tree('foo', []))
assert bar == 'baz'
```

with ` -> Callable[[_DECORATED], _DECORATED]`, type inference breaks and `bar` is considered `Any | Unknown`

without, `bar` is correctly considered `str`

Kache:
Before, the test script below fails Pyright's type-checking:

```
error: "assert_type" mismatch: expected "str" but received "Any | Unknown" (reportAssertTypeFailure)
```

(VSCode's python extension, Pylance, uses Pyright internally)

Test script:

```python
import sys

from lark import Transformer, Token, Tree, v_args

if sys.version_info < (3, 11):
    from typing_extensions import assert_type, reveal_type
else:
    from typing import assert_type, reveal_type

@v_args()
class Foo(Transformer[Token, str]):
    def __init__(self):
        super().__init__()

    def foo(self, meta, args):
        return 'baz'

bar = Foo().transform(Tree('foo', []))

assert_type(bar, str)
```

I've not added this test because it would require adding pyright as a second typechecker dependency when the project already uses mypy

fix: #1579

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -15,8 +15,8 @@ _Return_V = TypeVar('_Return_V')
 _Leaf_T = TypeVar('_Leaf_T')
 _Leaf_U = TypeVar('_Leaf_U')
 _R = TypeVar('_R')
-_FUNC = Callable[..., _Return_T]
-_DECORATED = Union[_FUNC, type]
+_DECORATED = Union[Callable[..., _Return_T], Type[_Return_T]]
+_DECORATOR = Callable[[_DECORATED[_Return_T]], _DECORATED[_Return_T]]
 
 class _DiscardType:
     """When the Discard value is returned from a transformer callback,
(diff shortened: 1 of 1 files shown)


erezsh:
Sorry, we still support 3.8 (for now)

Kache:
Yeah, I thought branching on `sys.version_info` might work, but now attempting `try/except`

(I'm not able to run 3.8 locally b/c of yanked dependencies or something)

Kache:
@erezsh fyi, now green

Kache:
hey @erezsh, just pinging again

erezsh:
Looks ok. Remove the `import sys` and I'll accept it.

Kache:
whoops, done

PR #1588, opened 2026-04-20
docs: migrate SPPF page from Bootstrap 4 to Bootstrap 5

### LINKED ISSUE #1494: Move documentation to bootstrap 5
At present the [Shared Packed Parse Forest (SPPF) documentation](https://github.com/lark-parser/lark/blob/2f7c9a4e52ab82bbf1fd00699c79e8a6826fd9c4/docs/_static/sppf/sppf.html) makes use of bootstrap 4. [Bootstrap 4 reached end-of-life on January 1, 2023](https://getbootstrap.com/docs/4.6/end-of-life/). Therefore it would be nice to migrate the documentation to bootstrap 5.

Helpful links for the migration can be found in the [corresponding Debian bug#1088592](https://bugs.debian.org/1088592).

akerem16:
## Summary

Fixes #1494.

Migrates \`docs/_static/sppf/sppf.html\` off of Bootstrap 4 (EOL since 2023-01-01) onto Bootstrap 5.

## Changes

- **CDN:** \`stackpath.bootstrapcdn.com/bootstrap/4.3.1\` → \`cdn.jsdelivr.net/npm/bootstrap@5.3.3\`. stackpath.bootstrapcdn.com never mirrored Bootstrap 5, so a switch of host is required. Added an SRI \`integrity\` hash computed locally against the jsDelivr asset.
- **Utility class:** the only Bootstrap utility on the page was \`pull-right\` (legacy Bootstrap 3 name that survived through v4). Renamed to \`float-end\` per the v5 migration guide.

Nothing else on the page uses Bootstrap classes, \`data-toggle\` attributes, or jQuery, so no further changes were needed.

## Test plan

- [x] HTML diff is two character-level changes; behavior on the SPPF page is unchanged (the back-to-top link stays right-aligned via \`float-end\`).
- [x] SRI hash regenerated from the upstream jsDelivr asset to match 5.3.3.

### DIFF
--- a/docs/_static/sppf/sppf.html
+++ b/docs/_static/sppf/sppf.html
@@ -3,7 +3,7 @@
 <html lang="en-us">
   <head>
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-<link rel="stylesheet" type="text/css" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css">
+<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous">
 
 <body>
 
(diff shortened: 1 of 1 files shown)


piush365:
Looks good to me! Verified the CDN switch from stackpath to jsDelivr is correct since stackpath never hosted Bootstrap 5. The pull-right → float-end rename is the right migration per the Bootstrap 5 docs. SRI hash approach is a good security practice.

PR #1590, opened 2026-04-28
Skip UTF-8 BOM before parsing

### LINKED ISSUE #407: Skip BOM (Byte order mark)
Perhaps it should be opt-in, but most usage would expect a BOM is ignored.

bcmeireles:
skip initial UTF-8 BOM before parsing text input and grammar files. Fixes #407

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -16,7 +16,7 @@ if TYPE_CHECKING:
     from .parser_frontends import ParsingFrontend
 
 from .exceptions import ConfigurationError, assert_config, UnexpectedInput
-from .utils import Serialize, SerializeMemoizer, FS, logger, TextOrSlice, LarkInput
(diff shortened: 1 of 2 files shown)


MegaIng:
IMO this isn't larks job to do:

- utf-8 BOM is very windows specific and generally not recommend - to the point where one can say input is malformed if it contains a BOM
- It's part of the file encoding, which is not the abstraction layer lark operates on. Open files with the `utf-8-sig` encoding and the bom will be removed by the encoding layer built into python (See also: https://discuss.python.org/t/utf-8-bom-not-being-consumed-when-opening-file/74870)
- We *especially* shouldn't do this for random `bytes` input where we don't even know if it's intended to be text.
- (By using slicing we are copying the data which is something we are explicitly trying to avoid. This also falsifies indices. This could be fixed, but I still don't think we should for the reasons above)

PR #1591, opened 2026-05-09
Make Lark generic on transformer return type

### LINKED ISSUE #1186: Wrong type hit when transformer is provided
When `transformer` is provided to the `Parser` constructor, the `parse` method may no longer return the parse `'ParseTree'`

erezsh:
Fixes #1186

One regression: subclasses of Lark using open() etc. will see a regression. But arguably it is pretty unusual, and we're fixing the much more common use-case.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -28,7 +28,7 @@ jobs:
         pip install pytest-cov
         pytest --cov=./ --cov-report=xml
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v1
+      uses: codecov/codecov-action@v5
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         files: ./coverage.xml
@@ -36,5 +36,4 @@ jobs:
         env_vars: OS,PYTHON
         name: codecov-umbrella
         fail_ci_if_error: false
-        path_to_write_report: ./coverage/codecov_report.txt
         verbose: true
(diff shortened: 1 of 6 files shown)


erezsh:
@MegaIng Any chance you can give it a quick look?

MegaIng:
IMO we should exclude `test_typing.py` from the coverage check to make the output cleaner. I don't think we currently have a coverage config file.

While reading through the logs I noticed that coverage can't parse `test_pattern_matching.py` since it's running with 3.8 and pattern matching is only support in 3.10+. I think moving up the coverage run from 3.8 to 3.10 is reasonable, but I am not going to insist if you don't want to.

MegaIng (changes requested):
The implementation looks good, we should just exclude the file from coverage.

The regression with subclasses is annoying, but I agree that this should be rare. It's jut a limitation of python's typing system not having HKTs

erezsh:
Looks ok?

MegaIng (approved):

InCogNiTo124:
Hi all, I'm using `lark==1.3.1` and today I wrote a parser with a transformer that broke pyrefly type inference. I wanted to actually contribute a bugfix, only to find it fixed in this PR but never released :sweat_smile: and when I tried to check what I wrote with this PR included, it worked :) what are your thoughts on bumping lark to `v1.3.2` to publish this feature @erezsh ?

erezsh:
I should publish a new release, but it's going to be a big one, and there are some loose ends to tie up first. But I'll try to do it soon.

PR #1593, opened 2026-05-17
Fix full-width character alignment

Selina-0017:
## Summary

Fix incorrect caret alignment in error context rendering when input contains full-width Unicode characters (e.g. Chinese characters).

Previously, caret positioning used character length (`len()`), which does not reflect terminal display width for full-width characters.

This caused misaligned error indicators in `UnexpectedInput.get_context()`.

## Changes

- Replace naive character-count spacing with display-width-aware computation
- Add regression test for full-width Unicode input

## Testing

- Added test covering full-width character alignment
- All existing tests pass

## Notes

No external dependencies added. Uses Python standard library only.

### DIFF
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -1,6 +1,6 @@
 from .utils import logger, NO_VALUE
 from typing import Mapping, Iterable, Callable, Union, TypeVar, Tuple, Any, List, Set, Optional, Collection, TYPE_CHECKING
-
+import unicodedata
 if TYPE_CHECKING:
     from .lexer import Token
     from .parsers.lalr_interactive_parser import InteractiveParser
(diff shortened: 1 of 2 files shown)


erezsh:
Hi, I ran it through a reviewer, can you address this finding?

> Tab + full-width lines now overshoot the caret (a small regression). expandtabs() computes tab stops by character count while _display_width counts display columns, and the two compose
>   incorrectly. Verified: for before = "你好\tX", a terminal renders the error at display column 9, the old code correctly emitted 9 spaces, and the new code emits 11. So for lines mixing wide chars and tabs, the PR
>   makes alignment worse than before. Fixing it properly needs a width-aware tab expansion (advance to the next multiple of 8 in display columns) instead of _display_width(before.expandtabs()).
>

PR #1602, opened 2026-06-02
Re-enable Joop Leo right-recursion optimization (fixes #397)

### LINKED ISSUE #397: Earley recogniser - Leo transitives
Hi, I noticed that there's an implementation of Joop Leo's optimisation for right recursion, but it is commented [out](https://github.com/lark-parser/lark/blob/f814d91f9dc0d9989a9ef413cca0f2622427eb74/lark/parsers/earley.py#L81). 

I'm fairly new to the concept of this algorithm so I'm trying to learn. Is there a reason to omit this, or maybe there's some other way of achieving linear time for LR(k) grammars?

erezsh:
Leo was disabled when predict_and_complete was refactored into a method, breaking is_quasi_complete's access to start_symbol. This caused incorrect behavior with self-referential start symbols and multiple start symbols.

- Restore TransitiveItem in earley_common.py
- Add start_symbol parameter to predict_and_complete so is_quasi_complete can correctly guard against self-referential cycles
- Move held_completions update before the Leo/regular branch so empty items populate it even when Leo fires (fixes nullable symbol parsing)
- Guard create_leo_transitives with NULLABLE check: Leo must not shortcut nullable right-recursive symbols, as that bypasses the intermediate SPPF nodes where priority/disambiguation lives
- Fix load_paths for multi-hop chains: use reduction.rule.origin instead of next_titem.s (which carries the outer rule's origin, not the intermediate symbol), and merge paths sharing the same intermediate symbol into one canonical SymbolNode to preserve SPPF sharing
- Pass start_symbol through in xearley.py
- Add tests for right-recursive Leo and multi-start-symbol correctness

### DIFF
--- a/lark/parsers/earley.py
+++ b/lark/parsers/earley.py
@@ -18,7 +18,7 @@ from ..exceptions import UnexpectedEOF, UnexpectedToken
 from ..utils import logger, OrderedSet, dedup_list
 from .grammar_analysis import GrammarAnalyzer
 from ..grammar import NonTerminal
-from .earley_common import Item
+from .earley_common import Item, TransitiveItem
 from .earley_forest import ForestSumVisitor, SymbolNode, StableSymbolNode, TokenNode, ForestToParseTree
 
 if TYPE_CHECKING:
@@ -75,7 +75,7 @@ class Parser:
         self.term_matcher = term_matcher
 
 
-    def predict_and_complete(self, i, to_scan, columns, transitives, node_cache):
+    def predict_and_complete(self, i, to_scan, columns, transitives, node_cache, start_symbol=None):
         """The core Earley Predictor and Completer.
 
         At each stage of the input, we handling any completed items (things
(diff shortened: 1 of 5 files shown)


erezsh:
@night199uk Does this seem right? It was vibe-coded, so I don't trust it much. But it does pass the tests.

lark/parsers/earley.py
```diff
@@ -86,6 +86,66 @@ def predict_and_complete(self, i, to_scan, columns, transitives, node_cache):
...
 
+        def is_quasi_complete(item):
+            if item.is_complete:
+                return True
+            quasi = item.advance()
+            while not quasi.is_complete:
+                if quasi.expect not in self.NULLABLE:
+                    return False
+                if quasi.rule.origin == start_symbol and quasi.expect == start_symbol:
+                    return False
```
chanicpanic, line 95: This statement isn't covered by the test cases. So, either we are missing a test case that exercises this branch, or `start_symbol` isn't actually needed here.

(There are a couple other lines in the Joop Leo code that aren't covered, but this one is the most interesting)

lark/parsers/earley.py
```diff
@@ -86,6 +86,66 @@ def predict_and_complete(self, i, to_scan, columns, transitives, node_cache):
...
+                quasi = quasi.advance()
+            return True
+
+        def create_leo_transitives(origin, start):
```
chanicpanic, line 99: The original implementation of this function contains code comments that would probably be helpful to retain.

lark/parsers/earley.py
chanicpanic: While we are touching the transitives code: it would be less confusing for `item.rule.origin` to be used here instead of `item.s`. The existing code is correct given that `item.s == item.rule.origin` when `item.is_complete`, but unexpected given the preceding `if item.rule.origin in transitives[item.start]` condition.

lark/parsers/earley.py
```diff
@@ -122,15 +191,6 @@ def predict_and_complete(self, i, to_scan, columns, transitives, node_cache):
                         items.append(new_item)
                 ###R Regular Earley completer
                 else:
-                    # Empty has 0 length. If we complete an empty symbol in a particular
```
chanicpanic, line 125: Note: None of the tests actually fail if this is not moved up

lark/parsers/earley_forest.py
```diff
@@ -67,11 +67,37 @@ def add_path(self, transitive, node):
         self.paths.add((transitive, node))
 
     def load_paths(self):
+        # Collect canonical intermediate symbol nodes for multi-hop Leo chains.
+        # When a node already IS the intermediate symbol (from a direct Leo completion),
+        # use it as the canonical node so all other paths to the same symbol merge into it,
+        # preserving SPPF sharing and rule-order-based disambiguation.
+        canonical_vn = {}
```
chanicpanic: This function contains the largest amount of new code and, unsurprisingly, is also the most questionable.

After playing around with it a while, I believe the logic is valid, but the implementation definitely reeks of LLM-generation.

I came up with a version that effectively does the same thing, but more concisely and in-line with the patterns of the existing earley code:

```python
node_cache = {(node.s, node.start, node.end): node for (_, node) in self.paths}
for transitive, node in self.paths:
    next_node = node
    if transitive.next_titem is not None:
        label = (transitive.next_titem.reduction.rule.origin, transitive.next_titem.reduction.start, self.end)
        is_node_the_next_node = label == (node.s, node.start, node.end)
        if not is_node_the_next_node:
            next_node = node_cache[label] if label in node_cache else node_cache.setdefault(label, type(self)(*label))
            next_node.add_path(transitive.next_titem, node)

    self.add_family(transitive.reduction.rule.origin, transitive.reduction.rule, transitive.reduction.start, transitive.reduction.node, next_node)
self.paths_loaded = True
```

erezsh: That does look better, thanks! I updated the code accordingly.

erezsh:
Updated the PR based on the review. (simplified, improved coverage)

lark/parsers/earley.py
chanicpanic: Even though one of the new test cases executes this branch, all the tests pass if this branch is removed, so I'm still not convinced it is necessary.

erezsh: Thanks. I guess I'll have to go a bit deeper into it myself, so I don't keep generating slop :)

PR #1603, opened 2026-06-04
Raise GrammarError for non-terminal names in %declare

sarathfrancis90:
`%declare` is meant for terminals, but the grammar accepts any symbol, so a lowercase (rule) name slips through and crashes the compiler:

```python
from lark import Lark
Lark('%declare foo')
# AttributeError: 'NoneType' object has no attribute 'iter_subtrees'
```

A declared rule name gets defined as an abstract rule with no tree (only declared *terminals* are handled that way), and the transform step later calls `iter_subtrees()` on the missing tree. All-uppercase `%declare X Y` is fine; it's only non-terminal names that hit this.

This rejects non-terminal names in `%declare` with a clear `GrammarError` instead of crashing. Added a regression test next to the existing `test_undefined_*` cases.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -1274,6 +1274,8 @@ class GrammarBuilder:
                 for symbol in stmt.children:
                     assert isinstance(symbol, Symbol), symbol
                     is_term = isinstance(symbol, Terminal)
+                    if not is_term:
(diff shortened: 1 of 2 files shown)


erezsh:
Thanks for contributing!

PR #1605, opened 2026-06-05
Show Indenter property docs + add working links

nchammas:
The `Indenter` class has abstract properties that the user needs to implement. They are all documented in code, but the documentation does not show up in the built docs. This PR fixes that, and also adds several links (confirmed to be working locally) to make it easier for the reader to jump to relevant references.

[Current rendered docs.][1] Screenshot from my local build:

[1]: https://lark-parser.readthedocs.io/en/stable/classes.html#indenter

Screenshot 2026-06-05 at 2 07 32 PM

### DIFF
--- a/docs/classes.rst
+++ b/docs/classes.rst
@@ -95,6 +95,7 @@ Indenter
 --------
 
 .. autoclass:: lark.indenter.Indenter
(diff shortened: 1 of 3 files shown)


MegaIng (approved):

erezsh:
Thanks!

PR #1614, opened 2026-06-24
Fix Transformer_NonRecursive crashing when root node is discarded

gaoflow:
## Bug

`Transformer_NonRecursive.transform()` raises `ValueError: not enough values to unpack (expected 1, got 0)` when the root tree node's callback returns `Discard`.

```python
from lark import Lark
from lark.visitors import Transformer_NonRecursive, Discard

class T(Transformer_NonRecursive):
    def start(self, children):
        return Discard

g = Lark('start: A+\nA: /[a-z]+/', lexer='basic')
tree = g.parse('hello')
T().transform(tree)  # ValueError: not enough values to unpack (expected 1, got 0)
```

The regular `Transformer` handles this correctly by returning `None` when the root is discarded (see `Transformer.transform()` lines 161–163).

## Fix

Add the same empty-stack guard to `Transformer_NonRecursive.transform()` before the destructuring assignment.

Two-line change in `lark/visitors.py` + a regression test in `tests/test_trees.py`.

### DIFF
--- a/lark/visitors.py
+++ b/lark/visitors.py
@@ -326,6 +326,8 @@ class Transformer_NonRecursive(Transformer[_Leaf_T, _Return_T]):
             else:
                 stack.append(x)
 
+        if not stack:
+            return None     # type: ignore[return-value]
         result, = stack  # We should have only one tree remaining
         # There are no guarantees on the type of the value produced by calling a user func for a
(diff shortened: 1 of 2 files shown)


erezsh:
Thanks!

PR #1617, opened 2026-07-02
Add Token.as_str() to decode bytes token values

### LINKED ISSUE #1526: Parsing bytestrings and token values
I tried using the calculator example with `use_bytes=True` but it wouldn't parse a NUMBER:

```
ValueError: could not convert string to float: Token('NUMBER', b'1')
```

I had to modify https://github.com/lark-parser/lark/blob/f68e3534f45ccfb481b68ca89ae387383a951e38/examples/calc.py#L46 to:

```python
    number = lambda _, token = float(token.value)
```

even though [`Token.value`](https://lark-parser.readthedocs.io/en/stable/classes.html#lark.Token.value) specifies that it is redundant to write `token.value` because it's always equal to `token`.

Using Lark 1.2.2.

apoorva-01:
`float(token)` breaks with `use_bytes=True`: `could not convert string to float: Token('NUMBER', b'1')`. Token subclasses str, but under use_bytes the value is bytes, so the string is the repr `"b'1'"`, not `"1"`.

Added `Token.as_str()` that decodes bytes (else returns as-is), so `float(token.as_str())` works either way. Used latin-1 since that's what lark uses for lexer patterns and it won't choke on odd bytes. Can switch if you want.

This is the small fix. It doesn't touch the deeper thing where `token.value == token` is false under use_bytes. Making Token bytes-aware would fix that too, but it's a bigger semantics change, so I kept it contained. Happy to go that way instead. Test parses bytes and checks `float(token.as_str())`, fails without the fix.

Closes #1526

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -253,6 +253,17 @@ class Token(str):
     def new_borrow_pos(cls: Type[_T], type_: str, value: Any, borrow_t: 'Token') -> _T:
         return cls(type_, value, borrow_t.start_pos, borrow_t.line, borrow_t.column, borrow_t.end_line, borrow_t.end_column, borrow_t.end_pos)
 
+    def as_str(self) -> str:
+        """Return the token's value as a ``str``.
+
+        When the parser is created with ``use_bytes=True``, ``value`` is ``bytes``,
+        so ``str(token)`` yields its repr rather than its text. This decodes it,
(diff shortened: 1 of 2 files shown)


erezsh:
I'm on the fence on this one. After all, the user knows they set use_bytes, and should expect the output to be bytes as well. So in what situations would this be used?

PR #1623, opened 2026-07-19
Fix Python comprehension filters after each for clause

### LINKED ISSUE #1525: comp_if should be moved into comp_for to correctly reflect Python's comprehension grammar
**Describe the Bug**

In Python, the following syntax is valid:

```python
[(i, j) for i in range(3) if i > 0 for j in range(5) if j % 2 == 0]
```

However, in the current `python.lark` grammar, the relevant rule is defined as:

```lark
comprehension{comp_result}: comp_result comp_fors [comp_if]
comp_fors: comp_for+
comp_for: [ASYNC] "for" exprlist "in" or_test
```

With this grammar definition, the two `if` conditions cannot be parsed correctly. It only allows cases such as:

```python
[(i, j) for i in range(3) for j in range(5) if j % 2 == 0 if j > 0]
```

In this structure, the `if` clauses are associated globally with the entire comprehension rather than individually with each `for` clause. This is incorrect because, in Python, each `for` clause in a comprehension can have its own set of `if` filters.

**To Reproduce**

I’m unable to provide a simple standalone reproduction script, but as described above, the provided example:

```python
[(i, j) for i in range(3) if i > 0 for j in range(5) if j % 2 == 0]
```

fails to parse correctly under the current grammar. The issue is that the `if` clauses are incorrectly attached to the `comprehension` rule as a whole, rather than being associated with their respective `comp_for`. Each `for` loop in a comprehension should be able to have its own `if` filters.

**Proposed Fix**

In my project, the issue was resolved by modifying the grammar as follows:

```lark
comprehension{comp_result}: comp_result comp_fors
comp_fors: comp_for+
comp_for: [ASYNC] "for" exprlist "in" or_test comp_if*
```

This change allows each `for` clause to optionally have its own `if` filters, which aligns with Python’s actual comprehension syntax.

amkdown-max:
## Summary

- attach zero or more `comp_if` clauses to each `comp_for`
- allow filters to appear between multiple comprehension `for` clauses
- add regression coverage for list, set, dict, generator, and async comprehensions

## Why

Python permits every comprehension `for` clause to have its own sequence of `if` filters. Lark's grammar instead placed one optional filter after the entire `comp_fors` sequence, so valid expressions such as:

```python
[(i, j) for i in range(3) if i > 0 for j in range(5) if j % 2 == 0]
```

failed when the parser reached the second `for`.

This change mirrors CPython's `for_if_clause` structure by keeping each clause's filters with that clause.

Fixes #1525.

## Validation

- `python3 -m tests TestPythonParser`
- `python3 -m tests` (`1311` tests passed, `174` skipped)
- `pre-commit run --all-files --show-diff-on-failure`
- parsed 51 real Python files from `lark/` and `tests/`
- verified the regression with both LALR/contextual and Earley/basic parsers
- verified the synchronous examples are accepted by both CPython and Lark

### DIFF
--- a/lark/grammars/python.lark
+++ b/lark/grammars/python.lark
@@ -252,7 +252,8 @@ kwargs: "**" test ("," argvalue)*
 
 comprehension{comp_result}: comp_result comp_fors
 comp_fors: comp_for+
-comp_for: [ASYNC] "for" exprlist "in" or_test comp_if*
+comp_for: [ASYNC] "for" exprlist "in" or_test comp_ifs
+comp_ifs: comp_if*
 ASYNC: "async"
 ?comp_if: "if" test_nocond
 
--- a/lark/grammars/python.lark
+++ b/lark/grammars/python.lark
@@ -250,9 +250,9 @@ kwargs: "**" test ("," argvalue)*
 ?argvalue: test ("=" test)?
 
 
-comprehension{comp_result}: comp_result comp_fors [comp_if]
+comprehension{comp_result}: comp_result comp_fors
 comp_fors: comp_for+
-comp_for: [ASYNC] "for" exprlist "in" or_test
+comp_for: [ASYNC] "for" exprlist "in" or_test comp_if*
 ASYNC: "async"
 ?comp_if: "if" test_nocond
 
--- a/tests/test_python_grammar.py
+++ b/tests/test_python_grammar.py
@@ -293,6 +293,27 @@ class TestPythonParser(TestCase):
             with self.assertRaises(ParseError):
                 self.python_parser.parse(case, start="file_input")
 
+    def test_comprehensions_with_interleaved_filters(self):
+        clauses = (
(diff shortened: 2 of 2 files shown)


erezsh:
Looks good! Just one nit, I prefer -

```
comp_for: [ASYNC] "for" exprlist "in" or_test comp_ifs
comp_ifs: comp_if*
```

It's more organized and easier to process. What do you think?

amkdown-max:
Agreed — this is clearer and keeps the rule easier to process. Updated in 085eba1. The full test suite (1,311 tests), mypy, and all pre-commit hooks pass. Thanks!

amkdown-max:
Hi @erezsh, just a gentle follow-up: the workflow runs for the latest commit (085eba1) are still awaiting approval. When convenient, could you approve them so the checks can run? No rush, and thanks again!

MegaIng (approved):

erezsh:
Thanks for contributing!

PR #1625, opened 2026-07-19
Fixes for PR 1607 - Reject cyclic grammar

erezsh:
Improves on PR #1607

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -74,6 +74,7 @@ class LarkOptions(Serialize):
     edit_terminals: Optional[Callable[[TerminalDef], TerminalDef]]
     import_paths: 'List[Union[str, Callable[[Union[None, str, PackageResource], str], Tuple[str, str]]]]'
     source_path: Optional[str]
+    allow_cyclic_rules: bool
 
     OPTIONS_DOC = r"""
     **===  General Options  ===**
@@ -154,6 +155,8 @@ class LarkOptions(Serialize):
             A List of either paths or loader functions to specify from where grammars are imported
     source_path
             Override the source of from where the grammar was loaded. Useful for relative imports and unconventional grammar loading
+    allow_cyclic_rules
+            Disable detection of cyclic rules; LALR may loop forever for some grammars. Added for backwards compatibility.
 
     **=== End of Options ===**
     """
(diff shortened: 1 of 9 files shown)


sarathfrancis90:
Read through both commits. Scoping to reachable rules is the right call — I'd been rejecting cycles the parser can never enter, which would have broken grammars that build fine today. Suite is green here (1314 passed).

One asymmetry the new test surfaces: the grammar it asserts is harmless on LALR hangs on CYK.

```python
Lark('start: "x"\na: b\nb: a', parser='lalr')  # builds, per the new test
Lark('start: "x"\na: b\nb: a', parser='cyk')   # hangs in to_cnf -> _unit
```

LALR never walks the unreachable rules, but `cyk.Parser` passes all of them to `to_cnf`, and unit-rule elimination spins on `a: b` / `b: a`. This hangs on master too, so it isn't from your change — my unscoped check just happened to hide it.

Filtering CYK's rules to the reachable set clears it while keeping the reachable-cycle rejection (unreachable builds, `start.1: "a" | start start*` still raises, normal grammars parse). Since `check_cyclic_grammar` already computes that set, it could return it instead of recomputing. Happy to add it here, or leave CYK alone if you'd rather keep this PR to the cycle check.

erezsh:
@MegaIng I added a new option, `allow_cyclic_rules`, because some existing grammars may be cyclic and still work (LALR auto-resolves some situations). That way users can set this flag instead of having to fix the grammar.

What do you think, good idea, or too much?

MegaIng:
If this rejects some grammars that were previously parsed successfully, then this is a breaking change and we shouldn't just add it in a minor version.

An option that defaults to False would have the benefit that it provides an easy transition path, at the cost of not providing the benefit immediately.

The option could select between "error", "warn", "allow" instead, default to "warn". That would give the benefit to most users and not break semver.

erezsh:
Strictly speaking, right now even grammars that loop forever could still compile and parse successfully, as long as the inputs don't activate the loop.

But I agree that warning is a better default, until we're ready to break backwards compatibility.

~~It can default to error when strict=true~~  (hmm maybe that's still a bad idea for the same reason of breaking changes)

PR #1626, opened 2026-07-25
Support backslash line continuations in lark.lark

### LINKED ISSUE #1620: could lark.lark be missing "\" end of line things
**Describe the bug**

I was trying to parse a lark file using the lark.lark file provided in the repo, and I realized it does not like:

``` lark
function: "ABS"i | "ARCTAN"i | "CEIL"i | "COS"i | "ERRORF"i | "EXP"i | "FLOOR"i |\
          "LOG"i | "LOG10"i | "MAPVAL"i | "MAX"i | "MIN"i | "MOD"i | "NORMAL"i
```

I understand it is not a biggie since lark is not bootstrapped, 

**To Reproduce**

Get [lark.lark](https://github.com/lark-parser/lark/blob/master/lark/grammars/lark.lark)

Save this dummy grammar to `parse_me.lark`

``` lark
start: (function? _NL)* function?
_NL: /(\r?\n)+\s*/
function: "ABS"i | "ARCTAN"i | "CEIL"i | "COS"i | "ERRORF"i | "EXP"i | "FLOOR"i |\
          "LOG"i | "LOG10"i | "MAPVAL"i | "MAX"i | "MIN"i | "MOD"i | "NORMAL"i
```

Then parse it using lark.lark

``` python
from lark import Lark

with open("lark.lark", "r") as f:
    lark_grammar = f.read()

parser = Lark(lark_grammar)

with open("parse_me.lark", "r") as f:
    problem = f.read()

parsed_problem = parser.parse(problem)
```

You will get 

```
lark.exceptions.UnexpectedCharacters: No terminal matches '\' in the current parser context, at line 3 col 82

"COS"i | "ERRORF"i | "EXP"i | "FLOOR"i |\
                                        ^
Expected one of: 
	* LSQB
	* TOKEN
	* __ANON_0
	* _VBAR
	* RULE
	* LPAR
	* REGEXP
	* _NL
	* STRING
```

Finding and replacing backward slash followed by a new line with "", is my current workaround, but probably `lark.lark` file needs to be updated.

Sanjays2402:
Fixes #1620.

`lark.lark` (the meta-grammar used to parse Lark grammars) does not accept backslash-newline line continuations, even though grammars parsed by Lark support them. Loading a grammar that uses a line continuation via `Lark.open_from_package('lark', 'lark.lark', ...)` raises `UnexpectedCharacters`.

```
start: "a" \\
       "b"
```

### Change
- Add a `BACKSLASH` terminal matching `\\[ ]*\n` and `%ignore` it, mirroring how `WS_INLINE` and `COMMENT` are handled.
- Add `test_line_breaks_in_lark_lark` covering the case.

### Verification
Run against the new test:
- without the `lark.lark` change: fails with `UnexpectedCharacters`
- with the change: passes

### DIFF
--- a/lark/grammars/lark.lark
+++ b/lark/grammars/lark.lark
@@ -58,5 +58,8 @@ _NL: /(\r?\n)+\s*/
 
 COMMENT: /\s*/ "//" /[^\n]/* | /\s*/ "#" /[^\n]/*
 
+BACKSLASH: /\\[ ]*\n/
+
 %ignore WS_INLINE
 %ignore COMMENT
+%ignore BACKSLASH
--- a/tests/test_grammar.py
+++ b/tests/test_grammar.py
@@ -328,6 +328,14 @@ class TestGrammar(TestCase):
         p.parse('ab')
 
 
+    def test_line_breaks_in_lark_lark(self):
+        # The `lark.lark` meta-grammar must accept line continuations too
+        lark_lark = Lark.open_from_package('lark', 'lark.lark', ('grammars',), parser='lalr')
+        lark_lark.parse(r"""start: "a" \
+                       "b"
+""")
+
+
     def test_symbol_eq(self):
         a = None
         b = Symbol("abc")


erezsh:
Thanks for contributing!

PR #1627, opened 2026-07-25
Fix misaligned error pointer for full-width characters

### LINKED ISSUE #1530: Misaligned error location when input text contain full-width characters
**Describe the bug**

When the input text contains [full-width characters](https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block)#Block), the error location indicator (^) will point at the wrong character because it uses half-width spaces (U+20) only but it should instead matches input characters' widths

**To Reproduce**

Use any CJK characters or [full-width version of latin scripts](https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block)#Block) and intentionally create a grammar error. The indicator (^) will point at the wrong location.

Current and expected behavior shown here:

```
    raise UnexpectedCharacters(stream, i, text_line, text_column, {item.expect.name for item in to_scan},
lark.exceptions.UnexpectedCharacters: No terminal matches '古' in the current parser context, at line 2 col 6
1.菝葀：古代一種象徵祥瑞的草。《廣韻．入聲．末韻》：「菝：菝葀，瑞草。」
     ^  // using U+20 (current)
  　　　^ // using U+20 and U+3000 (expected)
Expected one of: 
	* LPAREN
	* I_LQUOTE
```

yashasvi-srivastava21:
Fixes #1530

`get_context()` pads the `^` error pointer using one regular space
per preceding character. This assumes every character has the same
visual width, which breaks down for full-width (CJK) characters,
causing the pointer to land on the wrong column.

This PR adds a `_padding_for()` helper that checks each character's
Unicode east-asian-width property and substitutes a full-width space
(U+3000) for wide/full-width characters, keeping the pointer aligned
regardless of character width.

### DIFF
--- a/lark/exceptions.py
+++ b/lark/exceptions.py
@@ -21,10 +21,22 @@ def assert_config(value, options: Collection, msg='Got %r, expected one of %s'):
     if value not in options:
         raise ConfigurationError(msg % (value, options))
 def _padding_for(before):
-    return ''.join(
-        '\u3000' if unicodedata.east_asian_width(ch) in ('W' ,'F') else ' '
-        for ch in before.expandtabs()
-    )
+    col = 0
+    result = []
+    for ch in before:
+        if ch == '\t':
+           spaces = 8 - (col % 8)
+           result.append(' ' * spaces)
+           col += spaces
+        else:
+           if unicodedata.east_asian_width(ch) in ('W' ,'F') 
+               result.append('\u3000')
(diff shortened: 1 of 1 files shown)


erezsh:
This PR already tried solving it, but didn't finish: https://github.com/lark-parser/lark/pull/1593

PR #1628, opened 2026-07-26
don't follow symlinks or trust another user's file in FS.open

netliomax25-code:
1. With `cache=True` the cache path is derived only from the grammar, the options and the version strings, so `<tmpdir>/.lark_cache_<user>_<sha256>_3_12.tmp` is predictable to anyone else with access to the temp directory.
2. A file another user leaves at that path reaches `pickle.load` in `lark.py:377` before the sha256 line is compared on 378, and since that digest sits in the same file it is a cache key rather than an authenticity check, so the load is arbitrary code execution as whoever runs the parser.
3. Saving the cache followed a symlink left at the same path and overwrote whatever it pointed at.

Sent the non-atomicwrites branch of `FS.open` through an `O_NOFOLLOW` open that also refuses a file it doesn't own and keeps the cache at 0600, which is roughly what the atomicwrites branch gets from `mkstemp` today. A refused cache just falls through to the existing handlers, so the grammar is parsed normally. Three tests in `tests/test_cache.py` cover the load path, the save path and the file mode; they fail on master and pass here, and the rest of the suite is green with and without `atomicwrites` installed.

### DIFF
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -382,6 +382,11 @@ class Lark(Serialize, Generic[_Return_T]):
                 except FileNotFoundError:
                     # The cache file doesn't exist; parse and compose the grammar as normal
                     pass
+                except PermissionError as e:
+                    # FS.open refused the cache (a symlink, owned by another user, or
+                    # writable by others). This is an intentional refusal, not a crash,
+                    # so log a single line and rebuild from the grammar as normal.
+                    logger.warning("Not loading Lark from cache: %s. We will rebuild it.", e)
                 except Exception: # We should probably narrow done which errors we catch here.
                     logger.exception("Failed to load Lark from cache: %r. We will try to carry on.", cache_fn)
 
--- a/lark/lark.py
+++ b/lark/lark.py
@@ -386,7 +386,7 @@ class Lark(Serialize, Generic[_Return_T]):
                     # FS.open refused the cache (a symlink, owned by another user, or
                     # writable by others). This is an intentional refusal, not a crash,
                     # so log a single line and rebuild from the grammar as normal.
-                    logger.warning("Not loading Lark from cache: %s. We will rebuild it.", e)
+                    logger.warning("Failed to load cache due to a permissions error: %s. Rebuilding from the grammar.", e)
                 except Exception: # We should probably narrow done which errors we catch here.
                     logger.exception("Failed to load Lark from cache: %r. We will try to carry on.", cache_fn)
 
@@ -490,8 +490,11 @@ class Lark(Serialize, Generic[_Return_T]):
                     f.write(cache_sha256.encode('utf8') + b'\n')
                     pickle.dump(used_files, f)
                     self.save(f, _LOAD_ALLOWED_OPTIONS)
+            except PermissionError as e:
+                # FS.open refused the cache path (a symlink, or owned by another user).
+                logger.warning("Failed to save cache due to a permissions error: %s", e)
             except IOError as e:
-                logger.exception("Failed to save Lark to cache: %r.", cache_fn, e)
+                logger.exception("Failed to save Lark to cache: %r. (%s)", cache_fn, e)
 
     if __doc__:
         __doc__ += "\n\n" + LarkOptions.OPTIONS_DOC
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -308,7 +308,8 @@ except ImportError:
 
 def _open_private(name, mode, **kwargs):
     """Like open(), but refuses to follow a symlink, refuses a file that belongs to
-    another user, and keeps the file readable only by its owner.
+    another user, refuses to read a file that other users can access, and keeps the
+    file readable only by its owner.
 
     The parser cache is stored in a shared temporary directory under a name that is
     derived from the grammar, so another user on the same machine can predict it and
@@ -322,8 +323,14 @@ def _open_private(name, mode, **kwargs):
         flags |= getattr(os, "O_BINARY", 0)
     fd = os.open(name, flags | getattr(os, "O_NOFOLLOW", 0), 0o600)
     try:
-        if hasattr(os, "geteuid") and os.fstat(fd).st_uid != os.geteuid():
-            raise PermissionError("Refusing to use %r: it belongs to another user" % name)
+        st = os.fstat(fd)
+        if hasattr(os, "geteuid"):
+            if st.st_uid != os.geteuid():
+                raise PermissionError("Refusing to use %r: it belongs to another user" % name)
+            # On read, don't trust the contents of a file that group or others can
+            # write, since it could have been tampered with even though we own it.
+            if "w" not in mode and st.st_mode & 0o077:
+                raise PermissionError("Refusing to read %r: it is accessible to other users" % name)
         if "w" in mode and hasattr(os, "fchmod"):
             os.fchmod(fd, 0o600)
         return os.fdopen(fd, mode, **kwargs)
--- a/lark/utils.py
+++ b/lark/utils.py
@@ -305,6 +305,33 @@ try:
 except ImportError:
     _has_atomicwrites = False
 
+
+def _open_private(name, mode, **kwargs):
+    """Like open(), but refuses to follow a symlink, refuses a file that belongs to
+    another user, and keeps the file readable only by its owner.
+
+    The parser cache is stored in a shared temporary directory under a name that is
+    derived from the grammar, so another user on the same machine can predict it and
+    get there first.
+    """
+    if "w" in mode:
+        flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+    else:
+        flags = os.O_RDONLY
+    if "b" in mode:
+        flags |= getattr(os, "O_BINARY", 0)
+    fd = os.open(name, flags | getattr(os, "O_NOFOLLOW", 0), 0o600)
+    try:
+        if hasattr(os, "geteuid") and os.fstat(fd).st_uid != os.geteuid():
+            raise PermissionError("Refusing to use %r: it belongs to another user" % name)
+        if "w" in mode and hasattr(os, "fchmod"):
+            os.fchmod(fd, 0o600)
+        return os.fdopen(fd, mode, **kwargs)
+    except Exception:
+        os.close(fd)
+        raise
+
+
 class FS:
     exists = staticmethod(os.path.exists)
 
@@ -313,7 +340,7 @@ class FS:
         if _has_atomicwrites and "w" in mode:
             return atomicwrites.atomic_write(name, mode=mode, overwrite=True, **kwargs)
         else:
-            return open(name, mode, **kwargs)
+            return _open_private(name, mode, **kwargs)
 
 
 class fzset(frozenset):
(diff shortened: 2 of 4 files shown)


MegaIng:
Is this standard practice in other projects using cache files?

Would it be easier to put the cache files somewhere else?

netliomax25-code:
Both fair questions.

1. On prior art: it's what CPython does for its own temp files. `tempfile.mkstemp` opens with `O_RDWR|O_CREAT|O_EXCL|O_NOFOLLOW` at mode 0600 (`tempfile.py`, `_text_openflags` and `_mkstemp_inner`). Lark already inherits that on one path: with the `atomic_cache` extra installed, `FS.open` hands writes to `atomicwrites.atomic_write`, which creates the file through `mkstemp` and renames it into place, so that branch is already 0600 and already symlink-proof. This PR mostly gives the plain-`open()` fallback the same properties.

2. The gap that made me file it is that reads never reach that branch. `FS.open` only routes to atomicwrites when `"w" in mode`, so `'rb'` is a plain `open()` either way, and that's the call feeding `pickle.load` on lark.py:377 (the sha256 line is compared on 378, after that load). I re-checked on master just now: with atomicwrites installed, a pickle planted at the predictable cache path still runs on load; without the extra, the save path also follows a symlink and overwrites whatever it points at. On this branch both are refused and the grammar is parsed normally.

   The exposure is platform-dependent, which may be why it isn't obvious: on macOS `gettempdir()` is a per-user `/var/folders/.../T` at 0700, but on Linux and CI it's `/tmp` at 1777.

3. On moving the files, yes, and I'd say that's the better fix if you want to go that way. A directory only the user can write to removes the cause instead of defending against it, and it's the norm: pip, black, pre-commit, Cython and huggingface_hub all default to a per-user cache dir, and matplotlib only falls back to the temp dir when that's unwritable, and even then uses `mkdtemp` at 0700 and warns. Two things to weigh: lark has no runtime dependencies, so it'd mean hand-rolling `XDG_CACHE_HOME`/`~/.cache` plus `%LOCALAPPDATA%` rather than taking platformdirs, and it changes the default cache location, so files already in the temp dir get orphaned and the new directory wants creating at 0700.

The two aren't exclusive, and they're not the same size of change. I'm happy to convert this PR to a per-user cache directory, or leave it as the narrow fix and do the relocation separately. Whichever you'd rather carry.

Validation on the branch: full suite green with and without atomicwrites installed (1315 tests), CI passing.

netliomax25-code:
any preference?

MegaIng:
I think we should consider doing both. The change from this PR on it's own is a good idea already.

MegaIng:
The current implementation does not refuse to open files that have incorrect permissions for reading, it just changes them. Is this intentional?

And don't worry about cache invalidation, a new lark version is going to invalidate all caches anyway.

netliomax25-code:
1. On the read path it wasn't intentional, and you're right that it's the wrong default. Reads only checked ownership, so a file we own but that group or others can write could have been tampered with by another user and we'd still pickle.load it. Changing perms there doesn't help either, since the bad contents are already on disk.

2. Fixed it to refuse instead: on read, `_open_private` now rejects a cache whose group/other bits are set (`st_mode & 0o077`) rather than trusting or rewriting it. A refused cache falls through to the existing handler and the grammar is rebuilt, and since 0600 is what we write, only stale or planted files get refused. Good point on invalidation too, so I didn't add any migration for older 0644 caches; they'll just be rebuilt.

The write path still tightens to 0600 rather than refusing, which seems right since it truncates and we own it.

Added `test_load_refuses_group_or_world_accessible` alongside the existing three. Full suite green with and without atomicwrites installed.

On the relocation: agreed, I'll keep this as the narrow fix and open the per-user cache directory as a separate PR.

MegaIng (approved):

MegaIng:
@erezsh I can't approve workflows AFAICT.

erezsh:
@MegaIng I changed your role from “triage” to “write". Let me know if it still doesn't work.

MegaIng:
@erezsh Works, thank you. I am going to leave merging to you, but I think this PR is ready.

erezsh:
Thanks, I haven't had time to look into it yet, but I'll do it soon.

erezsh:
Fable had some comments, and at a glance they seem right. I would appreciate if you could address them.

1. O_TRUNC runs before the ownership check — _open_private, lark/utils.py:319 on the branch. In the non-atomicwrites write path the open truncates the target and only then fstats and refuses it. I reproduced it: a group-writable cache owned by nobody went from 16 bytes → 0 bytes, then PermissionError: ... belongs to another user. For a legitimately shared explicit cache="path" that means we destroy a colleague's cache and then don't replace it. Fix: drop O_TRUNC from the flags and call os.ftruncate(fd, 0) after the checks pass.

2. Refusals surface as ERROR + full traceback. The new PermissionErrors land in except Exception: logger.exception(...) at lark/lark.py:393, so an intentional "won't read this" looks like a crash. Concretely, anyone with an explicit cache="path" file written by the current release (0644 under default umask) gets a traceback on first load after upgrading — before this PR the same situation (version mismatch) fell through silently. (cache=True is unaffected since the version is part of the filename.) Suggest catching PermissionError separately in lark.py and logging a one-line warning, or raising a dedicated exception lark.py can recognise.

3. The read check is stricter than the threat. st.st_mode & 0o077 refuses group/world-readable files, but only write bits let another user tamper; 0o022 matches the stated rationale ("could have been tampered with"). More important: on filesystems that don't honour fchmod (vfat, WSL /mnt/c, some Docker bind mounts, NFS with squash) files report 0777/0755, so an explicit cache="path" there becomes permanently non-functional — reparsed on every Lark() with a traceback each time under the current logging. That's the scenario where point 2 turns from cosmetic to painful.

4. No test coverage of _open_private on Windows. 

(I'm not sure 2 is a bad thing, but worth keeping in mind.)

netliomax25-code:
Thanks, all four look right. Addressed them:

1. O_TRUNC before the ownership check. Dropped O_TRUNC from the write flags and moved to os.ftruncate(fd, 0) after the checks pass, so a file we end up refusing is left intact instead of emptied first. Confirmed on the plain-open path: a 0600 cache faked as owned by another uid now raises before any truncation and its contents are untouched.

2. Refusals surfacing as ERROR + traceback. lark.py now catches PermissionError separately from the load and logs a one-line warning, then rebuilds as normal. I also translated the O_NOFOLLOW symlink error (ELOOP, EMLINK on some BSDs) into a PermissionError inside _open_private, so a planted symlink takes the same quiet path rather than a traceback. The version-mismatch-on-upgrade case you described now logs 'Not loading Lark from cache: ... We will rebuild it.' and parses fine.

3. Read check stricter than the threat. Narrowed the mask from 0o077 to 0o022, so only group/other write bits are refused and a file we own that is merely group/world readable loads normally. On the fchmod-less filesystems (vfat, WSL /mnt/c, some bind mounts) an explicit cache path is still refused, but with point 2 that is now a one-line warning and a rebuild each time rather than a traceback, so it stays functional just uncached. Added a test that a 0644 file we own is accepted and one that a 0666 file is refused.

4. Windows coverage. Added TestCacheFilePortable (not gated on posix) that round-trips a write/read through FS.open and checks truncate-on-write, so _open_private is exercised on Windows where the ownership and symlink checks are no-ops.

Validation: full suite green with and without atomicwrites installed (1320 tests), mypy clean on the changed files.

erezsh:
I made small changes. Let me know if you agree @MegaIng & @netliomax25-code

netliomax25-code:
Looks good to me. The save-path PermissionError branch is the half I'd only done on the load side, so catching it there is the right symmetry, and the reworded warnings are clearer. I pulled the branch and ran the full suite green (1321 tests, with and without atomicwrites), and the new test passes both ways. Happy for it to go in.

netliomax25-code:
any update?

erezsh:
A question that just occurred to me, should we make it possible to disable this protection?

Because some users might not care about it, and might not need it, while it could break some legitimate workflows. (at least, legitimate in the context of personal use)

netliomax25-code:
I think yes, as long as it stays on by default and the opt-out is explicit.

1. Where it can bite: only an explicit `cache="path"`. With `cache=True` lark creates the file itself at 0600 under a name that includes the username, so the checks shouldn't fire on a file it wrote and there's nothing to opt out of. The cases I can see breaking are all explicit paths: a cache baked into a Docker image as root and loaded by a non-root user, a cache file shared between accounts, and a path on a filesystem that reports 0777 for everything (vfat, WSL `/mnt/c`, some bind mounts).

2. What happens there today: I checked the first and third on this branch (faking the uid and the mode bits). The parser still builds and parses fine, but the cache is refused on every construction, so it reparses each time. Since lark's logger defaults to CRITICAL the warning isn't visible either, so from the user's side the cache just quietly stops helping. That seems like a good enough reason to give them a switch.

3. Shape I'd suggest: a boolean Lark option, default on (something like `cache_strict=True`, naming is yours), that makes `FS.open` fall back to a plain `open()` when turned off, with a line in the option docs saying what it skips. I'd keep the checks unconditional for `cache=True`, since that's the predictable path in a shared temp dir and nothing legitimate gets refused there. There's precedent for the pattern: sshd's `StrictModes` and git's `safe.directory` are both ownership/mode checks that default to on and can be turned off explicitly.

I can add that to this PR with a test for both settings, or it could go in separately so this one isn't held up further. The default-on behavior here doesn't change either way.

erezsh:
I think `unsafe_cache = False` is a good name. WDYT?

MegaIng:
I don't quite like using "safe" as a term for a functionality that still loads arbitrary data. This doesn't fix the attack vector, it just makes it harder to use.

erezsh:
Maybe unguarded_cache is a better term? Anyway, I'm open to suggestions.

netliomax25-code:
`unguarded_cache` works for me.

1. On the name: I agree with MegaIng that "safe"/"unsafe" would oversell it, since the cache is still a pickle and the checks only narrow who can plant one. "Unguarded" says what gets switched off without claiming the default is safe, and `False` reads naturally as the default. I don't have a better candidate.

2. I pushed it under that name so there's something concrete to look at; it's a single rename if you settle on something else. `unguarded_cache=False` keeps everything in this PR as it is. With `True`, `FS.open` falls back to the plain `open()` it used before (writes still go through atomicwrites when that's installed). The option doc lists what gets skipped and says the cache is a pickle, so it should only be set when everyone who can write to the path is trusted.

3. Two things differ from what I sketched earlier:
   - It's left out of the cache key, since it only changes how the file is opened and not what's in it. Otherwise turning it on to get an existing cache accepted (say one baked into an image as root) would miss on the hash and rebuild anyway.
   - I'd suggested keeping the checks unconditional for `cache=True`. I dropped that: a temp dir on a filesystem that reports 0777 would have no way out, and a switch that only half applies seemed more surprising than useful. Easy to restrict to explicit paths if you'd rather.

Validation: the new tests cover both settings (a 0666 cache is refused and rebuilt by default, and loaded with `unguarded_cache=True`), plus a round trip that also runs on Windows. I also checked the other-owner case by faking the uid: the default rebuilds, unguarded loads from the cache. Full suite green with and without atomicwrites (1324 tests), mypy clean.

PR #1631, opened 2026-07-29
Add support for a (ASCII) and L (locale) regex flags

### LINKED ISSUE #1527: Add support for the `a` and `L` regex flags

Abhi-DevHub:
Closes #1527

Adds \\ (re.ASCII) and \L\ (re.LOCALE) to the supported regex flags in Lark grammar patterns.

### DIFF
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -122,7 +122,7 @@ SIGNED_INTEGER: /
  /x
 ```
 
-Supported flags are one of: `imslux`. See Python's regex documentation for more details on each one.
+Supported flags are one of: `imsluxaL`. See Python's regex documentation for more details on each one.
 
 Regexps/strings of different flags can only be concatenated in Python 3.6+
 
--- a/docs/grammar.md
+++ b/docs/grammar.md
@@ -122,7 +122,7 @@ SIGNED_INTEGER: /
  /x
 ```
 
-Supported flags are one of: `imsluxaL`. See Python's regex documentation for more details on each one.
+Supported flags are one of: `aiLmsux`. See Python's regex documentation for more details on each one.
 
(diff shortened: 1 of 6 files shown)


MegaIng:
Thank you for your interested in developing lark.

There is also #1589. That PR includes a bit more changes and importantly tests, but it's author has not been active on it for 3 months. If you want to extend this PR to include tests that would be great.

Abhi-DevHub:
Thanks @MegaIng! I've extended the PR with tests mirroring #1589's approach:

- \	est_token_flags_ascii\ — verifies the a flag restricts \\\\w to ASCII (rejects é)
- \	est_token_flags_locale_bytes\ — verifies the L flag works with \use_bytes=True\

Testing uncovered a related latent bug: with \use_bytes=True\, regexps were being validated as str patterns, so the L flag (bytes-only per Python) failed at validation time. Added the fix to \lexer.py\ — validation now encodes the pattern to bytes when \use_bytes=True\.

All tests pass: 1064 passed (test_parser, test_lexer, test_grammar). The tests run across all parser algorithms (EARLEY/CYK/LALR and variants).

docs/grammar.md
```diff
@@ -122,7 +122,7 @@ SIGNED_INTEGER: /
  /x
 ```
 
-Supported flags are one of: `imslux`. See Python's regex documentation for more details on each one.
+Supported flags are one of: `imsluxaL`. See Python's regex documentation for more details on each one.
```
MegaIng: Lets use the ordering from the official docs, `aiLmsux` (and use it consistently everywhere)

MegaIng (changes requested):
Only one last change, otherwise looks good.

MegaIng (approved):

PR #1632, opened 2026-07-30
Fix Template.search() dropping zero-variable matches

### LINKED ISSUE #1630: Fix Template.search() to yield zero-variable matches
Fix Template.search() in lark/tree_templates.py so it correctly yields matches when the template contains no template variables. Today, Template.match() returns an empty dict when a pattern with no variable placeholders successfully matches a subtree, which is the correct sentinel for "match found, nothing to bind". But search() checks the result with a plain truthiness test, and an empty dict is falsy in Python, so every zero-variable match is silently dropped. As a result, search() always returns an empty iterator when the template has no variables, and translate() never replaces anything in that case, even when the source pattern appears in the tree. The fix must distinguish between None (no match) and {} (match with no variable bindings), which is already the contract that match() documents and tests confirm.

Two existing tests in tests/test_tree_templates.py were written to match the broken behavior and assert an empty result for zero-variable templates; they must be corrected to assert the match that the fixed implementation correctly yields.

Sanjays2402:
Closes #1630

`Template.match()` returns `{}` for a match with no variables to bind and `None` for no match, but `search()` tested the result for truthiness — so `{}` was read as "no match" and every zero-variable match was dropped, leaving `translate()` a no-op for variable-free templates. `search()` now tests `res is not None`.

The two existing search tests asserted the broken empty result and are corrected; added a test that a genuinely different tree still yields nothing, plus a `TemplateTranslator` case for a variable-free translation. The three tests fail on master and pass with the fix; full suite is 1134 passed / 174 skipped.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/lark/tree_templates.py
+++ b/lark/tree_templates.py
@@ -143,7 +143,7 @@ class Template:
         tree = self.conf._get_tree(tree)
         for subtree in tree.iter_subtrees():
             res = self.match(subtree)
-            if res:
+            if res is not None:
                 yield subtree, res
 
     def apply_vars(self, vars: Mapping[str, Tree[str]]) -> Tree[str]:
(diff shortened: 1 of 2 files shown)


erezsh:
Thanks for contributing!

PR #1633, opened 2026-08-01
group alternations and escape range endpoints in terminal patterns

netliomax25-code:
1. `TerminalTreeToPattern.expansion` joins sub-patterns by plain concatenation, so a regexp literal holding a top-level `|` binds looser than the join: `T: /admin|user/ "_id"` compiles to `admin|user_id`, which rejects `admin_id` and accepts a bare `admin`.
2. `PrepareLiterals.range` splices the endpoints' raw source text into the character class, so an endpoint that is itself a metacharacter escapes it: `T: "^".."z"` compiles to `[^-z]`, a negated class matching tab and digits while rejecting `z`.

Either way a terminal written as a restriction matches outside the language it declares, which is the part that worries me when the grammar is what validates untrusted input.

Grouping is applied only when the item carries a `|` outside any group, so an escaped `\|`, one inside a character class, and an already-parenthesized one are left alone. Range endpoints are unescaped and then re-escaped for a class, kept ascii because the pattern may still be encoded to bytes under `use_bytes` (the dynamic lexer encodes utf-8, the basic one latin-1). The other two spellings of that alternation, `/a|b/i` and `("a"|"b")`, already went through `(?:...)`, so only the third one changes.

Compiled patterns for `common.lark`, `python.lark` and `lark.lark` come out identical to master. Full suite green, mypy clean, and the two tests added in `tests/test_grammar.py` fail on master and pass here.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -605,7 +605,7 @@ def _char_to_regexp(char: str) -> str:
     codepoint = ord(char)
     if char.isascii():
         return re.escape(char)
-    return '\\x%02x' % codepoint if codepoint <= 0xff else '\\U%08x' % codepoint
+    return f'\\x{codepoint:02x}' if codepoint <= 0xff else f'\\U{codepoint:08x}'
 
 
 @inline_args
@@ -626,28 +626,6 @@ def _make_joined_pattern(regexp, flags_set) -> PatternRE:
     return PatternRE(regexp, ())
 
 
-def _has_bare_alternation(regexp: str) -> bool:
-    """Whether ``regexp`` has a ``|`` outside of any group, i.e. one that isn't already
-    bound by surrounding parentheses.
-    """
-    depth = 0
-    in_class = False
-    chars = iter(regexp)
-    for c in chars:
-        if c == '\\':
-            next(chars, None)
-        elif in_class:
-            in_class = c != ']'
-        elif c == '[':
-            in_class = True
-        elif c == '(':
-            depth += 1
-        elif c == ')':
-            depth -= 1
-        elif c == '|' and not depth:
-            return True
-    return False
-
 class TerminalTreeToPattern(Transformer_NonRecursive):
     def pattern(self, ps):
         p ,= ps
@@ -660,10 +638,9 @@ class TerminalTreeToPattern(Transformer_NonRecursive):
         if len(items) == 1:
             return items[0]
 
-        # A bare '|' binds looser than concatenation, so an item carrying one has to be grouped
-        # before it's joined, or it swallows its neighbors: /a|b/ "c" must not become 'a|bc'.
-        regexps = [i.to_regexp() for i in items]
-        pattern = ''.join('(?:%s)' % r if _has_bare_alternation(r) else r for r in regexps)
+        # A bare '|' binds looser than concatenation, so each item is grouped before
+        # it's joined, or it swallows its neighbors: /a|b/ "c" must not become 'a|bc'.
+        pattern = ''.join(f'(?:{i.to_regexp()})' for i in items)
         return _make_joined_pattern(pattern, {i.flags for i in items})
 
     def expansions(self, exps: List[Pattern]) -> Pattern:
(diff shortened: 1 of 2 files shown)


lark/load_grammar.py
```diff
@@ -625,7 +660,10 @@ def expansion(self, items: List[Pattern]) -> Pattern:
         if len(items) == 1:
             return items[0]
 
-        pattern = ''.join(i.to_regexp() for i in items)
+        # A bare '|' binds looser than concatenation, so an item carrying one has to be grouped
+        # before it's joined, or it swallows its neighbors: /a|b/ "c" must not become 'a|bc'.
+        regexps = [i.to_regexp() for i in items]
+        pattern = ''.join('(?:%s)' % r if _has_bare_alternation(r) else r for r in regexps)
```
MegaIng: We can just unconditionally use an anonymous group, trying to manually check if there is an alternation is a recipe for infinite edge cases.

lark/load_grammar.py
```diff
@@ -596,23 +597,57 @@ def _literal_to_pattern(literal):
...
 
+def _char_to_regexp(char: str) -> str:
+    """Escape a single character for use inside a regexp character class.
+
+    The result stays ascii, because the pattern may later be encoded to bytes (``use_bytes``).
+    """
+    codepoint = ord(char)
+    if char.isascii():
+        return re.escape(char)
+    return '\\x%02x' % codepoint if codepoint <= 0xff else '\\U%08x' % codepoint
```
MegaIng: Lets use f-strings here.

MegaIng (changes requested):
While I agree that this PR implements the intended behavior, this is a breaking change.

People may have (accidentally) relied on this behavior to construct more complex terminals. But I suspect this is rare enough that we can put this off as a bug fix.

lark/load_grammar.py
netliomax25-code: Done, dropped the helper and the join now wraps every item in (?:...) unconditionally. One consequence worth flagging: the compiled patterns for the bundled grammars are no longer byte-identical to master (they pick up redundant groups), just equivalent. Full suite and mypy still green.

lark/load_grammar.py
netliomax25-code: Done, switched to f-strings.

netliomax25-code:
Agreed on the framing. Anyone relying on the old splicing was depending on the terminal matching outside what its own definition says, so treating it as a bug fix seems right to me. Both review points are addressed in the latest push: the join groups every item unconditionally (the alternation scanner is gone), and the escape helper uses f-strings.

erezsh:
Overall looks good, the bugs are real and the fix is in the right direction.

A few things to address:

1. We should not emit \U… escapes, because interegular can't parse them . Instead it drops the terminal from every collision check. It also breaks anonymous-terminal dedup, which compares pattern strings: `start: "а".."я" T` with `T: /[а-я]/ reused T` on master; on the PR it creates a duplicate.
The use_bytes rationale doesn't apply — a bytes pattern rejects \U anyway, and use_bytes already requires an ASCII grammar. Plain re.escape(char) is all that's needed (it leaves non-ASCII alone), so the helper can go. But I might be wrong.

2. We should still test if `|` at least appears in the regex, to avoid unnecessary grouping. It would filter out most of the regexps.

netliomax25-code:
Both addressed in the latest push.

1. Agreed on `\U`, it's gone. Plain `re.escape` alone turned out to be almost enough, but not quite: for endpoints in 0x80-0xff a literal character breaks `use_bytes`, because the dynamic lexer encodes the pattern utf-8 and a literal `'\xf0'` splits into two bytes (`test_bytes_utf8` fails that way, its CHAR ranges are written in the `"\x80".."\xbf"` style). So the helper is now just: `\xNN` for that band, `re.escape` for everything else, which leaves anything above 0xff literal. interegular parses `\xNN` fine, and your dedup example works: `"а".."я"` compiles to `[а-я]` and dedups with `/[а-я]/`. I added an assertion for that to the range test.

2. The join now wraps an item only when its regexp contains a `|` (plain substring check), so most regexps are left untouched.

Full suite and mypy are green.

netliomax25-code:
any update here?

PR #1635, opened 2026-08-12
refactor(grammar): modernize string formatting with f-strings and improve type hinting

ardozcnn:
Replaced legacy % string formatting across Symbol, Terminal, RuleOptions, and Rule classes with modern Python f-strings.

Added missing type hints for method parameters (other: Any, memo: Optional[Dict], f: Callable) and return types (-> None, -> str, -> int).

Included empty_indices in RuleOptions.__repr__ to accurately reflect all initialization parameters.

Formatted tuple entries in __serialize_fields__ consistently.

### DIFF
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -29,7 +29,9 @@ class Symbol(Serialize):
     def __repr__(self) -> str:
         return f'{type(self).__name__}({self.name!r})'
 
-    fullrepr: property = property(__repr__)
+    @property
+    def fullrepr(self) -> str:
+        return repr(self)
 
     def renamed(self, f: Callable[[str], str]) -> 'Symbol':
         return type(self)(f(self.name))
--- a/lark/grammar.py
+++ b/lark/grammar.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional, Tuple, ClassVar, Sequence
+from typing import Any, Dict, Optional, Tuple, ClassVar, Sequence, Callable
 
 from .utils import Serialize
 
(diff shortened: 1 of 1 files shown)


erezsh:
Overall looks okay. You can use tox to run the type checker with the official configuration.

ardozcnn (commented):
changed

ardozcnn:
I ran the code can you check?

erezsh:
Opus says (and I agree):

1. NonTerminal.serialize — the memo hint is wrong. memo is never a dict; it's a SerializeMemoizer (see Serialize.memo_serialize, and _serialize in utils.py, which already types it that way). mypy doesn't catch it because the base method leaves memo implicitly Any. Please use Optional[SerializeMemoizer], or Any to match lalr_parser.py.

2. renamed should return Self, not Symbol. The body is type(self)(...), so NonTerminal.renamed() returns a NonTerminal. Annotating it Symbol makes correct code fail to typecheck — e.g. rule.origin = rule.origin.renamed(f) errors with "expression has type Symbol, variable has type NonTerminal". It slips through today only because the single caller (load_grammar.py:1053) is in an untyped context.

We support Python 3.8 and have no runtime dependencies, so import it the way common.py already does:

if TYPE_CHECKING:
    if sys.version_info >= (3, 11):
        from typing import Self
    else:
        from typing_extensions import Self

ardozcnn:
Thanks for the detailed review! I've updated renamed() to return Self (handled via TYPE_CHECKING for Python 3.8 compatibility) and changed memo in NonTerminal.serialize to Any as suggested.

erezsh:
You accidentally deleted the `###}` comments that are needed for the standalone parser.

ardozcnn:
Sorry about that! I've restored the ###} comment for the standalone parser. Thanks for the review

ardozcnn:
Have you checked?

ardozcnn:
I've restored the ###} comment for the standalone parser.

ardozcnn:
It is now passing all tests.

PR #1636, opened 2026-08-12
Add pyparsing porting guide

### LINKED ISSUE #1008: pyparsing porting guide
It would be great to see a tutorial on how to port a pyparsing based project to Lark.

1cbyc:
## Summary
- add a new guide for porting pyparsing parsers to Lark
- cover common pyparsing-to-Lark equivalents, parse actions to transformers, ignored text, delimited lists, precedence, and `Lark.scan()`
- add the guide to the Tutorials and Guides toctree

Closes #1008

## Verification
- Ran the Python snippets for the minimal parser and `Lark.scan()` examples locally.
- Built docs with:
  `uv run --with-requirements docs/requirements.txt sphinx-build -b html docs docs/_build/html`

Notes: full docs build succeeds with existing unrelated warnings in generated example/composition pages and other current docs. The new `pyparsing_porting` page produced no warnings after adjustment.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,6 +23,7 @@ Welcome to Lark's documentation!
    json_tutorial
    how_to_use
    how_to_develop
+   pyparsing_porting
    recipes
    examples/index
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -96,6 +96,7 @@ Resources
 
    -  :doc:`how_to_use`
    -  :doc:`how_to_develop`
+   -  :doc:`pyparsing_porting`
 
 -  Reference
 
--- a/docs/pyparsing_porting.md
+++ b/docs/pyparsing_porting.md
@@ -1,0 +1,221 @@
+# Porting from pyparsing to Lark
+
+This guide is for projects that already have a working parser built with
+[pyparsing](https://pyparsing-docs.readthedocs.io/) and want to move it to
+Lark incrementally.
+
+The main change is architectural: pyparsing grammars are Python object graphs,
+while Lark grammars are usually written as EBNF text. In return, Lark can build
+a parse tree automatically, run different parser algorithms from the same
+grammar, and keep grammar and tree-processing code separate.
+
+## Porting checklist
+
+1. **Choose one representative input** from the pyparsing parser's tests.
+2. **Write the Lark grammar for only that input shape**, not the whole language.
+3. **Parse to a tree first**, before adding a transformer.
+4. **Move parse actions to a `Transformer`** once the tree shape is stable.
+5. **Add the rest of the syntax gradually**, keeping each pyparsing test paired
+   with an equivalent Lark test.
+
+## pyparsing concepts and Lark equivalents
+
+| pyparsing | Lark |
+| --- | --- |
+| `Literal("+")`, `Keyword("if")` | string terminals: `"+"`, `"if"` |
+| `Word(alphas)`, `Word(nums)` | regex terminals: `/[a-zA-Z]+/`, `/\d+/` or `%import common.INT` |
+| `expr + term` | sequence: `expr term` |
+| `expr | term` / `MatchFirst` | ordered alternatives usually become `expr \| term` in a rule |
+| `Optional(expr)` | `expr?` |
+| `ZeroOrMore(expr)` | `expr*` |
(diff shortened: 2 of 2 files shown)


erezsh:
Thanks, looks good! (granted, I'm not too familiar with Pyparsing)

I ran Fable on it, and it had a lot of useful comments. You don't have to address all of them, but definitely some of them are important.

I'll paste the output here:

```md
 ## Review: PR #1636 — pyparsing porting guide

Ran every code snippet against the current checkout. Four of five are correct; one is broken.

### Must fix

**1. The `%ignore`/COMMENT example doesn't work.**
`start: assignment+` with `VALUE: /[^\n#]+/` and `%ignore WS_INLINE` can only ever match one assignment:

- `"width=10 height=20"` → one assignment with `VALUE = '10 height=20'` (VALUE eats to end of line)
- `"width=10\nheight=20"` → `UnexpectedCharacters` (newline is neither ignored nor consumed)
- `"width=10 # comment"` → works, but `VALUE = '10 '` with a trailing space

Since the section is about `%ignore COMMENT`, simplest fix is to reuse `INT` and `%ignore WS` from the previous example:

    start: assignment+
    assignment: NAME "=" INT

    COMMENT: /#[^\n]*/

    %import common.CNAME -> NAME
    %import common.INT
    %import common.WS
    %ignore WS
    %ignore COMMENT

(Verified: parses `"width=10 # c\nheight=20"` into two assignments.)

**2. `Suppress(expr)` row is inaccurate.**
"literal strings or terminals whose values are not needed" — named terminals are always kept in the tree. Only anonymous literals (`"="`) are filtered automatically; terminals/rules must be prefixed with `_` to be dropped/inlined. The body text later says this correctly; the table should match.

### Should fix

**3. `\|` renders with a literal backslash.**
Through `python-markdown`'s `tables` extension (what `sphinx_markdown_tables` uses), `` `expr \| term` `` renders as `expr \| term`. The unescaped `` `expr | term` `` in the first column already renders fine — just drop the backslash.

**4. `expr | term` / `MatchFirst` → "ordered alternatives".**
Lark's `|` is *not* ordered (LALR resolves deterministically, Earley by priority). Worth one sentence — `MatchFirst` chains that rely on order will bite people on port.

**5. Whitespace gotcha isn't called out.**
pyparsing skips whitespace by default; Lark doesn't. Every example uses `%ignore WS` but never says why. A one-line callout near the top would prevent the most common first-run failure.

### Nice to have

- Add `Forward()` → recursive rule to the table; it's in nearly every non-trivial pyparsing grammar and the mapping isn't obvious. Also useful: `Regex(...)` → `/.../`, `QuotedString` → `common.ESCAPED_STRING`, `CaselessKeyword` → `"..."i`, `setResultsName` → `-> alias`.
- `infixNotation` → "or Lark's grammar composition patterns" — composition is `%import`/multi-file grammars, unrelated to precedence. Just point to the precedence section below.
- Mixed pyparsing naming: table uses legacy `setParseAction`/`delimitedList`, examples use pyparsing-3 `parse_string`/`as_list`. Pick one.
- `docs/index.rst` also has a visible "Guides" bullet list (`how_to_use`, `how_to_develop`) besides the hidden toctree — add `:doc:\`pyparsing_porting\`` there too.
- `scan()` section is accurate; could mention in one clause that it's LALR-only.
```

1cbyc:
Addressed the review feedback that looked actionable here:

- fixed the `%ignore COMMENT` example to use `INT` and `%ignore WS`
- clarified that Lark whitespace ignores must be explicit
- corrected the `Suppress(expr)` wording for literals vs named terminals/rules
- clarified `MatchFirst`/alternative ordering differences
- added common mappings for `Forward`, `Regex`, `QuotedString`, `CaselessKeyword`, and `setResultsName`
- added the guide to the visible Guides list in `docs/index.rst`

Verification run locally:
- `python3 - <<'PY' ... PY` snippet check for the guide examples -> passed
- `uv pip install --python .venv-docs/bin/python -q -e . -r docs/requirements.txt` -> passed
- `python -m sphinx -b html docs /tmp/lark-docs-html` -> build succeeded with existing warnings outside `docs/pyparsing_porting.md`

erezsh:
Thanks for contributing!

PR #1637, opened 2026-08-13
Update Python versions and add a windows run for tests workflow

MegaIng:
This updates the versions on which we run to include 3.14 and the preview version of 3.15. It also runs a proper test run on windows which was a signal I was missing from #1628.

It may also make sense to run codecov on pull_request instead of on push so that we see the changes done by external PRs, @erezsh.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,17 +3,21 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


erezsh:
Great, thanks!

> It may also make sense to run codecov on pull_request instead of on push 

Sure, fine by me.

MegaIng:
Actually, making codecov run on pull_request will not work for forked repos anyway, so it's not that valuable.

PR #1641, opened 2026-08-22
Fix InteractiveParser.copy() sharing the lexer thread with the original

undeemed:
# Fix `InteractiveParser.copy()` sharing the lexer thread with the original

### The problem

`InteractiveParser.copy()` makes a copy of the lexer thread for the new instance, but
`ParserState.copy()` keeps a reference to the *original* lexer thread (the `# XXX copy`
note in `lark/parsers/lalr_parser_state.py`). So a copied interactive parser ends up with
two different lexer threads: `ip.lexer_thread` (its own) and `ip.parser_state.lexer` (the
original's).

That matters because `resume_parse()` goes through `_Parser.parse_from_state()`, which
reads tokens from `state.lexer` — the original's thread. Resuming a copy therefore re-lexes
the input from wherever the *original* happens to be, and advances the original's lexer
state while doing so.

```python
from copy import copy
from lark import Lark

parser = Lark('start: A+\nA: "a"', parser='lalr')

ip = parser.parse_interactive("aaa")
ip_copy = copy(ip)
ip_copy.exhaust_lexer()          # the copy consumed all three tokens
print(ip_copy.resume_parse())
```

Before this change:

```
Tree('start', [Token('A', 'a'), Token('A', 'a'), Token('A', 'a'),
               Token('A', 'a'), Token('A', 'a'), Token('A', 'a')])
```

The three tokens the copy already fed are silently fed a second time, because the original's
lexer thread was still at position 0. The same happens through `ImmutableInteractiveParser`,
where every `feed_token()` returns a copy, so any `resume_parse()` on the resulting instance
replays the whole input.

Expected (and what you get after this change):

```
Tree('start', [Token('A', 'a'), Token('A', 'a'), Token('A', 'a')])
```

This contradicts the documented contract of `__copy__`: *"Create a new interactive parser
with a separate state. Calls to feed_token() won't affect the old instance, and vice-versa."*

### The fix

`ParserState.copy()` takes an optional `lexer` argument, and `InteractiveParser.copy()`
passes the lexer thread copy it already creates, so `parser_state.lexer is lexer_thread`
stays true for the copy, exactly as it is for a freshly created interactive parser
(`_Parser.parse()` builds it that way).

The default behaviour of `ParserState.copy()` is unchanged — it still shares the lexer
thread — so the internal caller in `parser_frontends.py` (used by `scan()`, which only
feeds `$END` to a throwaway state) does not pay for an extra lexer copy.

### Tests

Added `test_interactive_parser_copy_resume` next to the existing interactive-parser tests
in `tests/test_parser.py`. It fails on master with
`Tree('start', ['a','a','a','a','a','a']) != Tree('start', ['a','a','a'])`
and passes with the fix. It also checks that driving the copy leaves the original able to
parse its own input.

Full suite (`python -m tests`) passes: 1322 tests run, 148 skipped, and `mypy` is clean.

I am new to open source, and this pull request was made with AI assistance.

### DIFF
--- a/lark/parsers/lalr_interactive_parser.py
+++ b/lark/parsers/lalr_interactive_parser.py
@@ -67,12 +67,8 @@ class InteractiveParser:
         return self.copy()
 
     def copy(self, deepcopy_values=True):
-        lexer_thread = copy(self.lexer_thread)
-        return type(self)(
-            self.parser,
-            self.parser_state.copy(deepcopy_values=deepcopy_values, lexer=lexer_thread),
-            lexer_thread,
-        )
+        parser_state = self.parser_state.copy(deepcopy_values=deepcopy_values)
+        return type(self)(self.parser, parser_state, parser_state.lexer)
 
     def __eq__(self, other):
         if not isinstance(other, InteractiveParser):
--- a/lark/parsers/lalr_interactive_parser.py
+++ b/lark/parsers/lalr_interactive_parser.py
@@ -67,10 +67,11 @@ class InteractiveParser:
         return self.copy()
 
     def copy(self, deepcopy_values=True):
+        lexer_thread = copy(self.lexer_thread)
         return type(self)(
             self.parser,
-            self.parser_state.copy(deepcopy_values=deepcopy_values),
-            copy(self.lexer_thread),
+            self.parser_state.copy(deepcopy_values=deepcopy_values, lexer=lexer_thread),
+            lexer_thread,
         )
 
     def __eq__(self, other):
--- a/lark/parsers/lalr_parser_state.py
+++ b/lark/parsers/lalr_parser_state.py
@@ -1,5 +1,5 @@
 from copy import deepcopy, copy
-from typing import Dict, Any, Generic, List
+from typing import Dict, Any, Generic, List, Optional
 from ..lexer import Token, LexerThread
 from ..common import ParserCallbacks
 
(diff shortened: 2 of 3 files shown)


erezsh:
Thanks, it's a real bug and the fix is technically correct.

However I think the right thing to do is just always copy the lexer. I can't think of a reason for why the lexer should be shared in the first place. 

I'm thinking along these lines:

```python

# lalr_parser_state.py
    def copy(self, deepcopy_values=True) -> 'ParserState[StateT]':
        return type(self)(
            self.parse_conf,
            copy(self.lexer),
            copy(self.state_stack),
            deepcopy(self.value_stack) if deepcopy_values else copy(self.value_stack),
        )

# lalr_interactive_parser.py
    def copy(self, deepcopy_values=True):
        parser_state = self.parser_state.copy(deepcopy_values=deepcopy_values)
        return type(self)(self.parser, parser_state, parser_state.lexer)
```

undeemed:
Yeah that's simpler. Took your version of both methods and dropped the extra parameter.

The new test catches the bug on master and passes with the change. Full suite: 1322 passed, 148 skipped. mypy clean. The only other caller of ParserState.copy() is scan(), which never lexes from the copy, so it's safe there.

I benchmarked it: scanning 400 matches goes from about 19ms to 22ms, roughly 17% slower. That is one lexer copy per match, around 8µs each in practice. Real grammars won't notice, but if you'd rather scan() skip it, that's easy to keep internally.

erezsh:
Thanks for contributing!

Issue #1194, opened 2022-09-23
Standalone module generation output varies


bcr:
**Describe the bug**

When generating output using the minimal grammar of `start:` (as well as more complex grammars, but this reproduces it in my environment) the output varies. On cursory examination running several times, there appear to be two variants that differ by the `states` that are generated.

**To Reproduce**

```sh
echo "start:" > test.g
python3 -m lark.tools.standalone test.g > test.py
python3 -m lark.tools.standalone test.g > test1.py
python3 -m lark.tools.standalone test.g > test2.py
python3 -m lark.tools.standalone test.g > test3.py
diff test.py test1.py
diff test.py test2.py
diff test.py test3.py
```

The diff output on my machine is summarized below.

```diff
--- test.py	2022-09-22 23:17:36.000000000 -0700
+++ test1.py	2022-09-22 23:17:39.000000000 -0700
@@ -3091,7 +3091,7 @@
 
 import pickle, zlib, base64
 DATA = (
-{'parser': {'lexer_conf': {'terminals': [], 'ignore': [], 'g_regex_flags': 0, 'use_bytes': False, 'lexer_type': 'contextual', '__type__': 'LexerConf'}, 'parser_conf': {'rules': [{'@': 0}], 'start': ['start'], 'parser_type': 'lalr', '__type__': 'ParserConf'}, 'parser': {'tokens': {0: 'start', 1: '$END'}, 'states': {0: {0: (0, 1), 1: (1, {'@': 0})}, 1: {}}, 'start_states': {'start': 0}, 'end_states': {'start': 1}}, '__type__': 'ParsingFrontend'}, 'rules': [{'@': 0}], 'options': {'debug': False, 'keep_all_tokens': False, 'tree_class': None, 'cache': False, 'postlex': None, 'parser': 'lalr', 'lexer': 'contextual', 'transformer': None, 'start': ['start'], 'priority': 'normal', 'ambiguity': 'auto', 'regex': False, 'propagate_positions': False, 'lexer_callbacks': {}, 'maybe_placeholders': False, 'edit_terminals': None, 'g_regex_flags': 0, 'use_bytes': False, 'import_paths': [], 'source_path': None, '_plugins': {}}, '__type__': 'Lark'}
[...]

Issue #584, opened 2020-05-08
Generated standalone parsers are not always identical for the same input


kitchoi:
Given a fixed grammar file to generate a standalone parser using `lark.tools.standalone`, the resulting generated code is not always identical despite setting PYTHONHASHSEED.

Note that the syntax tree given by the parser is fine. This is just about the generated code not being static. People who try to generate a standalone parser and commit the code to their repository will find diff just by regenerating the parser, even if the grammar file has not changed.

**To Reproduce**

Grammar `grammar.lark`:
```
?start: item1* item2

item1: "a"
item2: "b"
```

Script to generate standalone parser:
```
export PYTHONHASHSEED=1
python -m lark.tools.standalone grammar.lark > file1.py
python -m lark.tools.standalone grammar.lark > file2.py
diff file1.py file2.py
```

Ideally, the diff should report nothing. However the diff almost always reports some differences. It seems those differences always appear in `DATA` and `MEMO`.

lark-parser version: 0.8.5
Python version: 3.8
OS version: Mac OSX 10.15

kitchoi:
Ah, a simpler grammar file would also reproduce this:
```
start: item
item: "a"
```
But the differences occur less often.

MegaIng:
Can not reproduce when `PYTHONHASHSEED` is set. Without (or with `=random`, or with `-R` flag), can reproduce.

erezsh:
This can happen because the hashes change between runs (which PYTHONHASHSEED should solve) but also because the id() function returns different values. That causes certain operations to occur at arbitrary order.

I don't consider this a bug, but I will accept a PR that fixes it.

Probably the easiest fix would be sorting the culprits during serialization, but that might not be enough. Possibly some of the `id()` calls will have to be replaced with more deliberate indexing.

Meanwhile I can recommend only generating the standalone parser if the file is modified (can be tested by date or hash of grammar).

fzlzjerry:
I can work on the remaining standalone-output reproducibility here. On current master, the grammar from the report still produces 10 different files across 10 fresh processes with `PYTHONHASHSEED=1`. All generated parsers give the same results on valid inputs and reject the same invalid inputs.

#1642 fixed regex flag ordering, but the parse-table/state ordering still varies. I'll focus on stable state numbering and serialization order, with cross-process output checks and parser round trips, without changing parsing decisions.

PR #1642, opened 2026-08-22
Make generated regexps deterministic across processes

ropoctl:
### Problem

`Pattern.flags` is a `frozenset`, and `_get_flags` iterates it directly to wrap the value in `(?f:...)` groups. Python randomizes string hashing per process, so a terminal with more than one flag renders to a **different regexp in every process**:

```console
$ for s in 0 1 2; do PYTHONHASHSEED=$s python -c "
from lark import Lark
print([t.pattern.to_regexp() for t in Lark('start: A+\nA: /x/imsu\n', parser='lalr').terminals])"; done
['(?i:(?u:(?s:(?m:x))))']
['(?m:(?u:(?i:(?s:x))))']
['(?u:(?i:(?m:(?s:x))))']
```

Matching is unaffected — every flag applies whatever the nesting order — but the rendered string is user-visible: `Pattern.__repr__` is `repr(self.to_regexp())`, so it appears in `LexError("Cannot compile token %s: %s")`, in `TerminalDef.user_repr()`, and in the collision warnings.

### Fix

Iterate `sorted(self.flags)`. `flags` stays a `frozenset`, so the subset check in `_create_unless` is unchanged.

This is one source of hash-order dependence, not all of them — it does **not** make the standalone generator or `serialize()` reproducible (I measured: still 5 distinct outputs over `PYTHONHASHSEED=0..4`, same as before). #584 and #1194 have further causes.

### Test

One test. It uses **two or more flags on a single terminal**, which as far as I can tell nothing else does — `test_token_flags` and `test_join_regex_flags` are single-flag-per-terminal, and with one flag there is no order to get wrong. Against unpatched `master` it fails on 10 of 10 `PYTHONHASHSEED` values; with the patch, 0 of 10.

It lives on the existing `TestLexer` class rather than a new one, since `tests/__main__.py` imports specific classes — a new `TestCase` would never be collected. Happy to move it to `test_parser.py` alongside the other flag tests if you'd prefer it there.

Full suite: 1313 tests, 0 failures.

### DIFF
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -69,8 +69,7 @@ class Pattern(Serialize, ABC):
         raise NotImplementedError()
 
     def _get_flags(self, value):
-        # Sorted: flags is a frozenset, so iterating it directly makes the result depend
-        # on hash randomization and differ between processes.
+        # Sorted to remove the hash randomization in sets, that differs between processes
         for f in sorted(self.flags):
             value = ('(?%s:%s)' % (f, value))
         return value
--- a/lark/lexer.py
+++ b/lark/lexer.py
@@ -69,7 +69,9 @@ class Pattern(Serialize, ABC):
         raise NotImplementedError()
 
     def _get_flags(self, value):
-        for f in self.flags:
+        # Sorted: flags is a frozenset, so iterating it directly makes the result depend
+        # on hash randomization and differ between processes.
+        for f in sorted(self.flags):
             value = ('(?%s:%s)' % (f, value))
         return value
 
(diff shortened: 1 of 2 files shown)


erezsh:
Reasonable contribution, but the comments and tests are a bit excessive.

ropoctl:
sorry for slop- i cut to 1 test

ropoctl:
btw, the impetus for looking at this was having an agent write it in rust, and noticing gaps in conformance test suite/ bugs
https://github.com/lark-parser/lark/discussions/1546
the rust version now seems to work ok

erezsh:
Looks better!

Btw FYI, I'm also working on a rust version of Lark.

erezsh:
Thanks for contributing!

PR #1643, opened 2026-08-22
Validate literal ranges with GrammarError instead of assert

ropoctl:
### Problem

`PrepareLiterals.range` validates literal ranges with bare `assert`s:

```python
assert start.type == end.type == 'STRING'
assert len(eval_escaping(start)) == len(eval_escaping(end)) == 1
```

Normally a bad range surfaces as `VisitError` wrapping a message-less `AssertionError`. More importantly, **`-O` strips asserts**, so the check disappears and the grammar is silently accepted:

```console
$ python -O -c "
from lark import Lark
p = Lark('start: A\nA: \"ab\"..\"z\"\n')
print([t.pattern.to_regexp() for t in p.terminals])"
['[ab-z]']
```

`[ab-z]` matches `a` plus the range `b`–`z` — not what the grammar says, and no error is raised. Input validation can't live in an `assert`.

### Fix

Raise `GrammarError`, following #1596.

```
GrammarError: Literal range must be defined by single characters, not "ab".."z"
```

### Tests

Two, both regression tests — a multi-character endpoint and an escaped one (`"\x41\x42".."z"`, so the check is on the *unescaped* length). Against unpatched `master` they **error** under normal execution and **fail** under `-O`, for different reasons: the wrong exception type escapes, versus no exception at all.

I dropped the valid-range test from the first revision — `test_unicode_literal_range_escape` and `test_hex_literal_range_escape` already cover that.

Full suite: 1314 tests, 0 failures, in both modes.

### DIFF
--- a/lark/load_grammar.py
+++ b/lark/load_grammar.py
@@ -605,7 +605,10 @@ class PrepareLiterals(Transformer_InPlace):
         assert start.type == end.type == 'STRING'
         start = start.value[1:-1]
         end = end.value[1:-1]
-        assert len(eval_escaping(start)) == len(eval_escaping(end)) == 1
+        # Not an assert: -O strips it, silently accepting "ab".."z" as the class [ab-z].
+        if len(eval_escaping(start)) != 1 or len(eval_escaping(end)) != 1:
+            raise GrammarError("Literal range must be defined by single characters, "
+                               "not \"%s\"..\"%s\"" % (start, end))
         regexp = '[%s-%s]' % (start, end)
         return ST('pattern', [PatternRE(regexp)])
 
--- a/tests/test_grammar.py
+++ b/tests/test_grammar.py
@@ -151,6 +151,14 @@ class TestGrammar(TestCase):
             """
         self.assertRaises(GrammarError, Lark, g)
 
+    def test_literal_range_multichar(self):
+        # Under `python -O` the old assert was stripped and the range silently compiled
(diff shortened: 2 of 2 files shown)


erezsh:
You included a test for the "ordinary case". But isn't it already tested?

ropoctl:
thanks for catching- removed redundant slop

erezsh:
Here's what Opus had to say ( and I agree ):

**The first check is unreachable — it should stay an assert**

> 
> The meta-grammar at lark/load_grammar.py:156 is 'range': ['STRING _DOTDOT STRING']. A range tree's children are fixed at parse time and are always STRING tokens — no template substitution or other transform produces one. So this branch cannot fire from any grammar a user can write.
> 
> That matters for two reasons:
> 
> - It's the wrong exception type. GrammarError tells the user they wrote a bad grammar. If this ever fired it would mean lark's own meta-grammar broke — an internal invariant violation, which is what assert is for. The PR's own rationale (-O strips the check → silent miscompilation) doesn't apply here: stripping an always-true assert changes nothing.
> - It's the only new branch with no test, and it can't have one.
> 
> Suggest reverting that half to assert start.type == end.type == 'STRING' and keeping the fix focused on the second check, which is the one carrying the actual bug. This also keeps the diff to what the title and body claim.
>

ropoctl:
ok i think that should be addressed

PR #1645, opened 2026-08-25
Add readable auto-names for >=, <=, and != terminals (#836)

### LINKED ISSUE #836: Better token formatting available?
**What is your question?**

I have a Lark grammar that has the following production:

```
!inneroperator : "has" | "=" | "<" | ">" | "<=" | ">=" | "!="
```

(The whole grammar should parse statements like `foo < 3`, pretty simple stuff actually)

When I give `parse` something invalid, such as `foo 3`, the error message reads:

```
No terminal defined for '3' at line 1 col 8
foo 3
Expecting: {'EQUAL', 'HAS', 'MORETHAN', '__ANON_1', '__ANON_0', '__ANON_2', 'LESSTHAN'}
```

This is already quite readable, and I'm impressed. However, I don't know what these `ANON` things are, and also I would ideally output `=` instead of `EQUAL`. Is that possible?

ChrisJr404:
Fixes #836.

Right now common two-character operators like `>=`, `<=` and `!=` turn into `__ANON_0`, `__ANON_1` and so on, which makes error messages harder to read since the "Expected one of" list just shows the anon names.

This adds those three to `_TERMINAL_NAMES` so they get `MOREEQUAL`, `LESSEQUAL` and `NOTEQUAL`, matching the existing single-char entries (`>` MORETHAN, `<` LESSTHAN, `=` EQUAL). @erezsh suggested exactly this in the issue.

If a grammar already defines a terminal with one of those names, it still falls back to `__ANON` as before, so nothing breaks. Added a couple of tests covering both the new names and the collision fallback.


erezsh:
Good addition, but the tests are a bit silly.

PR #1651, opened 2026-09-11
fix exponential backtracking in STRING and REGEXP grammar terminals

netliomax25-code:
1. The `STRING` and `REGEXP` terminals of the grammar lexer match escaped literals with `(\\"|\\\\|[^"\n])*?`, where a backslash can be consumed either as the start of an escape pair or as a plain character through the negated class, so the alternation is ambiguous.
2. On an unterminated literal built from backslashes the engine tries every way to group them before failing, which is exponential. Loading an untrusted grammar of ~60 bytes through `Lark()` runs effectively forever:

   ```python
   import lark
   lark.Lark('start: "' + '\\' * 60)   # never returns
   ```

Made the backslash deterministic instead: an escape now always consumes a pair (`\\.` / `\\[\s\S]`) and the plain-character class excludes backslash (`[^\\"\n]` / `[^\\/]`). Matching is linear and the accepted language of valid terminals is unchanged, since the only inputs that tokenized differently were ones already rejected later by `eval_escaping` as bad escaping. The same terminal appears in the `lark.lark` meta-grammar, so it is fixed there as well.

Validation: full suite green (`python -m tests`), mypy clean, and the added regression `test_terminal_literal_backtracking` hangs on master and passes on the branch.


erezsh:
Did you check if it actually runs faster with this change?

netliomax25-code:
Yes, measured on master (9a4fb9c) vs this branch, Python 3.11, timing only the `Lark(grammar)` call.

1. Unterminated literal of N backslashes (`start: "` + `\` * N). Master grows about 1.6x per extra backslash; the branch stays at the ~64 ms it costs to build the meta-parser in a fresh process:

   | N | master | branch |
   |---|---|---|
   | 24 | 0.07 s | 0.06 s |
   | 28 | 0.10 s | 0.06 s |
   | 32 | 0.34 s | 0.06 s |
   | 36 | 1.95 s | 0.06 s |
   | 60 | killed at 20 s | 0.06 s |
   | 100000 | killed at 20 s | 0.07 s |

   The REGEXP form (`start: /` + backslashes) behaves the same: master 2.9 s at N=36 and killed at 60, branch flat. On the branch, going from 100k to 400k backslashes adds about 8 ms, so the failure path is linear now.

2. Normal grammars are unchanged. `_parse_grammar` on python.lark, best of 7 per process over 6 interleaved processes per tree: master 10.30 ms min / 10.40 median, branch 10.26 ms min / 10.49 median. Full `Lark(python.lark, parser='lalr')` construction is ~330-340 ms on both, within run-to-run noise.

3. At the raw regex level, over the 698 STRING and 150 REGEXP literals in the shipped grammars and examples (both versions tokenize all of them identically), STRING is a wash (164 vs 160 ms for 2000 passes) and REGEXP is roughly 1 us per literal slower in isolation (774 vs 1079 ms for 2000 passes). That comes from `[^\\/]` compiling to a charset where `[^/]` was a single NOT_LITERAL op. Excluding the backslash from the plain class is what removes the ambiguity, so I don't see a way around it, and it doesn't show up in the loading times above.

The regression test in the PR is the N=60 case with a 5 s bound; it hangs on master and passes here.

Issue #1300, opened 2023-07-14
example: reconstruct_python.py failing in half cases


zarnovican:
## Short description

When running the example `examples/advanced/reconstruct_python.py`, the code fails on `AssertionError` in approx 50% of cases.

```
Traceback (most recent call last):
  File "/home/zarnovic/git/lark/examples/advanced/reconstruct_python.py", line 86, in <module>
    test()
  File "/home/zarnovic/git/lark/examples/advanced/reconstruct_python.py", line 80, in test
    assert tree == tree_new
           ^^^^^^^^^^^^^^^^
AssertionError
```

## To Reproduce

Execute the example in a loop. "1" indicate AssertionError. "0" is success.

```bash
$ cd examples/advanced/
$ for i in `seq 10`; do python reconstruct_python.py 2>&1 | grep -c AssertionError; done
1
1
1
1
0
1
0
0
1
0
```

## Test environment

OS: Linux
Python: 3.11
Lark: 1.1.5

I have tested few older Pythons and older Larks as well. No change.

## Long Description

As I understand the example, it is converting text Python to AST, then back to Python via "Reconstructor" and then, again to AST. The assertion is that the first and second AST trees should be the same.

After some debugging, I was able to isolate much smaller reproducer:

```python
from lark import Token, Lark
from lark.reconstruct import Reconstructor
from lark.indenter import PythonIndenter

# Official Python grammar by Lark
python_parser3 = Lark.open_from_package('lark', 'python.lark', ['grammars'],
                                        parser='lalr', postlex=PythonIndenter(), start='file_input',
                                        maybe_placeholders=False    # Necessary for reconstructor
                                        )

def special(sym):
    return Token('SPECIAL', sym.name)

def test():
    r = Reconstructor(python_parser3, {'_NEWLINE': special, })
    tree = python_parser3.parse('''foo('a', 'b')\n''')
    output = r.reconstruct(tree)
    print(output)

if __name__ == '__main__':
    test()
```

The above code is converting `foo('a', 'b')` to AST and back to python. The Reconstructor produces four variations of the code

```bash
$ for i in `seq 10`; do python reconstruct_python_test.py; done | sort -u
foo('a''b',)_NEWLINE
foo('a','b',)_NEWLINE
_NEWLINE foo('a''b',)_NEWLINE
_NEWLINE foo('a','b',)_NEWLINE
```

[...]

Issue #1172, opened 2022-07-27
GrammarError when rule definition contains TERM?rule


chanicpanic:
**Describe the bug**

Lark raises a `GrammarError` when the provided grammar contains a rule of the form `name: TERM?rule`. 

I expect `name: TERM?rule` to be valid lark syntax and be parsed equivalently to `name: TERM? rule`.

**To Reproduce**

```
>>> Lark('start: "A"?b\nb:')
Traceback (most recent call last):
  File "/__/lark/parsers/lalr_parser.py", line 125, in feed_token
    action, arg = states[state][token.type]
KeyError: 'RULE_MODIFIERS'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/__/lark/load_grammar.py", line 961, in _parse_grammar
    tree = _get_parser().parse(text + '\n', start)
  File "/__/lark/parser_frontends.py", line 96, in parse
    return self.parser.parse(stream, chosen_start, **kw)
  File "/__/lark/parsers/lalr_parser.py", line 40, in parse
    return self.parser.parse(lexer, start)
  File "/__/lark/parsers/lalr_parser.py", line 170, in parse
    return self.parse_from_state(parser_state)
  File "/__/lark/parsers/lalr_parser.py", line 187, in parse_from_state
    raise e
  File "/__/lark/parsers/lalr_parser.py", line 178, in parse_from_state
    state.feed_token(token)
  File "/__/lark/parsers/lalr_parser.py", line 128, in feed_token
    raise UnexpectedToken(token, expected, state=self, interactive_parser=None)
lark.exceptions.UnexpectedToken: Unexpected token Token('RULE_MODIFIERS', '?') at line 1, column 11.
Expected one of:
        * _LPAR
        * _NL_OR
        * _NL
        * STRING
        * OP
        * _OR
        * REGEXP
        * _TO
        * _DOTDOT
        * _LBRA
        * RULE
        * TERMINAL
        * TILDE


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/__/lark/lark.py", line 336, in __init__
    self.grammar, used_files = load_grammar(grammar, self.source_path, self.options.import_paths, self.options.keep_all_tokens)
  File "/__/lark/load_grammar.py", line 1408, in load_grammar
    builder.load_grammar(grammar, source)
  File "/__/lark/load_grammar.py", line 1233, in load_grammar
[...]

Issue #927, opened 2021-06-29
Tree Matcher/reconstructor incorrectly (un-)flattens children


MegaIng:
**Describe the bug**

Sometimes (e.g. not reliable/random chance), the `TreeMatcher` (and therefore the `Reconstructor` applies trees like this:

```
expr
    literal "a"
    "5"
    "5"
```

as if they were

```
expr
    expr
        literal "a"
        "5"
    "5"
```

**To Reproduce**

```
parser = Lark.open_from_package("lark", "lark.lark", ("grammars",))

lark_reconstructor = Reconstructor(parser, {
    "_NL": lambda _: "\n",
    "_VBAR": lambda _: "|",
})

pattern = parser.parse("""
start: "a"~5..5
start: ("b"~5)~5
""")
print(pattern.pretty())
print(lark_reconstructor.reconstruct(pattern))

```
This script sometimes wrongly produces

```
start:("a"~5)~5
start:("b"~5)~5
```

(I came across this while working on [ast_generator](https://github.com/MegaIng/lark_ast_generator))

---

Maybe related: Sometimes, the tree_matcher outputs normally illegal constructs of the shape `Tree('expansion',[Tree('expansion',[...])])` or `Tree('<origin>', [Tree('<alias>',[...])])` This can be dealt with, but is kinda annoying. I don't have a simple example. [Here](https://github.com/MegaIng/lark_ast_generator/blob/a283d46d0aaabe85b056b97d4e874ff24963dda2/rules_generator.py#L45) is where I deal with it.

erezsh:
Sounds like something that can be solved by changing priority / ordering of choices?

MegaIng:
@erezsh I am not sure. I don't know exactly how `tree_matcher` is implemented.

PR #1652, opened 2026-09-13
Reconstructor: separate ? from the item that follows it

Endika:
`Reconstructor` emits `r:a?b` for `r: a? b`. That output is not the grammar it came from:

- fed back to lark it does not load at all — `GrammarError: Unexpected colon, at line 1 column 5`;
- parsed with lark's own `lark/grammars/lark.lark` it re-parses into a **different** tree: `?b` is a single `RULE` token there, so the optional operator is gone and the reference to `b` has become a reference to `?b`.

```python
from lark import Lark

parser = Lark.open_from_package('lark', 'lark.lark', ['grammars'],
                                parser='lalr', maybe_placeholders=False)
tree = parser.parse('r: a? b\n')
out = Reconstructor(parser).reconstruct(tree)   # 'r:a?b'
parser.parse(out + '\n') == tree                # False
```

### Why only `?`

`OP` is `[+*]|[?](?![a-z_])` (`load_grammar.py:83`): a `?` immediately followed by a name character is not the operator. `*`, `+` and `~n` carry no such lookahead, so `r: a* b` reconstructs to `r:a*b` and round-trips today. I checked each one; this is specific to `?`.

### Root cause

`reconstruct.py:105` decides separation by character class — a space goes in only when the characters on both sides are `is_id_continue` — and `?` is not one. The change adds the single case where a trailing operator can still merge with the name after it.

### Prior art

#1172 reported this in 2022, found the same way (reconstructing `lark.lark`), and was closed with the advice to pass a `postproc` that inserts the space. That works and it is what the reporter did. This PR proposes the library do it instead, because the grammars lark itself ships are affected and every caller reconstructing a Lark grammar has to rediscover the same three lines. `a?b` stays illegal input — the only thing that changes is that lark stops emitting it.

### Blast radius, measured on this repo's own grammars

Reconstruct each of the 524 top-level statements in the 37 `.lark` files here, re-parse, compare trees, then repair *only* the `?` spacing to attribute the cause: **6 statements in 5 files** are this defect, and all 6 round-trip after the fix. Two are in grammars lark ships:

| file | statement | reconstructed |
|---|---|---|
| `lark/grammars/lark.lark` | `!import_path: "."? name ("." name)*` | `!import_path:"."?name("."name)*` |
| `lark/grammars/python.lark` | `?await_expr: AWAIT? atom_expr` | `?await_expr:AWAIT?atom_expr` |
| `examples/advanced/python2.lark` | `suite: … _NEWLINE? stmt+ …` | `…_NEWLINE?stmt+…` |
| `examples/advanced/template_lark.lark` | `!import_path{name}: "."? name ("." name)*` | `…"."?name…` |
| `examples/composition/json.lark` | `array` and `object` (`_WS? value`) | `…_WS?value…` |

The other statements that do not round-trip have unrelated causes (alternation handling and terminal patterns, cf. #927, #1300) and are untouched here.

### Tests

`python -m tests` → 1330 tests, OK (181 skipped); 1329 before, plus the one added. The new test is red before the change and green after. The existing reconstructor tests compare with whitespace stripped, which is why none of them could catch this.

### Scope

This is the narrow fix. The general one is lexical: ask the target grammar's lexer whether two adjacent items would re-lex as one token, rather than deciding by character class. That would also cover the opposite case — a grammar where `?` is a prefix sigil and the name is a separate token, where this patch inserts a space that is not wanted. I did not want to change the spacing policy of every reconstructed output on my own judgement, but I am happy to write the lexical version instead if you prefer it.


erezsh:
Low effort. At least read the code you're submitting.

PR #1655, opened 2026-09-18
feat: add support for 'a' (ASCII) and 'L' (LOCALE) regex flags

### LINKED ISSUE #1527: Add support for the `a` and `L` regex flags

amansri-30:
## Summary

Adds support for the `a` (ASCII) and `L` (LOCALE) regex flags in Lark grammars, as requested in issue #1527.

## Changes

- Updated `_RE_FLAGS` in `load_grammar.py` to include `a` and `L`
- Added `test_flag_order_with_new_flags` to verify the `a` flag works and is sorted deterministically
- Added `test_token_flags_ascii` to verify the `a` flag parses correctly

## Details

- **`a` flag (ASCII)**: Makes `\w`, `\W`, `\b`, `\B`, `\d`, `\D`, `\s`, `\S` perform ASCII-only matching instead of full Unicode matching. Works with both string and bytes patterns.
- **`L` flag (LOCALE)**: Makes `\w`, `\W`, `\b`, `\B` locale dependent. Note: In Python 3, this flag can only be used with bytes patterns (when `use_bytes=True`).

## Testing

All existing tests pass, and new tests verify the flags work correctly:

```bash
pytest tests/test_lexer.py tests/test_parser.py
```

Closes #1527


erezsh:
low effort

PR #1657, opened 2026-09-19
feat: add support for 'a' (ASCII) and 'L' (LOCALE) regex flags

### LINKED ISSUE #1527: Add support for the `a` and `L` regex flags

amansri-30:
## Summary

Adds support for the `a` (ASCII) and `L` (LOCALE) regex flags in Lark grammars, closing #1527.

Beyond adding the flags to `_RE_FLAGS`, this PR makes them safe to use: flag combinations are validated at grammar-compile time with clear `GrammarError`s, the `L` flag works end-to-end with bytes lexers, the self-hosted grammars accept the new flags, the flags are documented, and the tests verify the actual ASCII/LOCALE semantics rather than just successful compilation.

## Changes

**Grammar parsing** (`lark/load_grammar.py`)
- `_RE_FLAGS` now includes `a` and `L`, so `/regex/a`, `/regex/L`, `"select"i`-style flag suffixes accept them
- Using `a` and `u` on the same literal raises a clear `GrammarError` (they are mutually exclusive in `re`)

**Validation** (`lark/lark.py`)
- `L` requires `use_bytes=True` — in Python 3 the LOCALE flag is only valid for bytes patterns; previously this crashed with a cryptic `re.PatternError` deep inside `re.compile`
- `u` with `use_bytes=True` raises a clear error for the equivalent reason

**Lexer plumbing** (`lark/lexer.py`, `lark/utils.py`)
- `BasicLexer` validates terminal patterns in their final form (bytes under `use_bytes`), so `(?L:...)` patterns compile cleanly
- `get_regexp_width` neutralizes `(?L:` before handing the pattern to `sre_parse`, which rejects it on str patterns (flags don't affect match width, so this is safe)
- `Pattern._get_flags` already sorts flags, so `a`/`L` combine deterministically with the existing flags

**Self-hosted grammars**
- `lark/grammars/lark.lark` and `examples/advanced/template_lark.lark` REGEXP terminals updated to `[imsluxaL]*`

**Docs** (`docs/grammar.md`)
- Flag list updated to `imsluxaL`, with notes on `a`/`u` exclusivity and the bytes-only requirement of `L`

## Testing

New and strengthened tests:
- `test_flag_order_with_new_flags` — `a` and `L` sort deterministically (`(?a:(?L:x))`), including with `use_bytes`
- `test_conflicting_ascii_unicode_flags` — `a` + `u` raises `GrammarError`
- `test_token_flags_ascii` — proves ASCII semantics: `/\w+/a` rejects `é`, while the same pattern without the flag matches it
- `test_token_flags_locale_requires_bytes` — `L` without bytes raises `GrammarError`
- `test_token_flags_locale_bytes` — `/\w+/L` with `use_bytes=True` parses bytes input

Full suite: `pytest tests/` → 1213 passed, 148 skipped. `mypy` (as run by `tox -e type`) and the pre-commit hooks pass as well.

Closes #1527


amansri-30:
Hi @erezsh! This supersedes #1655 with a more complete implementation: compile-time validation for the new flags (with clear error messages), bytes-lexer support for the LOCALE flag, self-hosted grammar updates, docs, and tests that verify the actual ASCII/LOCALE semantics.

Could you please approve the workflow runs for this PR when you get a chance? The first-time-contributor approval window expired on the earlier run, so CI currently shows \u201crequired approval but was not approved before it expired\u201d even though the checks themselves should pass. Thanks!

